Issue #20, opened 2026-06-08
model_family misclassifies KeyATM/SeededLDA as "none" (breaks composition_theta; silently degrades tempotm R_t)


nealcaren:
## Summary

`effects.model_family` misclassifies `KeyATM` and `SeededLDA` as `"none"`, even though both are collapsed-Gibbs Dirichlet models. This breaks `composition_theta` for them and contradicts the function's own docstring.

## Detail

`model_family` selects the theta-uncertainty sampler with:

```python
if hasattr(cls, "alpha") and hasattr(cls, "doc_topic"):
    return "dirichlet"
```

Only `LDA` exposes an `alpha` getter. `KeyATM` and `SeededLDA` expose `doc_topic` but **not** `alpha`, so they fall through to `"none"`:

```
LDA        alpha=True   doc_topic=True  -> dirichlet   ✓
SeededLDA  alpha=False  doc_topic=True  -> none        ✗
KeyATM     alpha=False  doc_topic=True  -> none        ✗
```

This contradicts the `model_family` docstring, which states it returns `"dirichlet"` for "the collapsed-Gibbs models (LDA, keyATM, SeededLDA, ...)".

Consequence: `composition_theta(model)` raises `ValueError("... has no posterior over theta ... Use method='bootstrap'")` for KeyATM/SeededLDA, even though their `doc_topic` is exactly a Dirichlet posterior mean (`θ_d ~ Dirichlet(α + n_d)`).

## Repro

```python
import topica
for name in ["LDA", "SeededLDA", "KeyATM"]:
    cls = getattr(topica, name)
    print(name, "alpha:", hasattr(cls, "alpha"), "doc_topic:", hasattr(cls, "doc_topic"))
# LDA alpha: True ...
# SeededLDA alpha: False ...
# KeyATM alpha: False ...
```

## Downstream impact (tempotm)

`tempotm.dynamics.period_prevalence` relies on `composition_theta` to draw `θ`. When `model_family` returns `"none"`, tempotm's `_composition_draws` catches the error and **silently falls back to `S=1`** (a single point estimate), so the per-period measurement covariance `R_t` reflects document-sampling error only — it drops the topic-estimation-uncertainty term (Rubin's between-draw variance). That term is the entire reason tempotm propagates `R_t`. So today, running `period_prevalence` on a `KeyATM` or `SeededLDA` model quietly produces a degraded "honest" covariance with only a warning.

## Suggested fixes (priority order)

[...]

PR #28, opened 2026-06-08
fix: re-land #21 + close silent-degradation / API-drift gaps (#21, #24-27)

### LINKED ISSUE #21: model_family detection is incomplete across the model registry (~8 Dirichlet models -> "none"); add registry-parametrized test
Related to #20 (which covers the `KeyATM`/`SeededLDA` symptom). This issue scopes the **full blast radius** and the **test gap** that let it survive.

## Root cause: `model_family` keys "dirichlet" off `alpha`, which almost no Gibbs model exposes

A capability-surface audit across the whole model registry: `alpha` is present on **only `LDA` and `HDP`**. Every other collapsed-Gibbs model with a `doc_topic` falls through to `"none"`:

| should be `"dirichlet"` but `model_family` returns `"none"` |
|---|
| `KeyATM`, `SeededLDA`, `LabeledLDA`, `SupervisedLDA`, `DMR`, `PA`, `PT`, `GSDMM`*, `SAGE`* |

(*`GSDMM` has hard/degenerate θ and `SAGE` a different word parameterization — decide deliberately, but neither is logistic-normal.)

The `model_family` docstring promises "dirichlet for the collapsed-Gibbs models (LDA, keyATM, SeededLDA, ...)" — delivered for ~2 of ~10.

Repro:
```python
import topica
for n in ["LDA","HDP","KeyATM","SeededLDA","LabeledLDA","SupervisedLDA","DMR","PA","PT","GSDMM","SAGE"]:
    print(n, hasattr(getattr(topica, n), "alpha"))
```

## Consequences

1. **`composition_theta` raises** for all of the above ("no posterior ... use method='bootstrap'"), even though `doc_topic` is a Dirichlet posterior mean.
2. **Downstream silent degradation (tempotm):** `tempotm.dynamics.period_prevalence` catches that and falls back to `S=1`, so `R_t` carries document-sampling error only — dropping the topic-estimation-uncertainty term. Quietly-degraded "honest" covariance, warning only.
3. **viz CI gating:** `viz/capability.py` reuses `model_family` for `theta_posterior`, so the effect panel **silently shows point estimates instead of CI bands** for all these models. (Single source of truth — fixing `model_family` fixes both.)

## Why it survived: test-coverage gap

`tests/test_standard_errors.py::test_model_family_detection` asserts exactly three models — `LDA`→dirichlet, `STM`→logistic_normal, `BERTopic`→none — i.e. the one Gibbs model that happens to expose `alpha`. The MoC/SE suite only ever instantiates LDA/STM/CTM.

## Suggested fixes

1. **Expose an `alpha` property on the Gibbs models** (preferred — lets `dirichlet_theta_samples(prior=alpha)` use the proper concentration `α + n_d`).
2. **Or** detect Dirichlet by `doc_topic` present + not logistic-normal, instead of requiring `alpha`.
3. **Regression test:** parametrize `model_family` (and "`composition_theta` must not raise") over the **full** model class registry with an expected-family table. One test, catches all ~8.

### LINKED ISSUE #24: Silent document misalignment: display functions lack the kept_indices/length guard their siblings have
Related to #20 / #21 — same failure mode (silent, plausible-but-wrong output), different mechanism.

## Problem

`Corpus.from_documents` / `from_text_file` prune the vocabulary and **drop documents** left with no tokens, so `doc_topic` has `len(corpus.kept_indices)` rows, not `len(original_docs)`. Several display/analysis functions index a user-supplied `texts`/`groups` array by `doc_topic` row **without checking the length matches**. If the caller passes the original (pre-pruning) array, the function silently returns the **wrong document** — including the text of a doc that was pruned out. No error, no warning.

The damning part: sibling functions already have the guard, so this is an omission, not a design choice.

## Affected (missing guard)

| Function | File:line | Note |
|---|---|---|
| `find_thoughts` | `validation.py:435` (`texts[i]` at :454) | returns wrong representative doc |
| `document_intrusion` | `coherence.py:472` (texts use ~:518) | returns text of a pruned doc |
| `representative_docs` / `topic_info` | `analysis.py:99` / `:133` | inherit from `find_thoughts` |
| `plot_report` "topics per class" panel | `analysis.py:284–292` | misaligned `groups` → `IndexError` swallowed by bare `except: pass`, panel silently vanishes |

## Already guarded (the pattern to copy)

`find_thoughts_html` (`validation.py:974`), `prepare_pyldavis` (`:690`), `check_residuals` (`:792`), `estimate_effect` (`stm.py:254`), `by_strata` (`keyatm.py:129,153`), `topics_over_time` (`analysis.py:194`) all raise `ValueError` on a length mismatch.

## Repro

```python
import topica
docs = [["the","cat","sat","on","mat"], ["quuxfrobnitz"],   # 2nd doc pruned at min_doc_freq=2
        ["the","dog","ate","the","cat"], ["cat","mat","dog","sat"]]
texts = ["Text0", "DROPPED1", "Text2", "Text3"]
corpus = topica.Corpus.from_documents(docs, min_doc_freq=2)   # kept_indices = [0,2,3]
m = topica.LDA(num_topics=2, seed=42); m.fit(corpus, iterations=100)
print(topica.find_thoughts(m.doc_topic, texts, topic=0, n=3))  # returns "DROPPED1" for a real doc; no error
```

## Fix

Add the same guard `find_thoughts_html` uses, to each function that indexes `texts`/`groups` by `doc_topic` row:

```python
if texts is not None and len(texts) != theta.shape[0]:
    raise ValueError(
        f"texts has {len(texts)} entries but doc_topic has {theta.shape[0]} rows; "
        "pass texts aligned to the kept documents (corpus.kept_indices), not the "
        "original documents — pruning may have dropped some.")
```

### LINKED ISSUE #25: Swallowed exceptions / ignored params silently degrade bootstrap SEs, coherence, and reports
Related to #20 / #21 — same "silent degradation" failure mode (a result quietly less trustworthy than the API implies), via swallowed exceptions and ignored parameters.

### 1. `_bootstrap_refits` swallows genuine `TypeError`, corrupting bootstrap SEs — `effects.py:303`  (High, silent)
The `except TypeError` is intended to detect a single-arg `refit(picks)` hook. But a `TypeError` raised *inside* `refit` for a real reason (bad kwarg, type error in the hook body) is misread as an arity mismatch, and it retries `refit(picks)` — running every resample at the wrong/default seed. `standard_errors()` returns SEs computed from mis-seeded refits, no error or warning.
**Fix:** check arity with `inspect.signature(refit).bind(picks)` before retrying; re-raise the original `TypeError` if the 1-arg call is also invalid. Same pattern at `validation.py:147–154` (`_transform_theta`).

### 2. `quality_frontier` silently ignores `coherence_type` when `texts=None` — `validation.py:~1056`  (Medium, silent)
`quality_frontier(model, coherence_type="c_v")` without `texts` falls back to `model.coherence(n)` (always UMass). The returned array is UMass (range ≈ (−∞,0]) not c_v (range [0,1]) — a different metric on a different scale, with no warning.
**Fix:** if `texts is None` and `coherence_type != "u_mass"`, warn (or raise) instead of silently substituting.

### 3. `plot_report` — four bare `except Exception: pass` — `analysis.py:267,274,281,291`  (Medium, silent)
A genuine bug/error in any optional panel (quality frontier, correlation, topics-over-time, per-class) is swallowed; the figure is returned looking complete but missing panels, with nothing printed.
**Fix:** `warnings.warn` naming the panel + exception; keep only truly-optional (missing-matplotlib) skips silent.

### 4. `_top_words` bare except hides bugs in `model.top_words` — `analysis.py:122`, `validation.py:113`  (Medium, silent)
Any exception from `model.top_words(...)` is discarded and silently replaced by raw-φ top words, dropping any custom weighting (e.g. FREX) done inside `top_words`. Caller sees neither the exception nor a warning.
**Fix:** `warnings.warn` before the fallback.

## Verified non-issues (checked, safe)
`labeling._unknown_model_message` bare except (message-only), `_bootstrap_refits` `model_factory` except (re-raises `from exc`), `posterior_theta_samples` eigh fallback (mathematically correct), `coherence._occurrences` Rust→Python fallback (correct & documented).

### LINKED ISSUE #26: dirichlet_theta_samples double-counts the Dirichlet prior when prior>0 (biases draws toward uniform)
Related to #20 / #21 (same module's uncertainty machinery). Distinct, precise numerical bug.

## Problem — `effects.py:78`

```python
conc = theta * (lengths[:, None] + prior) + prior / theta.shape[1]
```

For a collapsed-Gibbs model, `doc_topic` is the posterior mean `θ_dk = (n_dk + α_k)/(N_d + Σα)`, so the posterior concentration is exactly `α_k + n_dk = θ_dk·(N_d + prior)` (with symmetric `α_k = prior/K`, `Σα = prior`). The code computes that **and then adds `prior/K` again** — double-counting the symmetric prior:

```
sum_k conc_k  =  N_d + 2·prior      (code)
sum_k conc_k  =  N_d + prior        (correct)
```

So draws are biased toward uniform; mean error per document ≈ `prior/(N_d + 2·prior)`, worst for short docs / large priors. The comment on the line itself says the concentration is "α + n_d", which is `θ·(N+prior)` alone — confirming the `+prior/K` addend is unintended.

**Scope:** only the non-default `prior > 0` path (default `prior=0` zeroes the term, so existing default-path results are unaffected). The `np.clip(conc, 1e-6, None)` on the next line already floors zero concentrations, so the addend isn't needed as a numerical guard either.

## Repro

```python
import numpy as np
D, K = 1000, 10
theta = np.full((D, K), 0.2/(K-1)); theta[:,0] = 0.8
lengths = np.full(D, 10.0); prior = 1.0
wrong   = theta*(lengths[:,None]+prior) + prior/K
correct = theta*(lengths[:,None]+prior)
Ew = (wrong/wrong.sum(1,keepdims=True)).mean(0)
Ec = (correct/correct.sum(1,keepdims=True)).mean(0)
print(Ew[0], Ec[0])   # ~0.743 (biased) vs ~0.800 (unbiased), true=0.8
```

## Fix

```python
conc = theta * (lengths[:, None] + prior)
```

(Separately worth considering: `prior=0` ignores the model's fitted `α` entirely — exposing `alpha` on the Gibbs models, per #21, would let this default to the proper concentration.)

### LINKED ISSUE #27: API surface drift: DMR type stubs, coherence/analysis 'any model' docstrings, and viz capability sets disagree with runtime
Related to #20 / #21 (capability/family detection); this issue collects the broader set where the **declared API surface — type stubs, docstrings, and viz capability descriptors — disagrees with runtime behavior.** Following them gives runtime errors or misleading advice.

## 1. `DMR` type stubs are wrong (copy-pasted from STM) — `_topica.pyi:185–228`
- `DMR.fit` stub says `fit(data, prevalence=None, *, prevalence_names=None, content=None, content_names=None, em_iters=50)` and the docstring says "Fit the **STM**". Actual signature: `fit(data, features, *, feature_names=None, iterations=1000, num_samples=5, sample_interval=25, ...)` — `features` is **required**. Following the stub → `TypeError`.
- `DMR.prevalence_effects` property does not exist at runtime; it's `feature_effects` → `AttributeError`. The `feature_names` docstring also references the non-existent `prevalence_effects`.

## 2. `coherence()` "works with any model here" is false — `coherence.py:22`
`coherence(model, texts)` calls `model.top_words(topn)`, but:
- **SAGE** `top_words(topic, *, group=None, n=10)` — first positional is `topic`, so `top_words(10)` → `ValueError: topic 10 out of range`.
- **DTM** `top_words(topic, time, n=10)` — needs two positionals → `TypeError: missing 'time'`.

`summary()` documents this limitation; `coherence()` doesn't guard it. **Fix:** in `_extract_topics`, catch and fall back to `topic_word + vocabulary` for models whose `top_words` doesn't take a plain `n`.

## 3. SAGE breaks the `(K,V)` analysis surface — `analysis.py:7`, `coherence.py:22`
Module docstrings claim the helpers "work uniformly across LDA, STM, CTM, keyATM, ... and the rest." SAGE's `topic_word` is **3-D** `(K, groups, V)`, so `_as_topic_word(sage)` breaks `frex`, `label_topics`, `exclusivity`, `diagnostics`, `check_residuals`, `mmr` — and `label_topics` emits a wrong "BERTopic/Top2Vec" hint. **Fix:** document SAGE as excluded (point to a group-marginal), or have `_as_topic_word` return the marginal with a warning.

## 4. viz capability descriptors miss `HLDA`/`DTM` — `viz/capability.py:17–20`
`HLDA` and `DTM` expose **no** `doc_topic`, but aren't in `_DEGENERATE_THETA`, so `capabilities(...).soft_theta == True`; theta panels then hit `AttributeError` (swallowed by the dashboard's per-panel try). `DTM.topic_word` is a `(time)` **callable**, but `DTM ∉ _CLUSTER`, so `prob_simplex_words == True` and `np.asarray(model.topic_word)` yields an object array → `TypeError`. **Fix:** add `HLDA`/`DTM` to the right sets (and handle DTM's time-sliced `topic_word`).

## 5. `bootstrap_stability` docstring promises `Corpus` it can't take — `validation.py:1107`
Docstring: `docs : ... or a Corpus`. Body does `[list(d) for d in docs]`; a `topica.Corpus` isn't iterable → `TypeError`. Siblings (`perplexity`, `prepare_pyldavis`) handle it via `if hasattr(docs, 'documents'): docs = docs.documents()`. **Fix:** add the same guard.

nealcaren:
Closes #21, closes #24, closes #25, closes #26, closes #27.

Re-lands #21 (which never reached `main` — see the comment below) alongside the sweep.

One combined PR (per request) for the silent-degradation / API-drift batch. Shared theme: a result quietly less trustworthy than the API implied. Every fix makes the failure loud (raise or warn) instead of silent.

## #26 — `dirichlet_theta_samples` double-counted the prior
`conc = theta·(N+prior) + prior/K` added the symmetric prior twice, biasing draws toward uniform on the `prior>0` path. Dropped the extra addend. Default `prior=0` path is unchanged (verified).

## #24 — missing alignment guards
`find_thoughts` and `document_intrusion` now raise on a `texts`/`doc_topic` length mismatch (the guard their siblings already had), so a pruned-out document can't be returned in place of a real one. `representative_docs`/`topic_info` inherit it via `find_thoughts`; `plot_report`'s per-class panel gets the same check.

## #25 — swallowed exceptions / ignored params
- `_bootstrap_refits` and `_transform_theta` decide arity by **inspecting the signature** instead of treating any `TypeError` as an arity mismatch (which silently re-ran every resample at the default seed — the High-severity one).
- `quality_frontier` warns when a windowed `coherence_type` is requested without `texts` (it falls back to UMass, a different scale).
- `plot_report` warns and names any optional panel it drops.
- the `_top_words` fallback warns before discarding custom weighting (FREX).

## #27 — API surface drift
- **DMR stub** was copied from STM: corrected `fit()` to `(data, features, ...)` (features required) and renamed the nonexistent `prevalence_effects` stub to `feature_effects`.
- **`coherence()`/analysis surface** now works for **SAGE** via the group marginal (`_as_topic_word` reduces its `(K, G, V)` `topic_word`) and **rejects DTM**'s time-sliced `topic_word` with a clear message instead of an object-array `TypeError`. `_extract_topics` falls back to `topic_word`+`vocabulary` when `top_words(n)` misfires (SAGE's `topic`, DTM's `time`).
- **viz capability** marks `HLDA` and `DTM` (no usable `doc_topic`) as not `soft_theta`, so theta panels disable instead of hitting `AttributeError`.
- **`bootstrap_stability`** now accepts a `Corpus`, as its docstring and siblings promise.

## Tests
New `tests/test_silent_degradation.py` (12 tests). Two existing tests updated because they encoded the old behavior: `bootstrap_stability` rejecting a `Corpus`, and SAGE's dashboard frontier being skipped (it now works via the marginal — a genuine improvement).

Full suite: 1120 Python (up from 1108) + 49 Rust pass; `mkdocs build --strict` clean. No Rust changes in this PR (Python + `.pyi` only).

### DIFF
--- a/python/topica/_topica.pyi
+++ b/python/topica/_topica.pyi
@@ -210,6 +210,12 @@ class DMR:
         ...
 
     @property
+    def alpha(self) -> numpy.typing.NDArray[numpy.float64]:
+        """The baseline document-topic Dirichlet prior alpha, shape (num_topics,):
+        exp(lambda_intercept), the per-topic prior at covariates = 0."""
+        ...
+
+    @property
     def prevalence_effects(self) -> numpy.typing.NDArray[numpy.float64]:
         """Learned prevalence weights gamma, shape (num_topics, num_features).
         Column 0 is the intercept; positive entries raise that topic's
@@ -680,6 +686,10 @@ class SupervisedLDA:
     @property
     def doc_topic(self) -> numpy.typing.NDArray[numpy.float64]: ...
     @property
+    def alpha(self) -> numpy.typing.NDArray[numpy.float64]:
+        """The symmetric document-topic Dirichlet prior alpha, shape (num_topics,)."""
+        ...
+    @property
     def coefficients(self) -> numpy.typing.NDArray[numpy.float64]:
         """Regression coefficients eta, shape (num_topics,) — how each topic
         moves the response per unit of topic frequency."""
@@ -770,6 +780,13 @@ class SAGE:
         ...
 
     @property
+    def alpha(self) -> numpy.typing.NDArray[numpy.float64]:
+        """The symmetric document-topic Dirichlet prior alpha, shape (num_topics,).
+        SAGE's sparse additive parameterization is on the word side; the document
+        side is an ordinary Dirichlet."""
+        ...
+
+    @property
     def groups(self) -> list[str]:
         """Group names, in the index order of topic_word's second axis."""
         ...
@@ -835,6 +852,11 @@ class LabeledLDA:
     def doc_topic(self) -> numpy.typing.NDArray[numpy.float64]:
         """theta matrix (num_docs, num_topics); only a document's label topics
         are non-zero, rows sum to 1."""
+        ...
+
+    @property
+    def alpha(self) -> numpy.typing.NDArray[numpy.float64]:
+        """The symmetric document-topic Dirichlet prior alpha, shape (num_topics,)."""
         ...
 
     @property
@@ -1116,6 +1138,10 @@ class PT:
     @property
     def doc_topic(self) -> numpy.typing.NDArray[numpy.float64]: ...
     @property
+    def alpha(self) -> numpy.typing.NDArray[numpy.float64]:
+        """The symmetric document-topic Dirichlet prior alpha, shape (num_topics,)."""
+        ...
+    @property
     def num_topics(self) -> int: ...
     @property
     def vocabulary(self) -> list[str]: ...
@@ -1213,6 +1239,10 @@ class PA:
     @property
     def doc_topic(self) -> numpy.typing.NDArray[numpy.float64]:
         """Document sub-topic matrix, shape (num_docs, num_sub)."""
+        ...
+    @property
+    def alpha(self) -> numpy.typing.NDArray[numpy.float64]:
+        """The symmetric sub-topic Dirichlet prior alpha, shape (num_sub,)."""
         ...
     @property
     def super_sub(self) -> numpy.typing.NDArray[numpy.float64]:
--- a/python/topica/_topica.pyi
+++ b/python/topica/_topica.pyi
@@ -185,18 +185,21 @@ class DMR:
     def fit(
         self,
         data: Corpus | Sequence[Sequence[str]],
-        prevalence: numpy.typing.NDArray[numpy.float64] | Sequence[Sequence[float]] | None = None,
+        features: numpy.typing.NDArray[numpy.float64] | Sequence[Sequence[float]],
         *,
-        prevalence_names: list[str] | None = None,
-        content: Sequence[object] | None = None,
-        content_names: list[str] | None = None,
-        em_iters: int = 50,
+        feature_names: list[str] | None = None,
+        iterations: int = 1000,
+        num_samples: int = 5,
+        sample_interval: int = 25,
+        progress: object | None = None,
+        progress_interval: int = 50,
     ) -> None:
-        """Fit the STM. prevalence is an (num_docs, F) covariate matrix making
-        topic prevalence depend on covariates (an intercept is prepended;
-        prevalence_names names the F columns). content is one group label per
-        document for a SAGE content model. At least one of prevalence/content
-        should be given (else use CTM)."""
+        """Fit by collapsed Gibbs with the per-document Dirichlet prior
+        alpha_{d,t} = exp(lambda_t . x_d). `features` is required: an (num_docs, F)
+        covariate matrix (no intercept column — one is prepended), with
+        feature_names naming the F columns. The L-BFGS optimization of lambda runs
+        every optimize_interval sweeps after burn_in; topic-word phi is averaged
+        over num_samples samples taken every sample_interval sweeps."""
         ...
 
     @property
@@ -216,15 +219,15 @@ class DMR:
         ...
 
     @property
-    def prevalence_effects(self) -> numpy.typing.NDArray[numpy.float64]:
-        """Learned prevalence weights gamma, shape (num_topics, num_features).
-        Column 0 is the intercept; positive entries raise that topic's
-        prevalence."""
+    def feature_effects(self) -> numpy.typing.NDArray[numpy.float64]:
+        """Learned feature weights lambda, shape (num_topics, num_features). Column
+        0 is the intercept; positive entries raise that topic's prevalence as the
+        feature increases."""
         ...
 
     @property
     def feature_names(self) -> list[str]:
-        """Feature names aligned with prevalence_effects columns ('intercept' first)."""
+        """Feature names aligned with feature_effects columns ('intercept' first)."""
         ...
 
     @property
--- a/python/topica/analysis.py
+++ b/python/topica/analysis.py
@@ -4,7 +4,10 @@ These helpers read only a fitted model's public attributes — ``topic_word``
 (K x V), ``doc_topic`` (D x K), ``topic_names``, ``vocabulary``, ``num_topics``,
 and the optional ``labels`` (hard document assignments on the embedding-cluster
 models, where ``-1`` marks a noise/outlier document) — so they work uniformly
-across LDA, STM, CTM, keyATM, Top2Vec, BERTopic, and the rest. The goal is the
+across LDA, STM, CTM, keyATM, Top2Vec, BERTopic, and the rest. SAGE's
+``(K, G, V)`` topic-word is reduced to its group marginal; DTM (time-sliced
+``topic_word(time)``) and HLDA (a topic tree, no ``doc_topic``) do not present
+this static surface and are not supported here. The goal is the
 overview a researcher reaches for first: how big each topic is, what it is about,
 who its representative documents are, and how its prevalence moves across time or
 across groups.
@@ -19,6 +22,8 @@ across groups.
 """
 
 from __future__ import annotations
+
+import warnings
 
 import numpy as np
 
@@ -122,8 +127,13 @@ def _top_words(model, t, n):
         try:
             pairs = method(n, topic=t)
             return [w for w, _ in pairs]
-        except Exception:
-            pass
+        except Exception as exc:
+            warnings.warn(
+                f"{type(model).__name__}.top_words failed ({type(exc).__name__}: "
+                f"{exc}); falling back to raw topic-word rows, which drops any "
+                "custom weighting (e.g. FREX) that top_words applies.",
+                stacklevel=2,
+            )
     phi = np.asarray(model.topic_word, dtype=np.float64)
     vocab = list(model.vocabulary)
     idx = np.argsort(phi[t])[::-1][:n]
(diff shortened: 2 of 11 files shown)


nealcaren:
⚠️ **Scope note / base correction.** While rebasing this branch I found that **#21 never reached \`main\`**: PR #23 was merged into the \`fix/keyatm-seeded-alpha-dirichlet\` branch rather than \`main\`, and #22 had already merged that branch (with only #20) into \`main\`. Verified against \`origin/main\`: it has the #20 alpha getters (KeyATM/SeededLDA) but **not** #21's (LabeledLDA/SupervisedLDA/DMR/PA/PT/SAGE) or the registry test.

So this PR is now rebased onto current \`main\` and carries **both #21 and #24–#27**. The two commits are:
- \`#21\` — expose alpha across the Dirichlet Gibbs models + registry family test (re-landed)
- \`#24–#27\` — the silent-degradation / API-drift sweep

`#20` is already in \`main\` and was dropped from the branch during rebase. Reopened #21 to track.

Issue #260, opened 2026-06-20
Cross-language FREX parity: topica_core::inspect vs topica.validation.frex (single source of truth)


nealcaren:
## Background

#259 adds `topica_core::inspect` (FREX / lift / score / semantic coherence / exclusivity), a Rust port of stm's `calcfrex`/`calclift`/`calcscore`/`semCoh1beta`/`exclusivity` for **stmata** and faSTM. topica's Python layer still has its own FREX in `python/topica/validation.py` (`frex`, used by `label_topics`).

So the repo now has **two independent stm ports of FREX** (Rust `inspect::frex_scores` and Python `validation.frex`). They are not wired together — no speed impact (neither is in a timed path; topica's fitting path is unchanged), but they can **drift**.

## Ask

1. **Parity test (first cut):** a check that `topica_core::inspect::frex_scores` (and lift/score/exclusivity/semantic_coherence) agree with `topica.validation.frex` (and friends) on poliblog, to tolerance. Guards against drift today.
2. **Single source of truth (later):** consider routing topica's Python FREX through the Rust `inspect` module via a binding, so there is one definition the three host languages (R/Python/Stata) share. Decide whether the small per-call cost of crossing the boundary is acceptable (FREX is a cheap post-hoc argsort, so almost certainly yes).

## Notes

- faSTM also FREXes in R; ideally all three converge on `topica_core::inspect`.
- The `w` (frequency/exclusivity weight, default 0.5) and tie-handling (average ranks) must match across implementations — that's the most likely drift point.

nealcaren:
Part 1 (parity test) done in #262. Concrete findings comparing `topica_core::inspect` (stm-faithful) vs `topica.validation`:

- **FREX** (no James-Stein shrinkage): **bit-for-bit identical**. ✅
- **score**: identical on dense beta; selected top words identical on sparse beta (the only diffs are negligible-probability words, from the differing log floors — Rust `f64::EPSILON` vs Python `1e-12`). ✅
- **FREX shrinkage**: Rust-only stm option; Python has no-shrink only. (Python could grow a `wordcounts=`/shrink option if we want stm's default.)
- **lift**: ❌ **different definitions** — Python `label_topics` lift = `beta / mean_k beta` (mean-beta ratio); stm/Rust = `log(beta) − log(empirical word freq)`. Captured as a strict xfail.

Part 2 (single source of truth) decisions left open:
[...]

PR #263, opened 2026-06-20
Unify FREX/lift/score through topica-core; fix lift to stm's definition (#260)

nealcaren:
Completes #260 part 2: make topica-core's stm-faithful Rust `inspect` the **single definition** of FREX / lift / score. Python's `frex` and `label_topics` now wrap the core (via the `_topica.inspect_*` bindings from #262) instead of reimplementing the math in numpy, so topica, faSTM, and stmata can't drift.

## Behavior changes

- **lift is now stm's lift** — `log P(w|topic) − log P(w)`, not the old `beta / mean_k beta` ratio (which wasn't lift at all). `label_topics(..., word_counts=)` gives the exact value from corpus word frequencies; without counts, `P(w)` is estimated from the column marginal, so the result is `log(beta) − log(marginal)` — **same ranking as before, now on the correct log scale**.
- **FREX gains stm's James-Stein shrinkage** via `frex(..., word_counts=)` / `label_topics(..., word_counts=)` (stm's default — damps rare words that land in one topic by chance). Without counts, the no-shrink path is **bit-identical** to before. (Shrinkage strength depends on count magnitude, so it's only applied with real counts, never faked.)
- frex/score with no `word_counts` keep the same selected words; the numpy `_ecdf_ranks` helper is deleted (the core supplies ranks now).

## Why `word_counts=` rather than shrink-by-default

Both stm-faithful lift and FREX shrinkage need empirical word frequencies, which fitted models don't carry. So they activate when you pass `word_counts=` (or, for lift, fall back to the marginal estimate). A `corpus=` convenience that derives the counts automatically would be a natural follow-up.

## Validation

- The **R-stm parity test** (`calcfrex` / `labelTopics`) still passes — frex/score remain stm-faithful after rerouting.
- `tests/test_frex_cross_lang.py` now asserts the Python wrappers equal the core exactly; the old strict lift xfail is a **passing equality**.
- Full suite: 2798 passed, 27 skipped. No Rust changed (pure Python + the existing bindings).

## Note
This is output-affecting (lift values change scale; FREX with `word_counts` changes words) — the API-freeze override you approved.

### DIFF
--- a/docs/guides/diagnostics.md
+++ b/docs/guides/diagnostics.md
@@ -42,6 +42,7 @@ embeddings. Both range over `[0, 1]`, and higher means more diverse.
 
 ```python
 topica.label_topics(model.topic_word, model.vocabulary, n=10)   # prob / frex / lift / score
+topica.label_topics(model.topic_word, corpus=corpus, n=10)     # stm-faithful: lift + FREX James-Stein shrinkage from corpus word counts
 topica.frex(model.topic_word, model.vocabulary, n=10)           # frequent + exclusive
 topica.relevance(model.topic_word, model.vocabulary, lam=0.6)   # LDAvis relevance
 topica.find_thoughts(model.doc_topic, texts, topic=0, n=3)      # representative docs
--- a/python/topica/_topica.pyi
+++ b/python/topica/_topica.pyi
@@ -159,6 +159,12 @@ class Corpus:
         ...
 
     @property
+    def word_counts(self) -> list[int]:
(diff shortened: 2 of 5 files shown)


nealcaren:
Added the `corpus=` convenience (the follow-up mentioned above) to this PR: `frex`/`label_topics` now accept `corpus=` and read its word frequencies for stm-faithful lift + FREX shrinkage, so the stm-faithful path is zero-effort. New `Corpus.word_counts` getter backs it. Full suite 2799 passed; docs build clean.

PR #291, opened 2026-06-24
fix(anchor): stop RecoverKL diverging to NaN on sparse co-occurrence

nealcaren:
## What

Found while benchmarking AnchorLDA on 20 Newsgroups: the model produced **degenerate topics (diversity 0.05, NMI 0.00)** with a `RuntimeWarning: invalid value encountered in divide`.

## Root cause

On a large-vocabulary corpus with sparse word co-occurrence (20NG, V~6k), some vocabulary words never co-occur with any selected anchor. Their reconstruction `recon = C·Q_anchors` hits the `1e-12` floor, so the exponentiated-gradient term `q/recon` explodes (`|upd| ~ 1e8`). After the row-max shift, every entry but the maximum underflows to **exactly 0**, and the simplex renormalization then divides `0/0` → `NaN`, which propagates and collapses every topic. poliblog's denser co-occurrence (V~2.6k) never triggered it.

## Fix

Floor the post-shift exponent at `-50` (`exp(-50) ≈ 2e-22`, negligible) so a row can never fully underflow. It's bounded and finite, and a no-op where the update is already well-scaled:
- **poliblog c_v bit-unchanged** (0.482 with and without the floor).
- **20NG**: recovery now yields valid, diverse topics (no NaN, diversity ~0.85, sensible clusters like `card/scsi/driver` and `marriage/pope/church`).

## Test

Adds a regression test on a sparse-co-occurrence corpus (short docs, large vocab) asserting `topic_word`/`doc_topic` stay finite and topics are non-degenerate.

## Note

This is a robustness fix, not a performance claim — the 20NG benchmark also showed AnchorLDA is *not* competitive there (NMF and LDA win on coherence, label-recovery, and speed once the vocabulary is large). AnchorLDA's honest niche remains a fast, deterministic first pass on small-to-moderate vocabularies. The fix just ensures it returns valid topics everywhere instead of NaN.


nealcaren:
**Scope expanded: also fixes anchor *quality* on large corpora (second commit).**

The 20NG failure had two causes; the NaN was only the first.

**2. Anchors were rare, noisy words.** The farthest-point search picked low-document-frequency words as anchors (df≈10 on 20NG: `cpr/typed/bureau`), giving weak topics. Added `anchor_min_doc_freq` (default `0.01` = 1% of docs; `<1` fraction, `≥1` absolute, `0` off; falls back to all words if too few qualify) — the standard Arora et al./Lee–Mimno refinement.

Combined effect (real model, FREX-ranked c_v):

| | poliblog K=50 | 20NG K=20 |
|---|---|---|
| before | 0.491 (NaN-prone) | **NaN (div 0.05)** |
| after | 0.491 / div 0.89 | **0.433 / div 0.98** |

AnchorLDA on 20NG now yields valid, highly distinct topics. It still trails LDA (~0.63) and NMF (~0.87) on coherence there — anchor-words trades coherence for speed/determinism by design — but it behaves correctly at scale now.

Also tested and **dropped**: random projection of Q rows for anchor finding — no coherence gain (0.415–0.447 vs 0.445), speed-only at very large vocab.

Tests for both fixes added; 574 pass, docs clean.

PR #332, opened 2026-07-13
feat: implement Online Tensor LDA (TensorLDA) topic model

nealcaren:
## Overview

This PR adds **Online Tensor LDA (`TensorLDA`)** to topica: a seed-reproducible, count-based method-of-moments topic model using second- and third-order cumulants. It implements the published method from Kangaslahti et al. (2026), *Political Analysis*, and is based on the upstream [TensorLy TLDA](https://github.com/tensorly/tlda) project.

## Status

`TensorLDA` is intentionally **experimental** and requires `topica.enable_experimental(True)`. It is suitable for exploratory topic-word recovery, with the usual coherence, stability, and human-validation checks. Its `weights` should be treated as model diagnostics, **not** as calibrated substantive topic prevalence or covariate-effect estimates.

## Included

- Rust implementation of whitening, factorized third-order cumulant optimization, and document-topic prediction.
- PyO3 bindings, serialization, type stubs, registry, model-invariant coverage, and deterministic fitting.
- Input validation for invalid ranks, corpus size, batch size, iteration counts, and numerical parameters.
- Optional `theta` and `n_eigenvec`; rectangular whitening ranks return exactly `num_topics` weights.
- `unwhitened_raw` inspection property.
- Documentation, citation, acknowledgement, and a dedicated TensorLDA validation record.
- Portable known-truth LDA benchmark: `python parity/tlda_true_lda_compare.py`.

## Validation boundary

- On a matched square-rank UCI comparison, topic-word cosine agreement with the upstream implementation is near exact.
- The upstream example's `n_topic=2, n_eigenvec=10` configuration returns ten weights for two topics, so it is not a valid weight-parity target.
- Known-truth LDA and realistic synthetic-corpus benchmarks are documented and reproducible. They show that further work is needed before interpreting weights as calibrated prevalence or removing the experimental gate.

## Verification

- `cargo test --lib tlda`
- TensorLDA Python integration and gold tests
- `cargo check --features python`
- `cargo clippy --lib -- -D warnings`
- Documentation build
- `python parity/tlda_true_lda_compare.py`


nealcaren:
## Review

I found three correctness issues to address before merging:

1. **P1 — small corpora can panic.** `fit()` permits fewer documents than topics. `randomized_svd()` then truncates to `min(rows, cols, k)`, while `fit_tlda()` indexes the result through `num_topics`. For example, `TensorLDA(2).fit([["a", "b"]])` passes the vocabulary check and then panics. Reject `num_docs < num_topics` before fitting (or handle truncated rank).

2. **P1 — `batch_size=0` can panic.** The constructor accepts zero, then training reaches `step_by(batch_size)`. Validate `batch_size > 0` in the constructor; validating the remaining numeric parameters there would also prevent invalid numerical states.

3. **P2 — `iters` is off by one.** The loop condition is `i < n_iter_train`, so `iters=1` makes zero training updates and every positive request executes one fewer pass than requested. Use an inclusive / zero-based bound.

Validation completed: `cargo test --lib tlda` (2 passed) and `cargo clippy --lib -- -D warnings` both pass.

## Reference-package check on `plot_uci_topics.py`

I ran the UCI Newsgroups data preparation from `/Users/nealcaren/Downloads/plot_uci_topics.py` without plotting and compared topic-word distributions after one-to-one topic alignment.

- With the script **as written** (`theta=5.005`, `n_eigenvec=10`), parity is not expected or obtained: matched cosines were **0.128** and **0.952** (mean **0.540**). The reference package also yields two topic-word distributions but **ten weights** in this configuration, so its weights are internally incompatible with the two-topic setup.
- With the reference rerun under the settings this implementation hard-codes (`theta=1`, whitening rank `2`; matching seed, iterations, learning rate, and smoothing), the results are effectively identical: matched cosines **0.999997** and **0.999974** (mean **0.999986**); max / mean absolute topic-word differences **0.001225** / **0.0000478**. Recovered weights are both approximately **0.005**, up to topic permutation.

So the Rust implementation reproduces the reference under equivalent settings. Supporting the example script exactly would require exposing `theta` and a separate whitening-rank / `n_eigenvec` parameter.

nealcaren:
## Recommendation: make rectangular-rank support a Topica improvement

The new `n_eigenvec` option is useful, but I recommend finishing it with a principled **K-topic weight-recovery rule** for the rectangular factor matrix (`n_eigenvec × num_topics`).

The reference package is not the right target for this particular detail: when `n_topic=2, n_eigenvec=10`, it returns two topic-word distributions but ten `weights_`, because it computes a norm per whitening-coordinate row. Those values are not a valid two-topic prevalence vector and cannot correctly feed its document-topic prediction routine.

Topica should instead use the extra whitening dimensions to improve the approximation while still recovering exactly `num_topics` weights. The current approach—deriving the K weights from only the first K rows—depends on arbitrary whitening-coordinate ordering and discards the other rows.

Suggested acceptance criteria:

- derive/document a K-length recovery rule for an `R × K` factor matrix (`R = n_eigenvec`);
- verify it is invariant to equivalent permutations/rotations of the whitening basis;
- add a synthetic corpus where true topic prevalences are deliberately unequal, and check that recovered weights track them for both `R=K` and `R>K`;
- confirm `weights.shape == (K,)` and that transform/prediction remains finite and stable.

That would make the rectangular-rank path a genuine improvement over the original package rather than reproducing its shape bug.

Issue #404, opened 2026-07-22
Manifest integrity/signing: stdlib HMAC tamper-evidence, and NOT public-key signing


nealcaren:
Analysis-manifest V2 (#394), split from the bundle work (#403). Lower priority; records a scoping decision as much as a feature.

### The fork
Signing a manifest/bundle has two very different forms:

- **HMAC (symmetric / keyed)** — sign the canonical manifest with a shared secret. Cheap, **stdlib** (reuses the BLAKE2b keyed mode already used for keyed fingerprints — nearly free). Proves tamper-evidence to anyone who holds the same secret, i.e. *internal* verification. Does not prove "this came from person X" to the public.
- **Asymmetric (public-key, e.g. ed25519)** — sign with a private key, anyone verifies with the public key. The genuinely "publicly attestable" version, but it needs a crypto dependency (topica is deliberately dependency-light) **and** key management (storage, distribution, revocation), which is outside a topic-modeling library's remit.

### Decision / scope
- **In scope (if/when desired)**: an optional HMAC tamper-evidence field over the canonical manifest JSON, reusing the keyed-fingerprint machinery. No new dependency. Clearly documented as symmetric/private, not public attestation (consistent with the existing keyed-fingerprint `keyed: true` flag).
- **Explicitly NOT building**: public-key signing / a PKI inside topica. Public attestation is better delegated to standard tooling researchers already use — git commit signing, `sha256sum` + PGP, Sigstore, OSF/Zenodo DOIs. Sign the bundle file (#403) with whatever the institution uses.

### Honesty caveat
A signature proves **who** and **integrity**, never **correctness** — a signed manifest of a bad analysis is still a bad analysis. This matches the manifest's non-goals (it records evidence, it does not vouch for conclusions).

Deferred; open for discussion. #403 (the bundle) does not depend on this.

nealcaren:
Closing as a documented scoping decision (the issue was filed to record the fork as much as to request a feature).

**Not building:** public-key signing / a PKI inside topica. Public attestation is better delegated to the standard tooling researchers already use — git commit signing, `sha256sum` + PGP, Sigstore, OSF/Zenodo DOIs. Sign the bundle file with whatever the institution uses.

[...]

Issue #394, opened 2026-07-22
Proposal: add a privacy-preserving analysis manifest (“flight recorder”)


nealcaren:
## Problem

Topica already makes individual fits unusually reproducible: models and corpora serialize, seeds are documented, diagnostics/reporting exist, and several fits are bit-exact. But a publishable analysis is larger than a fitted model. The corpus transformation, metadata/design matrix, fit arguments, external embeddings, diagnostics actually run, and interpretive decisions are usually scattered across notebooks. A saved `.tt` can therefore be reproducible without making the *analysis* auditable.

## Proposal

Add a lightweight, versioned **analysis manifest**: a JSON artifact that records the inputs and decisions needed to inspect or replay one fit, plus a verifier. “Flight recorder” is the metaphor; the public API should use precise names such as `AnalysisManifest` / `record_fit`.

This is explicitly **not** a second model serialization format, an automatic claim that preprocessing was captured, or a dashboard rewrite. It should compose with existing `Corpus.save`, `model.save`, diagnostics, and `plot_report`.

```python
corpus = topica.Corpus.from_documents(tokens, metadata=meta)
model = topica.STM(num_topics=20, seed=42)

record = topica.record_fit(
    model, corpus,
    prevalence=X, prevalence_names=names, iters=1000,
    privacy="aggregate",
)
record.add_diagnostic("stability", topica.bootstrap_stability(...))
record.add_decision("K", "Researcher chose 20 after inspecting 15/20/25.")
record.save("analysis.topica.json")
record.render("analysis.html")      # optional, later

# On another machine/session:
record = topica.AnalysisManifest.load("analysis.topica.json")
record.verify(corpus, model)          # reports exact/mismatched/unverifiable fields
```

The explicit `record_fit(...)` call is deliberate. A context manager or monkey-patching `.fit()` cannot honestly see preprocessing performed before the call, and “automatic provenance” that silently misses it is worse than an explicit boundary.

## V1 manifest contents

- schema and Topica version; Python/platform/architecture; thread count and determinism classification
- model class, serializable constructor settings, fit settings, seed, convergence/status, and hashes of model outputs
- corpus *description*: document count/length summary, vocabulary size, preprocessing parameters supplied to Topica, and an order-sensitive corpus fingerprint
- metadata/design-matrix schema, row count, column names/dtypes, and fingerprints—not values
- fingerprints and provenance fields for supplied embeddings, seed dictionaries, labels, timestamps, etc.
[...]

PR #405, opened 2026-07-22
feat(manifest): content-addressed analysis bundle (#394 V2)

### LINKED ISSUE #403: Content-addressed analysis bundle (manifest + artifacts as one verifiable unit)
Analysis-manifest V2 (#394). The manifest is a JSON file of fingerprints: it proves identity but does not carry the artifacts, so reproducing a fit needs the `.tt` files shipped separately with nothing tying them to the manifest. (V1 does not implement the "reference saved model/corpus by path + digest" idea either — `record_fit` never took a `model_path`/`corpus_path`.)

### Proposal
Package the manifest and the actual saved artifacts into one **content-addressed** bundle — a single shareable, self-verifying unit.

```python
record.bundle("analysis.zip", include_model=True, include_corpus=False)
```

produces (stdlib `zipfile`, no new dependency):

```
manifest.json                 # the AnalysisManifest
artifacts/<digest>.tt         # model.save(), file named by its content hash
artifacts/<digest>.tt         # corpus.save() (opt-in, sensitive)
```

- The manifest's model / corpus / inputs blocks gain artifact references `{"path": "artifacts/<digest>.tt", "digest": "<digest>", "algo": "blake2b-256"}`.
- **Content-addressed**: each artifact's filename IS its digest, so the bundle self-verifies — `AnalysisManifest.load_bundle(path)` recomputes each file's digest, checks it against its name and the manifest, and can `verify` / replay directly off the bundled artifacts.

### Design points
- **Format**: default to a `.zip` (one file to share); optionally a plain directory.
- **Inclusion**: manifest always; `include_model` opt-in; `include_corpus` opt-in and **sensitive** (a bundled corpus `.tt` embeds raw tokens — same privacy tension as the content fingerprint, so off by default with the same warning).
- **Two distinct digests, kept separate**: the *file* digest ("this exact saved file") vs. the *content* fingerprint ("the same corpus regardless of save-format version"). Both recorded; not conflated.
- **Bundle-layout version**, alongside the manifest schema and fingerprint spec.
- Reuses the fingerprint-v1 BLAKE2b hashing; pure Python.

### Non-goals
Signing/attestation is tracked separately (see the signing issue). The bundle proves artifact identity and integrity, not authorship.

nealcaren:
Closes #403. Next V2 increment (after V1 #395, render #396, compare #397, diagnostics #398).

## What
`AnalysisManifest.bundle(path, model=model, corpus=corpus)` packages the manifest and the saved artifacts into one self-contained, **content-addressed** `.zip` — a single shareable, self-verifying unit.

```python
record.bundle("analysis.zip", model=model)                                  # manifest + model
record.bundle("analysis.zip", model=model, corpus=corpus, include_corpus=True)
```
produces (stdlib `zipfile`, no dependency):
```
manifest.json            the manifest, plus artifact references
artifacts/<digest>.tt    model.save() / corpus.save(), file named by its hash
```

## Content-addressed + tamper-evident
Each artifact's **file name is its BLAKE2b digest**, and the manifest's model/corpus blocks gain an `artifact` reference. `AnalysisManifest.load_bundle(path)` recomputes every artifact's digest and checks it against *both* the content-addressed name and the reference, **raising on a corrupt or tampered bundle** (tested: flipping one byte is caught). `extract_bundle(path, dest)` writes the artifacts out so you can `topica.LDA.load(...)` them.

## Design points
- Since the manifest holds only fingerprints (not the objects), you **pass the fitted `model`/`corpus`** to `bundle()`, mirroring `verify(corpus, model)`.
- **`include_corpus` opt-in and sensitive** — a bundled corpus embeds raw tokens; off by default, flagged `sensitive: true`.
- The **file digest** ("this exact saved file") is kept **distinct from the content fingerprint** ("same corpus across save-format versions"); both are recorded.
- `BUNDLE_VERSION` guards the layout; an unknown version is rejected on load.

## Tests (`+8`, 56 total in the file)
Content-addressed layout, round-trip + `verify`, tamper detection, corpus opt-in + sensitive flag, requires-the-object error, extract + reload (reloaded `topic_word` matches), file-digest-vs-content-fingerprint distinct, unknown-version rejection.

## Gates
- Full pytest suite: **3368 passed, 30 skipped**.
- `mkdocs build --strict` clean; preflight green. Pure Python.

## V2 status
render (#396), compare (#397), diagnostics (#398) merged; this closes the bundle (#403). The only remaining manifest item is signing/tamper-evidence (#404), deferred by design.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,12 +9,15 @@ follow [Semantic Versioning](https://semver.org/spec/v2.0.0.html) once released.
 ### Added
 
 - **Content-addressed analysis bundle**: `AnalysisManifest.bundle(path, model=,
-  corpus=)` packages the manifest and the saved artifacts into one self-verifying
-  `.zip` — each artifact file is named by its BLAKE2b digest, and
-  `AnalysisManifest.load_bundle` verifies every artifact against its name and the
-  manifest reference (raising on a corrupt or tampered bundle).
-  `extract_bundle(path, dest)` recovers the artifacts for reloading. Bundling the
-  corpus is opt-in and sensitive (it embeds raw tokens).
+  corpus=)` packages the manifest and the saved artifacts into one `.zip` — each
+  artifact file is named by its BLAKE2b digest. `bundle` refuses to package a
+  model/corpus whose fingerprints do not match the manifest (the wrong fit), and
+  `AnalysisManifest.load_bundle` checks every artifact against its content-addressed
+  name and the manifest reference, raising on a corrupt bundle or artifacts that no
+  longer match. This is an integrity/content-addressing guarantee, not authenticity
+  (detecting a fully rewritten bundle needs a signature). `extract_bundle(path,
+  dest)` recovers the artifacts for reloading. Bundling the corpus is opt-in and
+  sensitive (it embeds raw tokens).
 - **Built-in diagnostic capture** for the manifest:
   `record_fit(..., diagnostics=["coherence", "exclusivity"])` computes those
   topic-quality metrics and records each as computed evidence (mean over topics),
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,13 @@ follow [Semantic Versioning](https://semver.org/spec/v2.0.0.html) once released.
 
 ### Added
 
+- **Content-addressed analysis bundle**: `AnalysisManifest.bundle(path, model=,
+  corpus=)` packages the manifest and the saved artifacts into one self-verifying
+  `.zip` — each artifact file is named by its BLAKE2b digest, and
+  `AnalysisManifest.load_bundle` verifies every artifact against its name and the
+  manifest reference (raising on a corrupt or tampered bundle).
+  `extract_bundle(path, dest)` recovers the artifacts for reloading. Bundling the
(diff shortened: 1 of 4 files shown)


nealcaren:
Follow-up: ran an independent Codex review of this PR, which caught a real issue and some hardening. Applied in the latest commit:

- **Softened the over-claim** — the bundle provides **integrity / content-addressing**, not authenticity. Reworded the docstrings, CHANGELOG, and docs away from "tamper-evident": content-addressing detects a corrupt bundle or artifacts that no longer match the manifest, but *not* a fully rewritten bundle (artifact + digest + reference all changed together), which needs a signature (#404). Not over-claiming is the whole point of the manifest, so this framing mattered.
- **Bound artifacts to the manifest** — `bundle()` now runs `verify()` on the supplied model/corpus and refuses to package a fit whose fingerprints don't match what the manifest recorded (catches accidentally bundling the wrong model).
- **Shared version validation** — `extract_bundle` now validates the bundle version like `load_bundle` (was inconsistent).
- **Exact artifact-path + algorithm check** in `_verify_artifact`.

New tests for wrong-model rejection and extract version-check. Full suite 3370 passed. The zip-bomb exposure Codex noted is left as a documented trust-model assumption (these are collaborator reproducibility bundles, not arbitrary untrusted input), not a blocker.

Issue #421, opened 2026-07-23
STM/CTM core review: content recompute_nu uses shared beta; empty-doc uniform theta; non-PD Sigma bound; SVI suff-stat/Sigma/shrinkage bugs


nealcaren:
Independent correctness review of the STM/CTM core (`topica-core/src/ctm.rs`), via the Codex second-opinion agent (`gpt-5.6-sol`), source-only, with the concrete findings re-verified against the code. **The full-batch, no-content variational-EM path — the R stm drop-in — is mathematically sound**: `ctm_lhood`/`ctm_grad`, the `hpbcpp` Hessian, `nu = H^-1`, `det_term = +0.5 log|nu|`, and the M-step for mu/Sigma/beta all match R stm. The reference category is the last topic (K-1) and the code is internally consistent about it. There is no fixed 1e-6 Sigma ridge (that earlier premise was stale; prevalence now uses adaptive `fit_gamma_enet`). Real bugs are in the content `recompute_nu`, empty-doc inference, the non-PD Sigma bound, and SVI.

## 1. [High, narrow] `recompute_nu` uses the shared beta for content models

The E-step infers each content document with its **group-specific** `content_beta[g[di]]` (`ctm.rs:1263`), but the saved snapshot is the **shared** `beta_estep = beta.clone()` (`ctm.rs:1222`). `recompute_nu` then applies that shared beta to every document (`ctm.rs:1698`).

For content fits with `keep_nu=false`, reconstructed covariances do not reproduce the fitted Hessians, so `estimate_effect` method-of-composition draws (which use `eta_cov`) can be materially wrong when content groups have strongly different word distributions. Non-content fits are unaffected (there `beta_doc == beta_estep`).

**Fix:** preserve the E-step group-specific beta set + each doc's group index, or reconstruct the correct group beta inside `recompute_nu`.

## 2. [Medium] Empty/OOV-only documents get uniform theta instead of the prior

`infer_theta` returns `vec![1.0/k; k]` for `words.is_empty()` (`ctm.rs:823`). With no observations the posterior mode is `eta = mu`, so it should return `softmax([mu, 0])`. This is wrong specifically for prevalence models where `mu` carries a document-specific `X.gamma`.

## 3. [Medium] Non-PD Sigma fallback makes objective and bound inconsistent

[...]

PR #442, opened 2026-07-23
fix(stm): recompute per-document ν against each group's β for content models (#421)

nealcaren:
## Summary

Fixes the content-`recompute_nu` item of #421. `recompute_nu` regenerates the variational covariances ν on demand when a fit was run with `keep_nu=false` — the path `posterior_theta_samples` / `estimate_effect` use for **method-of-composition** standard errors. It rebuilt every document's ν from the shared, group-averaged `beta_estep`.

But a **content model**'s E-step runs each document against its own group's topic-word distribution `content_beta[group[d]]`. So for content STM the reconstructed ν — and therefore the composition draws and SEs that consume it — were **materially wrong** whenever the content groups word the topics differently.

## Fix

- Track the per-document group index: new `CtmModel.groups`, stored by `fit_ctm`, persisted on the STM Python state and through `save`/`load` (serde-default, so old saves load fine).
- In `recompute_nu`, select `content_beta[groups[di]]` per document; fall back to `beta_estep` for non-content models (and for models saved before this was tracked).

`content_beta` is the final M-step value, so the recomputed ν differs from the stored one only by that last content update — far smaller than the group-vs-average error it replaces, and well below the Monte-Carlo noise of the draws that use ν.

## Verification

- **Core test** (no Python needed): fit a two-group content model with `keep_nu=true`, then assert the per-group `recompute_nu` matches the stored ν closely (mean L1 < 1e-2) while the old group-averaged path is **>20× worse**.
- Full `topica-core` ctm suite (13) green; default lib suite (183) green.
- `cargo check --features python` and `cargo clippy --features python` clean (the binding + save/load wiring type-checks with no warnings).

## Scope

`topica-core/src/ctm.rs` (core + test) and the STM/CTM region of `src/python/mod.rs` (new `groups` field on the STM state + stub + save/load) — disjoint from the SAGE region. No public Python API change. The non-PD-Σ and SVI items of #421 remain follow-ups.


nealcaren:
## Review status: needs changes before merge

Two codex review rounds confirmed the core fix is correct, but a final pass (high-effort) found two follow-up issues, and the branch now conflicts with `main` after #445/#446 landed. Summarizing what's left so this can be finished cleanly.

**Confirmed resolved**
- Fresh content fits (core `topica-core` and the Python `STM` object) reproduce the stored ν *exactly* — max diff `0` in both Laplace and diagonal modes. The `em_iters=1` core test and the parametrized `test_stm_content.py::TestSTMContentRecomputeEtaCov` are genuinely discriminating (pre-snapshot proxy differs by ~0.11, past `atol=1e-5`).
- Save-format documentation is accurate: appending `groups` to the positional-bincode `StmState` breaks loading of **all** earlier STM saves (content, prevalence-only, plain); format version stays `1`, so it surfaces as an invalid-file/EOF error (holistic fix tracked in #443).

**Blocking — must fix**
1. **Stale `content_beta_estep` across a refit (correctness).** In the STM `fit`, the snapshot is assigned only inside `if !keep_eta_cov` and never cleared. Fitting the same object `keep_eta_cov=False` then refitting `keep_eta_cov=True` leaves the old `Some(...)`, and `_recompute_eta_cov` then prefers the stale group β (codex reproduced a max covariance error of ~1.80 with swapped group labels). Fix: assign the snapshot **unconditionally** (to `None` when `keep_eta_cov=True`). The same latent bug exists for `beta_estep`/`sigma_estep` — clear all three in one place.

**Non-blocking — worth doing in the same pass**
2. **Unnecessary `G×K×V` snapshot on the default `keep_nu=True` path (perf).** The core clones `content_beta` every EM iteration and returns `Some` for every content fit regardless of `keep_nu`; the Python wrapper discards it for a fresh `keep_eta_cov=True` fit. Gate the core capture/return on `content.is_some() && !keep_nu` so the default path doesn't pay the copy + peak allocation.

**Must-add test**
- A same-instance `keep_eta_cov=False → True` refit test (swapped group labels expose it) asserting the recompute uses the current fit's β, not a stale snapshot.
- Optional: tighten the Python assertion from `assert_allclose(atol=1e-5)` to `assert_array_equal`, since the guarantee is exact f32 reproduction.

**Merge mechanics**
- Rebase onto `main` (now includes #445, which replaced the `spd_inverse`/`half_logdet` blocks in `recompute_nu` and the E-steps with `spd_inverse_and_half_logdet`, and #446). The conflict is in `recompute_nu` / the fit E-step siginv+entropy region — keep #445's `spd_inverse_and_half_logdet` call and layer the `content_beta_estep` selection on top.

nealcaren:
**Maintainer note:** the save-format break (blocker 2) is deprioritized to v1.0 and **non-blocking** — just keep the CHANGELOG note honest. The remaining must-fix is the correctness item: the Python `content_beta_estep` is assigned only under `if !keep_eta_cov` and never cleared, so a `keep_eta_cov=False` → `True` refit on the same object uses a stale snapshot. Assign it (and ideally `beta_estep`/`sigma_estep`) unconditionally. Plus rebase onto main (now has #445's `spd_inverse_and_half_logdet` in `recompute_nu`).

nealcaren:
## Re-review: snapshot-clear + rebase look right; two of my own core tests now fail CI

The `keep_eta_cov`/`content_beta_estep` clear-on-refit fix and the rebase over #445/#446 are correct — `recompute_nu` cleanly combines #445's `spd_inverse_and_half_logdet` with the per-group `content_beta_estep` selection, and the Python object now assigns the snapshot unconditionally (`None` on `keep_eta_cov=True`, the snapshot on `False`). Nice.

**But `rust tests (ubuntu)` fails** on two `topica-core` tests I originally wrote:
- `recompute_nu_uses_estep_content_beta_not_final_mstep` (panics at `content_beta_estep.is_some()`)
- `recompute_nu_uses_per_group_beta_for_content` (`err_fixed < 1e-9` fails)

**Root cause — good optimization, stale tests.** This PR now (correctly, per codex's earlier finding) gates the core capture: `let capture_content_beta = content.is_some() && !keep_nu;` (`ctm.rs:1248`). So the snapshot is captured only when it'll actually be used — i.e. `keep_nu=false`. But both of my tests fit with `keep_nu=true` (to get a stored `model.nu` to compare against) **and** assert the snapshot exists / reproduces exactly. With `keep_nu=true` the snapshot is now (intentionally) `None`, so the assertions fail. The feature and the optimization are both right; my tests conflate "store ν" with "capture snapshot," which can no longer both be true in one fit.

**Fix — two-fit design (keeps the optimization).** Since same-seed fits are bit-identical, use one fit for the ground-truth ν and a second for the snapshot:
```rust
// ground truth: keep_nu=true stores the E-step ν
let m_keep = fit_ctm(&docs, ..., /*keep_nu=*/ true, ...);
let stored = m_keep.nu.clone();
// recompute source: keep_nu=false captures content_beta_estep
let m_drop = fit_ctm(&docs, ..., /*keep_nu=*/ false, ...);   // same seed/args
assert!(m_drop.content_beta_estep.is_some());
let recomputed = recompute_nu(&m_drop, &sparse);
// exact: m_drop's snapshot == the β the E-step ran against == m_keep's stored ν
assert!(mean_l1(&stored, &recomputed) < 1e-9);
// anti-vacuity (per-group test): clearing the snapshot -> content_beta fallback is >20x worse
```
For the `not_final_mstep` (em_iters=1) test, keep the "final M-step moved β" guard by reading `m_drop.content_beta_estep` vs `m_drop.content_beta`. This preserves both tests' discriminating power while respecting the `!keep_nu` gate.

Everything else is green/pending; this is the only failure. Once the two tests are updated, this should be mergeable.

PR #447, opened 2026-07-23
fix(lda): stop turbo mode from skipping scheduled bookkeeping (#417)

### LINKED ISSUE #417: LDA core review: turbo-mode bookkeeping skip + packed-count u32 overflow ceiling
Independent correctness review of the core SparseLDA sampler (via the Codex second-opinion agent, `gpt-5.6-sol`), cross-checked against the source. The per-token collapsed-Gibbs conditional, the MALLET three-bucket decomposition, decrement-before-score/restore-after bookkeeping, asymmetric-alpha handling, `beta_sum == β·V` consistency, and the AD-LDA parallel delta-merge all verified correct. Two concrete issues surfaced, both confirmed in the code.

## 1. Turbo mode can skip scheduled bookkeeping (opt-in path only)

The training loop advances `iter += batch` and then tests exact divisibility for the periodic work:

- `src/python/mod.rs:1264-1277` — `iter.is_multiple_of(optimize_interval)`, `draw_thin`, `check_every`, `progress_interval`.

When `batch == merge_every > 1`, `iter` can jump over an exact multiple, so that round of alpha/beta optimization, theta-draws, convergence checks, and progress callbacks is silently skipped. `step` is `> 1` only when `num_threads > 1` **and** `turbo_merge_every > 1` (`src/python/mod.rs:1239`), so the default exact path (`turbo_merge_every=1`) is unaffected.

Example: `turbo_merge_every=3`, `optimize_interval=10`, `burn_in=0`, `iters=29` → reconciled iters are 3,6,9,...,27,29; the optimization scheduled at 10 and 20 never runs.

**Fix:** split each batch at the next scheduled bookkeeping boundary instead of a fixed `step` stride.

## 2. Packed `(count, topic)` u32 can overflow on a hyper-frequent word (extreme scale)

Word-topic counts are bit-packed as `Entry = (count << topic_bits) | topic` in a `u32` (`src/model.rs:9-11`), incremented without an overflow guard (`src/sampler.rs:251`, `:302`; parallel rebuild `src/python/mod.rs:2661`).

Max representable count for a single `(word, topic)` is `2^(32 − topic_bits) − 1`. At `K=400` → `topic_bits=9` → ~8.4M occurrences of one word in one topic before the entry corrupts. Reachable only on very large corpora with an extremely dominant retained token.

**Fix options:** `u64` packing, or reject up front when a corpus's per-word frequency can exceed the packed-count capacity.

## Test gaps noted
- No exact dense-conditional invariant test (reconstruct `n_wt`/`n_t`/`n_dt` from assignments and assert equality after each sweep).
- No direct alpha/beta optimizer unit test vs a raw-count reference.
- No turbo-scheduling regression (e.g. `merge_every=3, check_every=5, iters=9`).
- No packed-count boundary test.

_Filed from a Codex review pass; findings independently verified against the source._

nealcaren:
Closes #417. Two independent LDA-core correctness fixes, both verified against source (Codex's review flagged them; its "reproduced" numbers were discarded since it runs read-only).

## 1. Turbo-mode bookkeeping skip
In turbo mode (`num_threads>1` and `turbo_merge_every>1`) the SparseLDA fit loop advances the iteration counter by the merge stride, then tested exact divisibility `iter.is_multiple_of(interval)`. Since `iter` only ever lands on multiples of the stride, any interval that doesn't divide the stride fires at the `lcm(stride, interval)` cadence — or **never**, when `stride * interval > iters`.

With the stock defaults (`optimize_interval=50`, `check_every=10`) and `turbo_merge_every=8`:
- α/β optimization fires every **200** sweeps instead of every 50,
- the convergence trace and progress callbacks are coarsened the same way.

**Fix:** replace the divisibility tests with `crossed_multiple(prev, iter, I)`, which fires once for any scheduled point inside the reconciled window `(prev, iter]`.
- In the exact per-sweep path (`step==1`) it reduces **exactly** to `iter % I == 0` → single-threaded path is byte-for-byte unchanged.
- `batch==step` is untouched, so the reconcile cadence and turbo RNG stream are unchanged; only genuinely-skipped events now fire.

Measured (num_threads=2, merge_every=8, progress_interval=10, iters=200): old code fires the progress callback at `[40,80,120,160,200]` (5 calls); fixed fires ~20 (one per crossed multiple of 10). The regression test asserts `16 in fired` — impossible under the old `is_multiple_of(10)` path — and is **proven to fail** on the reverted binding.

## 2. Packed-count u32 overflow
`TopicModel` packs `(count, topic)` into a `u32` with the count in the high `32 - topic_bits` bits. A single `(word, topic)` count reaching `2^(32-topic_bits)` shifts into/past bit 32 and **silently corrupts** the entry (reachable at high K with a very frequent word — e.g. K=65536 → ceiling 65535).

**Fix:** add `max_packable_count()` and fail fast in `initialize()` with a clear message when the most frequent word's total occurrence count (a safe upper bound on any single count) would exceed the ceiling — a clean error instead of corruption mid-sweep. Single choke point, covers every model using the packing, no signature change.

## Tests
- `tests/test_lda_turbo_bookkeeping_417.py`: crossed-window regression (fail-on-old verified) + exact-path no-op check.
- `src/model.rs`: ceiling-math unit test + `initialize` guard reject-above / accept-at-ceiling.

## Gates
`cargo fmt --all --check` ✓ · `cargo clippy` (default + `--features python`) ✓ · `cargo test --lib` 193 ✓ · LDA pytests 595 passed / 5 skipped ✓

### DIFF
--- a/src/model.rs
+++ b/src/model.rs
@@ -10,12 +10,15 @@ use rand::Rng;
 /// single topic.
 #[inline]
 pub fn max_packable_count(topic_bits: u32) -> u32 {
-    if topic_bits >= 32 {
+    match topic_bits {
+        // K == 1: the topic index needs no bits, so the entire 32-bit width holds
+        // the count. `1u32 << 32` is an out-of-range shift (it would wrap to 1 and
+        // wrongly yield a ceiling of 0, rejecting every K=1 fit), so special-case it.
+        0 => u32::MAX,
         // Degenerate: no bits left for the count. Cannot happen for real K, but
         // keep the arithmetic well-defined rather than shifting by 32.
-        0
-    } else {
-        (1u32 << (32 - topic_bits)) - 1
+        b if b >= 32 => 0,
+        b => (1u32 << (32 - b)) - 1,
     }
 }
 
@@ -444,6 +447,7 @@ mod tests {
     #[test]
     fn max_packable_count_matches_the_high_bit_width() {
         // Count occupies the high (32 - topic_bits) bits.
+        assert_eq!(max_packable_count(0), u32::MAX); // K=1: full 32-bit count width
         assert_eq!(max_packable_count(1), (1u32 << 31) - 1);
         assert_eq!(max_packable_count(10), (1u32 << 22) - 1); // K up to 1024
         assert_eq!(max_packable_count(16), 65_535); // K up to 65536
@@ -505,6 +509,25 @@ mod tests {
         let corpus = Corpus {
             id_to_word: vec!["w0".to_string()],
             docs: vec![vec![0u32, 0, 0]], // 3 == ceiling
+            doc_names: vec!["d0".to_string()],
+            doc_labels: vec![String::new()],
+            doc_freqs: vec![0u32; 1],
+            total_freqs: vec![0u32; 1],
+        };
+        let mut rng = ChaCha8Rng::seed_from_u64(1);
+        m.initialize(&corpus, &mut rng); // must not panic
+    }
+
+    // Regression for the K=1 panic (#447): num_topics=1 => topic_bits=0 => the count
+    // uses the full 32-bit width, so the ceiling is u32::MAX and a frequent word must
+    // be accepted rather than rejected by the overflow guard.
+    #[test]
+    fn initialize_accepts_a_frequent_word_when_k_is_one() {
+        let mut m = TopicModel::new(1, 1.0, 0.01, 1);
+        assert_eq!(m.topic_bits, 0);
+        let corpus = Corpus {
+            id_to_word: vec!["w0".to_string()],
+            docs: vec![vec![0u32; 49]], // 49 occurrences, as in the CI-failing fits
             doc_names: vec!["d0".to_string()],
             doc_labels: vec![String::new()],
             doc_freqs: vec![0u32; 1],
--- a/src/model.rs
+++ b/src/model.rs
@@ -2,6 +2,23 @@ use crate::corpus::Corpus;
 use crate::estimator::{DirichletModel, Estimator, ModelFamily};
 use rand::Rng;
 
+/// Largest per-`(word, topic)` count representable in the packed `u32` entry
+/// when `topic_bits` low bits are reserved for the topic index. The count lives
+/// in the high `32 - topic_bits` bits, so a count of `2^(32 - topic_bits)` (or
+/// more) shifts into or past bit 32 and silently corrupts the entry. `initialize`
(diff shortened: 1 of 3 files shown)


nealcaren:
## Review status: needs a fix — CI red (real bug, all 3 OSes)

The `test` (pytest) job fails on ubuntu/macos/windows; lint/docs/rust-tests pass. Root cause found:

```
PanicException: a word occurs 49 times, but the packed (count, topic) table for
num_topics=1 can represent at most 0 occurrences of a single word in one topic
```
Failing: `test_fit_stats.py::TestPerplexityRecoversK::test_k2_lower_than_k1`, `TestCoherence::test_coherence_higher_for_good_model`, `test_heldout.py::test_better_fit_scores_higher_loglik` — all of which fit a **K=1** baseline.

### Bug: `max_packable_count` is wrong for `topic_bits == 0` (i.e. `num_topics == 1`)

`TopicModel::initialize` sets `topic_bits` from the topic index width (`src/model.rs:33-41`). For `num_topics == 1`: `1.is_power_of_two()` is true → `mask = 0` → `topic_bits = 0u32.count_ones() = 0`.

Then `max_packable_count(0)` evaluates `(1u32 << (32 - 0)) - 1` = `(1u32 << 32) - 1`. A shift by 32 is out of range for `u32`; on x86 the count is masked mod 32, so `1u32 << 32` becomes `1u32 << 0 = 1`, and the function returns **0**. The new `assert!(worst <= ceiling)` in `initialize` then panics on *any* word — every K=1 fit dies.

(With `topic_bits == 0` the topic index needs no bits, so all 32 bits hold the count and the true ceiling is `u32::MAX` — K=1 should never be rejected.)

### Fix
Special-case `topic_bits == 0`:
```rust
pub fn max_packable_count(topic_bits: u32) -> u32 {
    match topic_bits {
        0 => u32::MAX,        // K=1: no topic bits, full 32-bit count
        b if b >= 32 => 0,    // degenerate, well-defined
        b => (1u32 << (32 - b)) - 1,
    }
}
```
Add a `max_packable_count(0) == u32::MAX` case to `max_packable_count_matches_the_high_bit_width`, and a K=1 `initialize` smoke test (a frequent word must be accepted).

### Scope note (optional)
The packed-count overflow guard is unrelated to #417 (turbo bookkeeping). Consider splitting it into its own PR/commit. The actual #417 fix — replacing `iter.is_multiple_of(interval)` with the half-open-window `crossed_multiple(prev, iter, interval)` — looks correct and is not implicated in the failure; both new turbo tests should pass once the K=1 guard is fixed.

Issue #68, opened 2026-06-11
HDP topic count runs away over iterations (K: 3→774 in 800 sweeps); defaults lowered as partial mitigation


nealcaren:
## Summary

On a standard corpus, topica's `HDP` infers an implausibly large number of
topics with its default settings, far more than a reference HDP infers on the
same data.

## Observation

Benchmarking against `tomotopy` on a seeded 3,500-document subsample of stm's
`poliblog5k` (2,632-word vocabulary), 500 Gibbs sweeps, default settings:

| | inferred topics |
|---|---|
| `topica.HDP()` | **1254** |
| `tomotopy.HDPModel()` | 2 |

1254 topics on 3,500 short blog posts is not a reasonable discovery; tomotopy's
2 is plausibly too few, but the topica result is the clear outlier. The runaway
topic count also makes HDP look far slower than it is in any head-to-head, since
it is sampling over more than a thousand topics.

## Likely cause

`HDP`'s default concentration parameters are `alpha=1.0`, `gamma=1.0` with
`resample_conc=True` (`src/python.rs:5850`). The corpus-level `gamma` controls
how readily new topics are created, and `gamma=1.0` with resampling appears to
drive continual topic creation on this data. Reference HDP implementations use
smaller, often resampled-from-a-prior concentrations (tomotopy defaults to much
more conservative values), so topica's defaults likely run hot.

## What to check

- Whether the default `alpha`/`gamma` (both 1.0) are appropriate, or should
  match the smaller reference defaults / a vaguer Gamma hyperprior on the
  concentrations.
- Whether the concentration resampling (`resample_conc=True`) is implemented as
  the Teh et al. (2006) auxiliary-variable scheme, and whether it is converging
  to a sane `gamma` or drifting upward.
- Validate the inferred topic count against tomotopy and the Teh et al.
  reference on a corpus with a known rough topic count, and document the
  expected behavior.

This is separate from the sampler-speed work in #67; it is about HDP correctness
and sensible defaults, not raw throughput.

nealcaren:
## Update: this is a runaway feedback loop, not just bad defaults

Shipped a partial mitigation in `5bc9a07`: the `HDP` default concentrations are now
`alpha=0.1`, `gamma=0.1` (were `1.0`, `1.0`), matching the reference convention.
That alone takes the inferred K on the 3,500-doc poliblog corpus from ~1254 down
to ~70 at 500 sweeps, and the top ~25 topics are high quality. **But it only
[...]

Issue #443, opened 2026-07-23
Save format (bincode) does not migrate across schema changes despite #[serde(default)] — audit before release


nealcaren:
Surfaced during the codex review of #441 (SAGE sparse prior).

## The gap
topica's save format is **bincode** (`src/saveformat.rs`), which is positional and *not* self-describing — struct fields are serialized in declaration order with no field names. The many `#[serde(default)]` / `#[serde(default = "...")]` annotations across the `*State` structs in `src/python/mod.rs` (e.g. `SageState`, `DmrState`, `CtmState`, …) **imply** field-level backward compatibility, but bincode does not honor them for genuinely older files: deserializing reads each field positionally and errors on EOF (or, worse, silently misreads if a field was inserted mid-struct), rather than falling back to the default.

Concretely: a struct that gained a trailing `#[serde(default)]` field cannot load a file written before that field existed — bincode hits unexpected-EOF. A field inserted in the *middle* is worse: every subsequent field is misaligned. (This is why #441 moved its new fields to the end and dropped its "old files load as gaussian" claim.)

## Why it hasn't bitten
topica is pre-release with a sole active user and no meaningful corpus of persisted old `.tt` files, so cross-version load has never actually been exercised — the `#[serde(default)]` fields only round-trip within a single version (which the save/load tests do cover).

## Impact
As-is this is a latent correctness/expectation gap, not an active bug. But before the release + paper, saved models become a durable artifact, and the current annotations promise a compatibility the format cannot deliver.

## Options
1. **Document the limitation** honestly: topica save files are not guaranteed loadable across schema changes; re-fit after upgrading. (Cheapest; matches reality.)
2. **Version + migrate**: the 8-byte header already carries a version byte (`saveformat.rs`); bump it per schema change and dispatch to a versioned decoder for the structs that need it. (Real migration; more work.)
3. **Switch to a self-describing / field-named format** for the payload (e.g. MessagePack via `rmp-serde`, or CBOR), where `#[serde(default)]` for missing fields works as intended. (Cleanest long-term; changes the on-disk format once.)

## Recommendation
For 0.5, at minimum do (1) — and either audit the existing `#[serde(default)]` annotations to note they are round-trip-only, or pick (2)/(3) before the format is frozen for the paper. A `test_save_load` with a committed pre-schema binary fixture would turn this from "believed to work" into "known".

PR #452, opened 2026-07-23
fix(hdp): make the resampled-concentration cap configurable (#433)

### LINKED ISSUE #433: HDP review: core sampler verified correct; CONCENTRATION_MAX=2.0 clamp biases resampled concentrations (opt-in), plus minor robustness (beta_u cancellation, sample_index NaN, input validation)
Independent correctness review of the HDP direct-assignment collapsed Gibbs sampler (`src/hdp.rs`, Teh/Jordan/Beal/Blei 2006), via the Codex second-opinion agent (`gpt-5.6-sol`) plus a full independent read of the source. 

**The core sampler is correct.** Verified against the reference, both by hand and by Codex:
- Direct-assignment predictive `(njk + alpha*beta_t) * (nkw+eta)/(nk+V*eta)` for existing topics and `alpha*beta_u*(1/V)` for a fresh topic (the base-measure marginal of one word is exactly 1/V, independent of eta); leave-one-out counts removed before scoring and restored after (`hdp.rs:366-368`, `353-355`).
- Antoniak table sampling `m = sum_i Bernoulli(theta/(theta+i))` (`sample_tables`, `hdp.rs:162`) — correct; the `i=0` term is always 1 so `m>=1` whenever `n>=1`, making the `.max(1)` at `:245` redundant but harmless.
- `resample_beta`: `beta ~ Dirichlet(m_.1, ..., m_.K, gamma)` via normalized Gamma draws, preserving `sum(beta)+beta_u = 1` (`hdp.rs:235`).
- Both concentration updates: `resample_gamma` (Escobar-West 1995, top level) and `resample_alpha` (Teh et al. 2006 second level, matching blei-lab/hdp) — correct shape/rate and auxiliaries, no level mix-up (`hdp.rs:266`, `287`).
- `compact()`: drops empty topics, returns their beta mass to beta_u, remaps counts and z (safe because an `nk==0` topic has no assignments) (`hdp.rs:199`).
- `sample_gamma` (Marsaglia-Tsang, incl. the `shape<1` boost) and `sample_beta_dist` — correct.

**Not a bug (explicitly checked):** the fit loop (`hdp.rs:416-424`) orders `sweep -> compact -> resample_beta (beta | m, gamma) -> resample_gamma (gamma | m, K) -> resample_alpha`. beta_u is drawn with the pre-update gamma, but `resample_gamma`/`resample_alpha` are *collapsed* updates conditioning on the table counts and K (beta marginalized out), so this is a valid Gibbs scan — the next sweep redraws beta with the new gamma. The stationary distribution is preserved.

### 1. [Low-medium, opt-in] `CONCENTRATION_MAX = 2.0` clamp biases resampled concentrations

`hdp.rs:196` caps both resampled concentrations at 2.0 (`.clamp(1e-3, CONCENTRATION_MAX)` at `:280` and `:311`). When `resample_conc = true`, any posterior with appreciable mass above 2.0 gets a **spurious atom at exactly 2.0**, biasing `gamma`/`alpha` downward and thereby biasing K. The doc comment (`hdp.rs:188-195`) asserts "healthy fits keep both in roughly [0.05, 1.5]", but that is corpus-dependent — larger corpora with many topics can legitimately want `gamma > 2`. The **default fits with fixed concentrations** (`resample_conc = false`), so this only bites the opt-in resampling path. Consider raising the cap substantially (e.g. to the number of topics, or a large constant) or making it configurable, and rely on the `1e-3` floor plus the Gamma prior for stability rather than a hard low ceiling.

### 2. [Low] `beta_u` can cancel to exactly 0
`beta_u *= 1.0 - b` (`hdp.rs:345`) can underflow to exactly 0 after many births or with small `gamma`, suppressing further topic births until the next `resample_beta` restores it. Self-heals each iteration, but retaining the residual directly (rather than as `1 - b`) would reduce cancellation.

### 3. [Low] `sample_index` does not reject nonfinite / nonpositive totals
`sample_index` (`hdp.rs:172`) correctly accepts unnormalized weights, but on NaN/inf it silently falls through to the last index (the fresh-topic option), masking numerical corruption. An assertion or explicit error would surface it.

### 4. [Low] Public `fit_hdp` lacks input validation
The core `fit_hdp` does not validate positive/finite concentrations, positive `eta`, nonzero vocabulary, or in-range word IDs; the Python constructor covers hyperparameters, but direct Rust callers can trigger NaNs / division-by-zero / indexing panics. `u32` count tables also wrap above ~4.29e9 tokens (release mode).

## Suggested tests
- Exact assignment-weight test vs the closed form after leave-one-out, across several `eta` (confirm fresh likelihood is exactly 1/V).
- Count-reconstruction invariant after every sweep + compaction (rebuild nkw/nk/njk from z; assert equality and valid indices).
- Antoniak distribution test vs the exact unsigned-Stirling distribution for small n, theta (assert m>=1 when n>0).
- Dirichlet-beta means `E[beta_k]=m_k/(M+gamma)`, `E[beta_u]=gamma/(M+gamma)`.
- Concentration test with `resample_conc=true` where the true posterior has mass above 2.0 -> assert no atom at exactly 2.0 after raising/removing the clamp.
- Compaction fixture with interleaved live/dead topics: beta-mass conservation, exact z remap, exact count reconstruction.
- `sample_index` rejects empty / nonfinite / negative / nonpositive-total inputs.

_Reviewed via Codex + a full independent source read; the core-correctness verifications above were confirmed by hand, and the beta/gamma ordering was checked to be a valid collapsed Gibbs scan (not a defect)._

nealcaren:
Closes #433. The HDP direct-assignment Gibbs sampler was verified correct by review (Codex `gpt-5.6-sol` + a full source read; Codex's "reproduced" numbers were discarded — it runs read-only). One substantive finding plus three minor robustness items.

## Main fix — configurable concentration cap (finding 1)
The opt-in resampler (`resample_conc=True`) hard-capped both `α0` and `γ` at `CONCENTRATION_MAX = 2.0`. The Escobar-West `γ` update draws from `Gamma(a+K, b-log η)`, whose posterior mean grows with the topic count K, so on a large-K corpus the true posterior sits **above 2.0** — and the hard cap pins `γ` at exactly 2.0, a spurious atom that biases `γ` (and hence K) downward.

The 2.0 backstop came from the #68 runaway, but the **primary** #68 fix was defaulting `resample_conc=False`. The clamp only bites the opt-in path, and 2.0 is too low to be a statistical bound.

**Fix:** expose it as `concentration_max` (Python constructor + `fit_hdp`), **default 2.0 so existing opt-in behavior is unchanged**, documented as a divergence backstop rather than a prior. Corpora that legitimately support larger concentrations can raise it and let `γ` float. Threaded through `settings()`, save/load (`serde` default 2.0 for older models), and the `.pyi` stub.

**Measured:** on a large-K state the resampled `γ` is pinned at exactly 2.0 under the default cap and rises above 2.0 under a raised cap — confirming the atom is real and the knob removes it (Rust `raising_the_cap_removes_the_spurious_atom_at_two`, Python `test_raising_the_cap_lets_the_learned_concentration_exceed_two`).

## Minor robustness (findings 2–4)
- **β_u cancellation:** break the new stick mass off by subtraction (`β_u -= new_beta`) instead of rescaling by `(1 - b)`, which cancels to exactly 0 when `b ≈ 1` (small `γ`) and stalls topic births.
- **`sample_index` NaN:** `debug_assert!` the weights sum to a finite positive total instead of silently falling through to the fresh-topic index on NaN/inf (per-token hot path → debug-only).
- **`fit_hdp` validation:** check `num_types`/`alpha`/`gamma`/`eta`/`concentration_max` up front so direct Rust callers get a clear message instead of NaNs / index panics.

## Not changed (verified correct in review)
Assignment weights + leave-one-out bookkeeping, Antoniak table sampling, `resample_beta` Dirichlet, both concentration updates (no level mix-up), `compact()`, and the β/γ ordering (a valid collapsed Gibbs scan).

## Gates
`cargo fmt --all --check` ✓ · clippy (default + `--features python`) ✓ · `cargo test --lib` 191 ✓ · HDP pytests 73 existing + 4 new ✓ · stub-sync 78 ✓

### DIFF
--- a/python/topica/_topica.pyi
+++ b/python/topica/_topica.pyi
@@ -986,15 +986,23 @@ class HDP:
         beta: float = 0.01,
         seed: int = 42,
         resample_conc: bool = False,
+        concentration_max: float = 2.0,
         eta: Optional[float] = None,
     ) -> None:
         """alpha/gamma are the document- and corpus-level DP concentrations.
         gamma is the dominant lever on the inferred topic count (0.1 is
         conservative; raise it for more topics). resample_conc defaults to False
         (fixed concentrations -> a stable topic count); set it True to adapt the
-        concentrations to the data, which is now capped to avoid the runaway
-        topic count it used to cause (issue #68). beta is the topic-word Dirichlet
-        (base measure). alpha, gamma, beta must be > 0.
+        concentrations to the data, which is capped at concentration_max to avoid
+        the runaway topic count it used to cause (issue #68). beta is the
+        topic-word Dirichlet (base measure). alpha, gamma, beta must be > 0.
+
+        concentration_max (default 2.0) bounds the resampled alpha/gamma when
+        resample_conc=True. It is a divergence backstop, not a statistical prior:
+        a posterior with mass above it is pinned at the cap, biasing the
+        concentrations (and the topic count) downward, so raise it for corpora
+        that legitimately support larger concentrations. No effect when
+        resample_conc=False. Must be finite and > 1e-3.
 
         eta is a deprecated alias for beta."""
         ...
--- a/src/hdp.rs
+++ b/src/hdp.rs
@@ -357,11 +357,7 @@ impl HdpModel {
             // Instantiate a new topic; break a Beta(1, γ) piece off β_u.
             let b = sample_beta_dist(1.0, self.gamma, rng);
             let new_beta = b * self.beta_u;
-            // Subtract the mass actually broken off rather than rescaling by
-            // `1 - b`; when b ≈ 1 (small γ) the `1 - b` form cancels to exactly 0
-            // and stalls further topic births, whereas this keeps β_u consistent
-            // with the mass handed out.
-            self.beta_u -= new_beta;
+            self.beta_u *= 1.0 - b;
             self.beta.push(new_beta);
             self.nkw.push(vec![0u32; v]);
             self.nk.push(0);
@@ -711,23 +707,33 @@ mod tests {
         let mut lo = large_k_state(DEFAULT_CONCENTRATION_MAX); // 2.0
         let mut rng_hi = ChaCha8Rng::seed_from_u64(0);
         let mut hi = large_k_state(1.0e6);
-        let mut hi_exceeded_two = false;
+        let mut hi_gamma_exceeded_two = false;
+        let mut hi_alpha_exceeded_two = false;
         for _ in 0..50 {
             let (m_lo, t_lo) = lo.resample_beta(&mut rng_lo);
             lo.resample_gamma(m_lo, &mut rng_lo);
             lo.resample_alpha(&t_lo, &mut rng_lo);
+            // The cap binds BOTH concentrations, not just gamma.
             assert!(lo.gamma <= DEFAULT_CONCENTRATION_MAX + 1e-9);
+            assert!(lo.alpha <= DEFAULT_CONCENTRATION_MAX + 1e-9);
 
             let (m_hi, t_hi) = hi.resample_beta(&mut rng_hi);
             hi.resample_gamma(m_hi, &mut rng_hi);
             hi.resample_alpha(&t_hi, &mut rng_hi);
             if hi.gamma > DEFAULT_CONCENTRATION_MAX + 1e-6 {
-                hi_exceeded_two = true;
+                hi_gamma_exceeded_two = true;
+            }
+            if hi.alpha > DEFAULT_CONCENTRATION_MAX + 1e-6 {
+                hi_alpha_exceeded_two = true;
             }
         }
         assert!(
-            hi_exceeded_two,
+            hi_gamma_exceeded_two,
             "with a raised cap the large-K posterior should let gamma exceed 2.0"
+        );
+        assert!(
+            hi_alpha_exceeded_two,
+            "with a raised cap the second-level posterior should let alpha exceed 2.0"
         );
     }
 
--- a/src/hdp.rs
+++ b/src/hdp.rs
@@ -36,6 +36,9 @@ pub struct HdpModel {
     pub beta_u: f64,        // leftover stick mass (the "new topic" weight)
     pub z: Vec<Vec<usize>>, // per-document token topic assignments
     pub njk: Vec<Vec<u32>>, // num_docs × K document-topic counts
+    /// Divergence backstop for resampled concentrations (see
+    /// [`DEFAULT_CONCENTRATION_MAX`]); only consulted when `resample_conc = true`.
+    pub concentration_max: f64,
     /// Discovery + convergence trace, one entry per recorded sweep:
     /// `(iteration, num_active_topics, per-token log-likelihood, alpha, gamma)`.
     /// The topic count and the (resampled) concentrations show the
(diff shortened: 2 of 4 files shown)


nealcaren:
## Review status: core fix is correct, three items to resolve before merge

The configurable-cap fix for #433 is implemented correctly — the Escobar-West γ/α updates are sound, the cap is applied to both concentrations with the default still 2.0, validation rejects non-finite/too-small caps, determinism holds, and the discovery trajectory matches the base commit exactly under both fixed and resampled concentrations. Three things to fix:

### 1. `.pyi` stub disagrees with the runtime signature (concrete contract bug)
The pyo3 signature uses a Rust const expression:
```rust
#[pyo3(signature = (..., concentration_max=hdp::DEFAULT_CONCENTRATION_MAX, eta=None))]
```
PyO3 renders that as `concentration_max=Ellipsis` in the runtime signature, so `inspect.signature(topica.HDP)` reports `...` while `_topica.pyi` promises `2.0`. Construction still defaults to 2.0, but introspection/docs/stub disagree. Use a literal `concentration_max=2.0` in the `#[pyo3(signature = ...)]`.

### 2. Mid-struct `HdpState` field corrupts old-save loads; the migration claim is false
`concentration_max` is inserted in the **middle** of the positional-bincode `HdpState` (between `resample_conc` and `fitted`), and `default_concentration_max`'s comment claims pre-#452 saves "load at 2.0". They don't — `#[serde(default)]` is inert for positional bincode (the same file documents this at `mod.rs:232`). A mid-struct insert is worse than appending: it misaligns every downstream field. Loading a base-commit save fails with `not a valid topica model file: tag for enum is not valid, found 4`. At minimum, **append the field at the end** of `HdpState` so old saves fail cleanly (EOF) rather than corrupt, and correct the comment to state the break honestly (refit required; holistic fix is #443). This is the same positional-bincode trap seen across the batch — not something this PR must fully solve, but the claim must match reality.

### 3. Drop the unrelated `beta_u` change — it doesn't fix its stated problem
`beta_u *= 1.0 - b` → `beta_u -= b * beta_u` (`src/hdp.rs:357`). These are algebraically identical; when `b` rounds to exactly `1.0` **both** yield 0, so the "stalls topic births" cancellation is not fixed. For representable `b < 1`, `1 - b` is actually the better residual; the multiply-then-subtract adds a rounding step. It's also unrelated to #433 and can perturb edge-case topic-birth trajectories. Drop it from this PR (or, if a real fix is intended, compute both Beta proportions directly from the underlying Gamma draws and do it in its own PR).

### Also worth adding
- `inspect.signature(HDP)` reports `concentration_max=2.0` (guards #1).
- A base-vs-default-cap topic-count-history regression, and a test that raising the cap frees **α** above 2.0 (not just γ).

The input-validation asserts in `fit_hdp` are a reasonable safety add and can stay.

nealcaren:
**Maintainer note:** save/load backward-compat isn't a concern until v1.0 (only Neal has saved models so far), so item #2 (mid-struct `HdpState` / old-save break) is **non-blocking** — no need to move the field or add migration; just don't *claim* old saves load (correct the `default_concentration_max` comment). The real blockers stand: **#1** the pyo3 signature renders `concentration_max=Ellipsis` while the .pyi says `2.0` (use a literal `2.0`), and **#3** drop the unrelated `beta_u` change (it doesn't fix its stated cancellation).

Issue #428, opened 2026-07-23
ProdLDA/VAE review: BatchNorm running-var biased vs PyTorch unbiased; Dirichlet-prior transform uses softmax(mu) not Weibull median; Weibull shape floor=1 restricts sparse family


nealcaren:
Independent correctness review of the ProdLDA/AVITM VAE core (`src/prodlda.rs`, the hand-coded batched VAE shared by ProdLDA/ETM-VAE/CombinedTM/ZeroShotTM/Scholar/InfoCTM), via the Codex second-opinion agent (`gpt-5.6-sol`), directed at what the existing finite-difference gradient tests CANNOT catch (KL/ELBO values, batchnorm train/eval, eval-path consistency). Findings re-verified against the source.

**The core is correct.** Confirmed faithful: the three KL closed forms (Laplace logistic-normal; Weibull-Gamma, zero at k=lam=alpha=1; stick-breaking), the KL/reconstruction relative scale and the `1/n` minibatch averaging, the product-of-experts decoder `softmax_v(BN(theta . beta))` (matches Pyro's affine-free decoder BN), all three reparameterizations, and Adam (`prodlda.rs:1432`: b1=0.99, b2=0.999, eps=1e-8, correct 1-b^t bias correction, coupled L2 decay = torch Adam not AdamW). The findings below are eval-time parity and non-default-prior deviations, not core bugs — consistent with the multi-seed PyTorch parity having been validated on the default Laplace path.

## 1. [Medium] BatchNorm running variance is biased; PyTorch/Pyro use unbiased

`forward_train` computes the biased batch variance (`var[j] /= n`, `prodlda.rs:123`) and `update_running` folds that SAME biased variance into `running_var` (`prodlda.rs:141`). PyTorch `BatchNorm1d` uses the biased variance for the in-batch normalization but the **Bessel-corrected `/(n-1)` unbiased** variance for the running statistic. topica's eval-time `running_var` (used by `forward_eval_row`, `prodlda.rs:150`, at transform time) is therefore systematically too small by `(n-1)/n`.

Negligible at the default `batch_size=1000` (x0.999 — why parity held), but material for small batch sizes and for the small final batch. Fix: feed `var * n/(n-1)` (n>1) into `update_running`, or compute an unbiased running var separately.

## 2. [Medium] The Dirichlet prior's transform uses softmax(mu), not the Weibull median it trains under

[...]

PR #455, opened 2026-07-23
fix(prodlda): transform Dirichlet-prior models via the Weibull median (#428)

nealcaren:
Addresses #428. ProdLDA/AVITM review follow-up (Codex `gpt-5.6-sol` + independent source read; Codex's "reproduced" numbers discarded — read-only sandbox). Finding 1 (BatchNorm running variance) already shipped in **#436**; this covers the remaining substantive item and documents the two deliberate family/objective choices.

## Finding 2 — Dirichlet train/eval inconsistency (the fix)
The `Prior::Dirichlet` forward trains `θ` as the **normalized Weibull posterior median** (`u=0.5`), but `transform_with_emb` returned `softmax(mu)` for every non-stick prior. So a Dirichlet model reported `doc_topic`/`transform` from a *different* point estimate than the one it was optimized under. The transform only kept the mean head (`bn_mu`); the median also needs the log-variance head (`bn_lv`), which was never retained — that's why the shortcut existed.

- Add `bn_lv: Option<BatchNorm>` to `ProdldaModel`, populated at fit.
- `transform_with_emb` branches `Prior::Dirichlet` to `weibull_median_theta(mu, lv)`, a shared helper matching the training median exactly. **Laplace (`softmax(mu)`) and stick-breaking are unchanged — the default path is byte-for-byte identical.**
- `Option` keeps old saves working: a model loaded without `bn_lv` falls back to `softmax(mu)` (previous behavior) rather than fabricating stats.
- Persist `bn_lv` running stats wherever `bn_mu` already is (`ProdldaState` + `CtmEmbState` for CombinedTM/ZeroShotTM), `#[serde(default)]` so older saves load. SCHOLAR/InfoCTM fits are Laplace-only → `bn_lv: None` (never read).

**Measured:** on a fitted Dirichlet model the median transform genuinely differs from `softmax(mu)` (so the change is observable, not a no-op), and it is identical across save/load.

## Findings 3 & 4 — documented, not changed (both would need re-validation)
- `WEIBULL_SHAPE_FLOOR = 1.0` is a *family restriction* (caps posterior sparsity, so `Dirichlet(α<1)` is approximated by a shape-1 posterior), now stated as such.
- The Gaussian KL regularizes all K coordinates; for stick-breaking only K−1 affect `θ`, so the K-th is a regularized-but-unused latent — a valid ELBO for the augmented model, a deliberate deviation from Miao et al.'s K−1 projection.

Finding 5 (reconstruction floor / `D=1` / partial-batch drop) left for a separate focused change — altering the recon objective changes `bound` semantics and touches parity.

## Tests
- **Rust:** a fitted Dirichlet model's `transform` == the median recomputed from eval-time heads and ≠ `softmax(mu)`; `bn_lv=None` falls back to `softmax(mu)`.
- **Python:** Dirichlet transform identical across save/load (`bn_lv` round-trips); Dirichlet vs laplace transforms differ on the same corpus.

## Gates
`cargo fmt --all --check` ✓ · clippy (default + `--features python`) ✓ · `cargo test --lib` 191 ✓ · VAE-family + stub pytests 410 passed / 1 skipped ✓

### DIFF
--- a/src/infoctm.rs
+++ b/src/infoctm.rs
@@ -563,6 +563,8 @@ fn finish(
         epochs_run,
         weights: s.w,
         bn_mu: s.bn_mu,
+        // Laplace transform is softmax(mu) and never reads bn_lv (#428).
+        bn_lv: None,
         prior: Prior::Laplace,
     };
     let empty: Vec<Vec<f64>> = vec![Vec::new(); docs.len()];
--- a/src/prodlda.rs
+++ b/src/prodlda.rs
@@ -692,6 +692,10 @@ pub(crate) const WEIBULL_FLOOR: f64 = 1e-4;
 /// coefficient of variation, so the reparameterized topic vector is too noisy to
 /// train; flooring the shape at 1 keeps the variational posterior concentrated
 /// enough to learn (a standard WHAI-style choice) while staying strictly positive.
+/// This is deliberately a family restriction, not just a numerical floor: it caps
+/// how sparse the posterior can be, so a `Dirichlet(alpha)` prior with `alpha < 1`
+/// is approximated by a shape-1 (exponential) posterior rather than a matching
+/// sub-1 shape. Loosen it only alongside re-validating stability (#428, finding 3).
 pub(crate) const WEIBULL_SHAPE_FLOOR: f64 = 1.0;
 
 /// Map a post-BN head pair `(a, b)` to a Weibull `(shape, scale)` and, given the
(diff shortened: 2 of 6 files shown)


nealcaren:
## Review status: fix is correct, but needs test + doc changes before merge

The transform fix itself is **mathematically correct and well-guarded by the Rust test**. Verified: the training Weibull parameterization `k=softplus(μ)+1`, `λ=softplus(lv)+1e-4`, `g=λ·exp(ln[-ln(1-u)]/k)` at `u=0.5` gives `g=λ(ln 2)^(1/k)`, exactly the median; non-Dirichlet paths (`softmax(mu)`, `stick_break`) are untouched; determinism holds; the `scholar.rs`/`scholar.rs`/`infoctm.rs` `bn_lv: None` call sites are consistent. The Rust unit test `dirichlet_transform_uses_the_weibull_median_not_softmax_mu` is genuinely discriminating.

Two things to address:

### 1. The two new Python tests are non-discriminating (they pass pre-fix)
Running `tests/test_prodlda_dirichlet_transform_428.py` against a **pre-fix** binary passes both tests: save/load preserves the old (wrong) transform result, and a separately-trained Laplace vs Dirichlet model already differ for reasons unrelated to the transform. So they don't actually guard the fix. Replace with one of:
- A **frozen expected-value** regression: fit a fixed-seed Dirichlet model, assert `transform(docs)` equals hard-coded θ values that match the Weibull median (would fail under `softmax(mu)`).
- Or an in-session assertion that `transform` ≠ `softmax(mu)` computed from the *same* model's eval heads (mirroring the Rust test), rather than comparing two independently-trained models.
- An independent numeric check of `weibull_median_theta` (the Rust test computes its expectation with the same helper, so an algebra error there would pass).

### 2. Save-format comments overstate backward compatibility (the #443 trap)
`bn_lv_running_mean`/`bn_lv_running_var` are appended to the positional-bincode `ProdldaState` and `CtmEmbState` with `#[serde(default)]`, and the comments say older saves "fall back to softmax(mu)". They don't: bincode hits EOF before the model is constructed, so loading **any** pre-PR ProdLDA / CombinedTM / ZeroShotTM save fails with `not a valid topica model file: unexpected end of file` (format version stays `1`). This is the known positional-bincode limitation (#443), not something this PR must solve — but the comments should state the break honestly (old saves must be refit) rather than claim a fallback that can't run. The in-memory `bn_lv: None` fallback is still correct and *is* exercised by SCHOLAR; only the "older save" clause is wrong.

### Minor
- "the `Prior::Dirichlet` forward trains under" the median slightly overstates: non-contrastive reconstruction trains on Weibull *samples*, not specifically the median (the contrastive view does use the median). Soften the wording.
- Stale `theta = softmax(mu)` docs at `src/python/neural.rs` and the `ProdldaModel::transform_with_emb` doc comment should be updated for the Dirichlet branch.

nealcaren:
**Maintainer note:** save/load backward-compat is deprioritized to v1.0, so item #2 is **non-blocking** beyond correcting the inaccurate "falls back to softmax" comment. The real item is **#1** — the two new Python tests pass against a pre-fix binary, so they don't guard the fix; please make them discriminating (frozen expected θ, or compare against `softmax(mu)` from the same model's eval heads). The Rust test already guards it, so this is test-quality, not a math problem.

PR #470, opened 2026-07-23
fix(tlda): make TensorLDA parity portable and gate topic recovery (#451)

### LINKED ISSUE #451: TensorLDA review: make upstream parity portable and CI-runnable
## Review outcome

TensorLDA remains correctly marked experimental: its published validation record does not support using weights as calibrated prevalence estimates. No new source-level correctness defect was established in this review.

The remaining fidelity blocker is reproducible upstream validation. parity/tlda_compare.py imports the reference wrapper from a hard-coded personal scratch path, so it is neither portable nor an automated test. The validation record already identifies portable, automated square-rank reference comparison as a prerequisite for promotion.

- [ ] Package or document an installable reference dependency and remove the machine-specific sys.path mutation.
- [ ] Make the valid square-rank upstream comparison runnable in CI or a separately scheduled integration job.
- [ ] Preserve the separate status of n_eigenvec greater than num_topics: it is a Topica extension, not an upstream-parity target.
- [ ] Turn known-truth LDA simulations into explicit acceptance criteria for topic recovery and keep prevalence calibration as a separate question.

Relevant surfaces: parity/tlda_compare.py and docs/replications/tlda.md.

nealcaren:
Closes #451.

TensorLDA's parity infrastructure had two portability problems: the upstream comparison hard-coded a machine-specific `sys.path` to a personal scratch directory, and the known-truth LDA simulation was a print-only report with no acceptance criteria.

## What changed

**Portable reference (a):** New `parity/tlda_ref.py` locates the [TensorLy TLDA](https://github.com/tensorly/tlda) reference via the documented `TOPICA_TLDA_REF` env var — never a baked-in path — trying flat-checkout, packaged, and wrapper import shapes, returning `None` when absent. `parity/tlda_compare.py` drops the personal `sys.path` mutation and uses it.

**CI-runnable / schedulable (b):** `parity/tlda_compare.py` now exits 0 when the reference is unavailable, so it is safe to wire into CI or a scheduled integration job: a no-op wherever the ref is not installed, a real square-rank comparison wherever it is. Verified end-to-end against a real reference checkout (streaming aligned cosine ≈ 1.0).

**Explicit topic-recovery acceptance criteria (d):** New hermetic `tests/test_tlda_recovery.py` turns the square-rank (`n_eigenvec == num_topics`) known-truth simulation into a gate with a mean-cosine floor (0.50) and per-topic floor (0.35), below the current baseline (0.529 / min 0.443). Runs in ~4s, imports only topica + numpy.

**Prevalence kept separate (d) + extension status (c):** planted-weight and doc-topic recovery are computed as diagnostics but **not** asserted (the `weights` are not yet a validated prevalence estimator); `n_eigenvec > num_topics` is exercised as a Topica extension, not an upstream-parity claim. Docs updated with the reference-install steps and the gate/prevalence split.

## Verification
- `tests/test_tlda.py` + `test_tlda_gold.py` + `test_tlda_recovery.py`: 19 passed
- Discriminating check: raising the floor to 0.60 (above baseline) fails the gate; restored floor passes
- `cargo fmt --check`, `scripts/preflight.sh`, `mkdocs build --strict`: all green
- `parity/tlda_compare.py` with no env var exits 0; with `TOPICA_TLDA_REF` set runs the full comparison

### DIFF
--- a/docs/replications/tlda.md
+++ b/docs/replications/tlda.md
@@ -79,6 +79,24 @@ is init-sensitive, both implementations recover the same topics — including th
 same partial recoveries on hard corpora. `parity/tlda_compare.py` runs the
 comparison when the reference package is available.
 
+## Running the upstream comparison
+
+The reference ([TensorLy TLDA](https://github.com/tensorly/tlda)) is not on PyPI
+and is not a topica dependency, so the parity script locates it at run time and
+skips cleanly when it is absent. Obtain it once and point `TOPICA_TLDA_REF` at
+the checkout:
+
+```bash
+git clone https://github.com/tensorly/tlda /path/to/tlda
+export TOPICA_TLDA_REF=/path/to/tlda
+python parity/tlda_compare.py     # runs the square-rank comparison; exits 0 if the ref is absent
+```
+
+`parity/tlda_ref.py` centralizes this lookup. Because the script exits 0 when
+`TOPICA_TLDA_REF` is unset, it is safe to wire into CI or a scheduled integration
+job: the job is a no-op wherever the reference is not installed, and a real
+comparison wherever it is. No machine-specific path is baked into the repository.
+
 ## Current evidence
 
 ### Upstream comparison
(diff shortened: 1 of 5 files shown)


nealcaren:

**Blocker 1 — the locator does not honor the env-var-only contract.** `load_reference_tlda()` only conditionally inserts the `TOPICA_TLDA_REF` checkout on `sys.path`, but then unconditionally attempts `__import__` of `tlda_final` / `tlda` / `tlda_wrapper`. When the env var is unset, a stray installed `tlda` package or a leftover local `tlda_wrapper.py` silently flips `HAS_REF` to `True` and runs the comparison against whatever is on the ambient path — which reintroduces exactly the machine-dependence this PR set out to remove. Fix: return `None` immediately when `reference_dir()` is `None`, and ideally verify the resolved module originates under the configured checkout.

**Blocker 2 — the recovery test leaks a process-global setting.** `_fit_and_score()` calls `topica.enable_experimental(True)` and never restores the prior value; the Rust flag is a global static, so this bleeds into later tests in the same pytest process. Wrap it in an autouse save/restore fixture like `tests/test_anchor.py:12-20`.

**Open question** — when `TOPICA_TLDA_REF` is set but its deps are broken, the `ImportError` is swallowed into a silent skip. A visible config error might be preferable so a misconfigured integration job does not look like a clean pass.

Verified sound: the recovery-gate floors sit safely below baseline (mean 0.50 vs 0.529, per-topic 0.35 vs 0.443) with both the simulation and TensorLDA seeded (2026), so no flakiness; `tlda_ref.py` has no hard-coded personal path.

nealcaren:
Addressed in the latest push (0e1c66f):

- **Blocker 1 (env-var-only contract):** `load_reference_tlda()` now returns `None` immediately when `TOPICA_TLDA_REF` is unset — it never probes the ambient path, so a stray installed `tlda` or a leftover `tlda_wrapper.py` can't flip `HAS_REF`. When the var is set, a resolved module is accepted only if its `__file__` lives under the configured checkout (`realpath(mod.__file__).startswith(realpath(ref)+os.sep)`), so an ambient install shadowing the name is rejected. Verified: unset → None, set-to-empty-dir → None.
- **Open question (silent skip on broken deps):** when the var is set but nothing usable resolves, it now warns to stderr (`… is set but no usable TLDA reference resolved under it …; treating as absent`) so a misconfigured integration job is visible.
- **Blocker 2 (leaked global):** removed the inline `enable_experimental(True)` and wrapped it in an autouse save/restore fixture mirroring `tests/test_anchor.py`, so the process-global gate no longer bleeds into later tests.

Recovery suite green (3 passed); preflight clean.

Issue #454, opened 2026-07-23
STS: add a reference-compatible fitting profile, beginning with adjusted kappa estimation


nealcaren:
## Summary

Topica STS implements the authors’ core likelihood, gradient, and Hessian, but its fitting defaults and several update choices do not yet reproduce the current reference R package. The largest gap is that the R package defaults to `kappaEstimation = "adjusted"`, whereas topica exposes only `"lasso"` and `"ridge"` and defaults to ridge.

This issue is about reference fidelity, not an allegation that the existing core STS math is incorrect.

## Reference behavior

The CRAN R package sts 1.4 defines:

- `kappaEstimation = "adjusted"` as the public default.
- `"adjusted"` as an adjusted aggregated Poisson regression followed by glmnet L1-path/AIC selection.
- a reference initialization that obtains eta and the prevalence covariance from the initial STM fit.
- damped kappa updates: each new coefficient is averaged with the previous coefficient.

Sources: R/sts.R and R/opt.kappa.R in https://github.com/cran/sts.

## Current topica behavior

- Default is `kappa_estimation="ridge"`; accepted values are `"ridge"` and `"lasso"`.
- The lasso option is a native IRLS/coordinate-descent approximation of the glmnet path, not glmnet itself.
- Initialization uses topica spectral beta recovery, with an identity initial covariance, rather than the reference STM eta/covariance initialization.
- The M-step replaces kappa directly rather than applying the reference half-step damping.

The existing offline STS gold test is useful but compares topica STS to a frozen R STM fit on a slim corpus. The live R-STS comparison needs the authors’ Poliblogs replication data, so it is not a portable CI gate.

## Proposed scope

- [ ] Add `kappa_estimation="adjusted"`, matching the reference adjusted aggregation and L1/AIC selection.
- [ ] Preserve the current ridge mode as an explicitly topica-native fast alternative; do not silently change existing user results.
- [ ] Add a `reference_compatible` fitting profile, or documented equivalent options, selecting R-aligned kappa behavior, initialization, and kappa damping.
- [ ] Decide whether the reference-compatible profile should require an external glmnet-compatible backend or whether a validated native numerical match is sufficient.
- [ ] Implement reference STM-derived initialization for eta and covariance, or document and validate a statistically equivalent substitute.
- [ ] Apply the reference kappa half-step update in the compatibility path.
[...]

PR #474, opened 2026-07-23
feat(sts): reference-fidelity fitting profiles (paper + CRAN) (#454)

nealcaren:
Addresses #454. STS implemented the authors' core likelihood/gradient/Hessian, but its fitting defaults reproduced neither R reference, and the docstring overstated that `"lasso"` matches R `sts` exactly.

## Two references, genuinely different

The **paper replication code** (Chen & Mankad 2024) and **CRAN `sts` 1.4** differ in exactly two ways; both share the same STM-derived initialization:

| | Default estimator | κ damping |
|---|---|---|
| **Paper** (`estimation="lasso"`) | `"lasso"` — plain group-mean aggregation | none |
| **CRAN** (`kappaEstimation="adjusted"`) | `"adjusted"` — φ-mass-weighted aggregation | half-step `(κ_new+κ_old)/2` |

So this ships **both**, via a `reference=` selector.

## What's added

- **`kappa_estimation="adjusted"`** — the CRAN public default: the same L1/AIC Poisson solve as `"lasso"` but with a φ-mass-weighted group mean of the sentiment covariate (each doc weighted by its topic-`t` token mass), reproducing the `weighted_alpha` reweighting in CRAN `opt.kappa.R`.
- **`reference="paper"` / `reference="cran"`** profiles — both use the STM-derived spectral-eta / sentiment-prior-variance-20 init; they differ in estimator + damping exactly as the references do. A conflicting explicit `kappa_estimation` raises. Default `reference="none"` keeps the fast topica-native ridge path unchanged.
- **ridge relabeled topica-native**; the overstated "matches R sts exactly" claim removed from the docstring + `.pyi`.

### A note on the CRAN `opt.kappa.R` inner loop
CRAN's `"adjusted"` wraps the reweighting in `while (kk <= max_inner_iter)` with `max_inner_iter <- 10`, but never reassigns `coef` inside the loop — so `delta` is identically zero and it breaks on iteration 1. The effective computation is a single φ-mass-weighted aggregation + one L1/AIC fit, which is what every published CRAN `sts` result actually computes. topica ports the *effective* behavior (documented in the code + `docs/replications/sts.md`).

## Validation

- **`parity/sts_kappa_glmnet.py`** validates the native Poisson L1/AIC κ solver head-to-head with R `glmnet` on the actual STS design (block-diagonal topic dummies + sentiment slopes): aligned cosine `> 0.999`, essentially bit-for-bit, with the occasional AIC-selection-boundary case. Skips cleanly (exit 0) when Rscript/glmnet are unavailable, so it is CI/scheduled-job safe.
- **Rust unit tests** for the weighted aggregation (hand-computed), half-step damping, and end-to-end adjusted / reference-compatible recovery — 208 lib tests pass.
- **Python tests** for all three estimators, both profiles, determinism, and the conflict/validation error paths.
- Full suite: **3626 passed, 30 skipped**; `cargo fmt`/`clippy`, `scripts/preflight.sh` (stub sync), `mkdocs build --strict` all green.

## Scope / honest limits
Full end-to-end R `sts` parity needs the authors' Poliblogs replication data (not a portable CI gate). What's committed and validated is the κ-solver kernel (vs glmnet) and the exact aggregation + damping arithmetic (Rust unit tests). The reference init is a documented statistically-equivalent substitute for STM's spectral `eta`/`invsigma`, not a byte-identical STM port. Item #454 checklist items left as follow-ups: a checked-in authors-data-derived end-to-end artifact.

### DIFF
--- a/docs/replications/index.md
+++ b/docs/replications/index.md
@@ -14,3 +14,4 @@ diverge, we say so.
 - [Structural Topic Model: the `stm` vignette](stm.md) (Roberts, Stewart & Tingley)
 - [Dynamic keyATM: U.S. Supreme Court opinions](keyatm-dynamic.md) (Eshima, Imai & Sasaki 2024)
 - [TensorLDA: implementation validation](tlda.md) (Kangaslahti et al. 2026)
+- [STS: reference-fidelity fitting profiles](sts.md) (Chen & Mankad 2024)
--- a/docs/replications/sts.md
+++ b/docs/replications/sts.md
@@ -1,0 +1,107 @@
+# STS: reference-fidelity fitting profiles
+
+`STS` is topica's Structural Topic and Sentiment-Discourse model (Chen & Mankad
+2024), which extends STM with a per-document, per-topic continuous
+sentiment-discourse latent. topica implements the authors' core likelihood,
+gradient, and Hessian. This page documents how topica's fitting defaults relate
+to the two reference implementations and how to select a reference-fidelity
+configuration.
+
+## Two references, one model
+
+There are two R implementations, and they differ in their fitting defaults:
+
+- The **paper replication code** (Chen & Mankad 2024, the *Management Science*
+  supplement) defaults to `estimation = "lasso"` and applies no damping to the
+  topic-word coefficients.
+- The **CRAN `sts` package** (1.4) defaults to `kappaEstimation = "adjusted"` and
+  damps each topic-word update against the previous one.
+
+Both share the same initialization: an STM anchor-word spectral fit
+(`max.em.its = 0`) supplies the initial per-document `eta`, the topic-block prior
+covariance (`invsigma`), and the initial topic-word matrix, with the sentiment
+block given prior variance 20.
+
+## What topica exposes
+
+The topica default is **topica-native, not a reference target**: a fast
+ridge-penalized Poisson κ estimator (`kappa_estimation="ridge"`) with topica's
+own spectral initialization. It is deterministic and supported, and it is the
+right choice when you want a fast, stable fit and are not trying to reproduce a
+specific R run.
+
+To reproduce an R run, select a `reference` profile:
+
+```python
+import topica
+
+model = topica.STS(num_topics=20, seed=1)
+
+# CRAN sts (kappaEstimation = "adjusted"):
+model.fit(docs, sentiment_seed=stars, prevalence=X, reference="cran")
+
+# Chen & Mankad 2024 replication code (estimation = "lasso"):
+model.fit(docs, sentiment_seed=stars, prevalence=X, reference="paper")
+```
+
+Both profiles use the reference-style initialization (STM-derived spectral `eta`,
+sentiment-block prior variance 20). They differ exactly as the two references do:
+
+| `reference` | κ estimator | κ aggregation | κ damping |
+|-------------|-------------|---------------|-----------|
+| `"none"` (default) | as `kappa_estimation` (ridge) | — | none |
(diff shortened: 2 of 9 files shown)


nealcaren:
Follow-up (82cd28c): the end-to-end R `sts` comparison turned out to be **portable** — `poliblog5k` ships with the `stm` package, so no replication data is needed.

Added `parity/sts_r_package_compare.py`: it fits the **live R `sts` package** and topica STS (`reference="cran"`) on the same `poliblog5k` corpus and compares aligned topic-word cosine at mean sentiment. Validated locally: **0.886** at 300 docs / K=5 / 8 iters, on an identical 2539-word vocabulary (floor 0.80). It's a statistical check (the two initialize independently), and skips cleanly (exit 0) when Rscript/`sts`/`stm` are absent. It complements the existing `sts_r_compare.py`, which compares against the authors' frozen RDS fit and needs `STS_REPL_DIR`.

This closes the last substantive gap from the #454 checklist (item 7's portable end-to-end fixture). `docs/replications/sts.md` corrected accordingly.

nealcaren:

**Blocker — `reference` + an explicit `kappa_estimation="ridge"` is silently accepted.** The conflict check uses `user_set_kappa = kappa_estimation != "ridge"`, so `reference="cran"` (or `"paper"`) paired with an explicit `kappa_estimation="ridge"` is treated as "not set" and silently overwritten with the profile estimator (`adjusted`/`lasso`). That contradicts the docstring/`.pyi` promise that "pairing it with a conflicting explicit `kappa_estimation` raises," and it means a user who asked for ridge silently runs a different estimator. Fix: make `kappa_estimation` a sentinel/`Optional` at the binding boundary so an explicit `"ridge"` is distinguishable from the default (then raise on the ridge+profile conflict too), or change the contract to "profiles always override." Add a `reference="cran", kappa_estimation="ridge"` test — the current conflict tests only cover the non-default estimators.

**Open question (documentation, non-blocking)** — the `phi_d`-derived prevalence logits are a documented statistically-equivalent substitute for STM spectral `eta`/`invsigma`, not an import of them. That is stated honestly in the code and `docs/replications/sts.md`; just keep it framed as an approximation.

Verified correct: `fit_sts` gained `reference_init` then `kappa_damping` and every call site (tests + binding) passes them in that order; `group_means_weighted` matches the φ-mass-weighted formula with the sentiment block correctly indexed at `a[k-1+t]`; `phi_d` is reset once per EM iteration before the serial reduction (no stale/double-count); reference init is finite-protected with a `1e-2` topic-block floor and sentiment variance 20; the glmnet parity hook (`_sts_poisson_lasso`) validates the shared solver.

nealcaren:
Addressed in the latest push:

- **Blocker (silent ridge override):** `kappa_estimation` is now a sentinel `Option<&str>` at the binding boundary (default `None` → resolves to `"ridge"`). `user_set_kappa = kappa_estimation.is_some()`, so an explicit `kappa_estimation="ridge"` is distinguishable from the default and conflicts with a reference profile like any other estimator. `reference="cran", kappa_estimation="ridge"` now raises (`… uses the "adjusted" estimator; remove kappa_estimation="ridge"…`) instead of silently running `adjusted`. The `.pyi` and docstring are updated to `str | None = None` and the honest contract.
- Extended `test_conflicting_estimator_and_profile_raises` with the `("cran","ridge")`/`("paper","ridge")` cases, and added `test_matching_explicit_estimator_and_profile_is_allowed` so passing the profile's own estimator explicitly still fits.

The `phi_d` prevalence-logit substitute stays framed as a documented statistically-equivalent approximation (no change). 29 STS tests pass; preflight (incl. stub sync) clean.

PR #475, opened 2026-07-24
fix(keyatm): validate public-Rust-API inputs (#418, item 3)

### LINKED ISSUE #418: keyATM review: dynamic slice-sampler retains rejected proposal + turbo_alpha_stride bias overclaim + Rust-API input validation
Independent correctness review of `src/keyatm.rs` (via the Codex second-opinion agent, `gpt-5.6-sol`), cross-checked against the source and the AJPS 2024 paper. The per-token `(z,s)` keyword-switch sampler, weighted-count bookkeeping, the base alpha slice-sampler's exact path, the dynamic Chib FFBS/HMM, and the AD-LDA parallel merge all verified correct. Two genuine bugs and several input-validation gaps surfaced; three flagged "R deviations" are intentional paper-faithful choices and are **not** in scope here.

## 1. Dynamic per-state alpha slice-sampler retains a rejected proposal

`src/keyatm.rs:2031-2049` — the `for _ in 0..MAX_SHRINK_TIME` shrinkage loop. If every proposal is rejected and the exact-equality branch (`new_p == previous_p`) never fires, the loop falls through with `alpha[k]` left at the **last rejected proposal** rather than restored to `keep`. Only the equality branch restores `keep`.

Rare (requires exhausting `MAX_SHRINK_TIME` shrinks), but a real slice-sampler invariant violation: a fully-rejected slice step must return the current value, not a rejected one.

**Fix:** restore `alpha[k] = keep` on loop fall-through (or track acceptance and restore if never accepted).

## 2. `turbo_alpha_stride` is not an unbiased estimator of alpha

`src/keyatm.rs:1950-1958` — the data term subsamples deterministically (`d += stride` from `doc_start`) and scales by `total/sampled`. Two problems:

- The subset is **deterministic and order-dependent**: it always takes documents at fixed stride offsets in current corpus order, so document ordering systematically shifts the estimated alpha. A random subset would be needed for the inverse-fraction scaling to be unbiased.
- Slice-sampling against this substituted target samples the **subset** posterior, not the full-data posterior.

The docstring claim of an "unbiased estimate of alpha" (`src/python/mod.rs`, `turbo_alpha_stride` doc, ~line 12996) overclaims. Default `stride=1` is unaffected.

**Fix:** either randomize the sampled subset each sweep, or soften the documentation to "approximate speed knob that biases the alpha estimate; use stride=1 for exact."

## Input-validation gaps (public Rust API only; the Python wrapper guarantees these)

- **Dynamic `time_index` contiguity** (`fit_keyatm_dynamic`, ~2112): only checked non-decreasing, not gap-free `0..max`. A missing period leaves zero-valued range entries and can underflow at `time_doc_start[t+1] - 1`.
- **Keyword-topic contiguity**: the sampler assumes keyword topics occupy `0..num_keyword_topics`; an interspersed empty keyword list routes switch factors to the wrong topics.
- **Duplicate keywords** within one topic's list are not rejected; `KeywordIndex` can score one position but restore counts through a different matching position.
- Public Rust fits should validate positive hyperparameters, nonzero topic count, and in-range vocab/keyword IDs rather than relying on the Python layer.

## Lower-severity correctness notes
- Covariate SE (`~1853`) is computed from final counts, but L-BFGS optimization is periodic; if the final sweep is not an optimize step, lambda need not be stationary for those counts, weakening the inverse-Hessian SE interpretation.
- Post-hoc +/-5 clipping of lambda (`~1825`) is not box-constrained optimization; with correlated features the clipped vector can violate the constrained KKT conditions. Consider L-BFGS-B / projected optimization.

## Explicitly out of scope (intentional paper-faithful design, not bugs)
- Weights use paper Eq. 6 `-log2(freq/N)`, not R keyATM 0.5.5's smoothed `(freq+1)/(N+V)`.
- theta reported from weighted `ndk` (paper Eq. 9), not R's `n_dk_noWeight`.
- Covariate model uses periodic MAP via the DMR lambda prior, not R's per-iteration lambda slice sampler.

These are correct against the paper and only diverge from seed-for-seed R parity.

## Suggested tests
- Slice exhaustion: scripted RNG/target that rejects `MAX_SHRINK_TIME` proposals; assert the prior alpha is restored.
- Turbo permutation: permute document order with counts fixed; the exact (`stride=1`) alpha log-likelihood must be invariant, the strided approximation will not be.
- Count reconstruction after each sequential and parallel sweep: rebuild `ndk/nkw/nkx/nk0/nk1` from assignments+weights and compare.
- Input-invariant tests: reject gapped times, interspersed keyword/regular topics, duplicate keywords, out-of-range IDs, nonpositive priors.

_Filed from a Codex review pass; findings #1, #2, and the underflow path independently verified against the source._

nealcaren:
Completes the keyATM correctness review (#418). Items 1 (rejected-proposal slice step) and 2 (turbo_alpha_stride bias docs) landed in #464; the covariate-SE note in #468. This PR adds the remaining **item 3** input-validation guards to the public Rust fits, which previously assumed the preconditions the Python wrapper guarantees and could silently corrupt counts or underflow for a direct Rust caller.

## Shared validation (in `init_state`, covers all four public fits)
- **Keyword-topic contiguity** — the sampler derives `num_keyword_topics` as the count of non-empty keyword lists but then assumes those topics are `0..m`, so an empty list *before* a non-empty one misroutes keyword-switch factors. Now rejected.
- **Duplicate keywords** within a topic (a duplicate lets `KeywordIndex` score one position but restore counts through another), **out-of-range keyword ids**, **out-of-range document word ids**, **positive** `beta`/`beta_key`/`gamma1`/`gamma2`, `num_topics >= 1`.

## Per-fit guards
- base / cvb0: reject a non-positive fixed `alpha` (base only when `estimate_alpha=false`, since `alpha` is only a start when estimated — verified by an accept-test).
- dynamic: reject non-positive `eta1/eta2/eta1_reg/eta2_reg` and a **gap-free `time_index`** — every segment `0..num_time` must be non-empty and it must start at 0 (a gap left `time_doc_start` entries at 0 and could underflow `time_doc_start[t+1]-1`).

## Python layer
A seeded topic whose keywords are all pruned / out-of-vocabulary would leave a hole in the keyword-topic prefix; `KeyATM.fit` now raises a clear `ValueError` instead of letting the core's invariant guard panic across the FFI boundary. (Previously this silently misrouted.)

## Tests
- 10 Rust guard tests (`should_panic` for each invariant + the `estimate_alpha` exemption).
- A Python clean-error test for the all-OOV seeded topic.
- The conformance harness now fits its placeholder-keyword seed models (KeyATM/SeededLDA) on a corpus that contains those keywords — scoped so the shared toy corpus (and every other model's behavior) is untouched.

## Verification
- `cargo test --lib`: 214 passed (+10). `cargo fmt`/`clippy`: clean. `scripts/preflight.sh`: green.
- Full pytest: 3625 passed, 30 skipped (the one transient NMF failure was my earlier over-broad corpus edit, fixed by scoping the seed-model corpus).

Closes #418.

### DIFF
--- a/src/keyatm.rs
+++ b/src/keyatm.rs
@@ -1226,6 +1226,10 @@ fn validate_keyatm_inputs(
     gamma2: f64,
 ) {
     assert!(num_topics >= 1, "num_topics must be at least 1");
+    // An empty corpus has no counts to sample and, with estimate_alpha, the base
+    // fit passes `docs.len() - 1` (usize) to the α sampler, which would wrap to
+    // usize::MAX. Reject it up front so every public fit rejects it identically.
+    assert!(!docs.is_empty(), "docs must be non-empty");
     assert_eq!(
         keywords.len(),
         num_topics,
@@ -2222,7 +2226,15 @@ pub fn fit_keyatm_dynamic<R: Rng>(
             "documents must be sorted by time_index (non-decreasing)"
         );
     }
-    let num_time = time_index.iter().copied().max().map(|m| m + 1).unwrap_or(0);
+    let num_time = time_index
+        .iter()
+        .copied()
+        .max()
+        .map(|m| {
+            m.checked_add(1)
+                .expect("time_index entry is usize::MAX; segment count overflows")
+        })
+        .unwrap_or(0);
     assert!(
         num_time >= num_states,
         "num_time ({num_time}) must be >= num_states ({num_states})"
@@ -3414,6 +3426,15 @@ mod tests {
 
     fn two_docs() -> Vec<Vec<u32>> {
         vec![vec![0u32, 1], vec![1u32, 0]]
+    }
+
+    #[test]
+    #[should_panic(expected = "docs must be non-empty")]
+    fn rejects_empty_corpus() {
+        // With estimate_alpha=true and iters>0 the base fit passes
+        // `docs.len() - 1` to the α sampler; an empty corpus would wrap it to
+        // usize::MAX. The validator rejects it first.
+        run_base(&[], 2, 1, &[vec![0]], 0.1, 0.1, true);
     }
 
     #[test]
--- a/src/keyatm.rs
+++ b/src/keyatm.rs
@@ -1115,12 +1115,15 @@ pub fn fit_keyatm<R: Rng>(
     alpha_stride: usize,
     rng: &mut R,
 ) -> KeyAtmModel {
-    assert_eq!(
-        keywords.len(),
-        num_topics,
-        "keywords length must equal num_topics"
-    );
-
+    // A fixed symmetric α (estimate_alpha = false) must be a valid Dirichlet
+    // parameter; when α is estimated the passed value is only a start and is not
+    // constrained here (#418, item 3).
+    if !estimate_alpha {
+        assert!(
+            alpha.is_finite() && alpha > 0.0,
+            "alpha must be finite and positive when estimate_alpha is false (got {alpha})"
+        );
+    }
     let (mut model, mut assignments, ki) = init_state(
         docs, num_types, num_topics, keywords, alpha, beta, beta_key, gamma1, gamma2, weights, rng,
     );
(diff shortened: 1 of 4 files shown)


nealcaren:

**Blocker — an empty corpus still underflows in the base fit.** With `docs=[]`, `estimate_alpha=true`, `iters>0`, `fit_keyatm` calls `dyn_sample_alpha_state(&mut alpha_vec, num_keyword_topics, 0, docs.len() - 1, ...)` at `src/keyatm.rs:1169`, and `docs.len() - 1` (usize) wraps to `usize::MAX`. This is exactly the "direct Rust caller could underflow" precondition class this PR sets out to guard, so the shared validator should reject it. Fix: add a non-empty-`docs` check to `validate_keyatm_inputs` plus a `should_panic` regression test.

**Minor** — the dynamic fit computes `num_time = max + 1` before the new gap check runs, so a `time_index` containing `usize::MAX` overflows first. Consider `checked_add(1)` for a clean panic.

Verified correct: keyword-topic prefix contiguity (a real sampler invariant — `num_keyword_topics` is a count while the switch logic assumes topics `0..m`), duplicate/out-of-range keyword ids, out-of-range doc word ids, the gap-free `time_index` underflow guard, the alpha exemption (base rejects a non-positive fixed α only when `estimate_alpha=false`; cvb0 always requires positive α), all four public fits reaching `init_state`, and the removed `keywords.len() == num_topics` asserts being subsumed.

nealcaren:
Addressed in the latest push:

- **Blocker (empty corpus underflow):** `validate_keyatm_inputs` now asserts `!docs.is_empty()` ("docs must be non-empty"). Since every public fit routes through `init_state` → this validator, they all reject an empty corpus identically, before the base fit can pass `docs.len() - 1` to the α sampler. Added a `should_panic` regression (`rejects_empty_corpus`) that drives the exact `estimate_alpha=true, iters>0` path.
- **Minor (num_time overflow):** the dynamic fit now uses `checked_add(1)` for `num_time = max + 1`, so a `time_index` entry of `usize::MAX` panics with a clear message instead of wrapping before the gap check runs.

25 keyATM Rust tests pass; Python keyATM/convergence suites green (217 passed); preflight clean.

nealcaren:
Second review pass (Claude + an independent Claude reviewer, since Codex is rate-limited until Jul 29). The empty-corpus + `num_time` overflow fixes are **verified correct** — the empty-docs assert is reached by all four public fits via `init_state`, and `checked_add(1)` is placed before `num_time` is used. But the pass surfaced one **remaining blocker** worth closing before merge:

**Duplicate keyword strings panic across the FFI instead of raising `ValueError`.** `seed_word_ids` (`src/python/mod.rs:11840`) maps keyword strings to ids without deduplicating, so a plausible Python input like

```python
topica.KeyATM({"econ": ["tax", "tax", "jobs"]}, num_topics=4).fit(docs)
```

produces `[tax_id, tax_id, jobs_id]` and reaches this PR's new duplicate check in `validate_keyatm_inputs` (`src/keyatm.rs:1263`), aborting with `pyo3_runtime.PanicException: topic 0 lists keyword id 0 more than once` (reproduced on a branch build). Two distinct surface forms that stem to the same token hit this too.

This is the exact failure mode the PR already guards against for the empty-list case (`src/python/mod.rs:13506`, "Fail clearly here rather than let the core's invariant guard panic across the FFI boundary") — the duplicate case just slipped through, so the guard is internally inconsistent. Small fix, either:
- dedup in `seed_word_ids` (duplicate seed words are semantically harmless), or
- extend the Python guard to detect duplicate resolved ids and raise a clear `ValueError`.

Add a Python test for the duplicate-keyword input alongside the existing all-OOV one.

Everything else verified clean: `cargo test --lib keyatm` 25 passed; `pytest test_keyatm_workflow / test_input_validation / test_convergence_interface` 255 passed, 10 skipped; out-of-range ids, non-contiguous keyword prefix, non-positive hyperparameters, and the estimated-alpha exemption all correctly handled.

Issue #420, opened 2026-07-23
Add a metamorphic/invariant test suite to cover the parity blind spot (permute-order, extend-sweeps, reconstruct-counts, dense-conditional, boundary)


nealcaren:
Three recent code-level reviews (LDA #417, keyATM #418, DMR #419) surfaced correctness bugs that our cross-implementation **parity** checks structurally cannot catch, because parity is an equivalence test against one reference config on benign data. It validates the estimator's central tendency (topic overlap, coherence, held-out likelihood, lambda point estimates) and is blind to:

1. **Tail/edge regimes** — tiny alpha (`log_gamma -> +inf`), extreme covariate scale (unbounded `exp`), degenerate settings (`num_samples=0`). Parity fixtures are curated and never hit these.
2. **Uncertainty machinery** — SE constructions with no comparable reference oracle (R keyATM samples lambda; R stm uses method-of-composition), so our observed-information SEs were never diffed and the point estimate they hang off of is correct.
3. **Second-order effects** — the SE-at-drifted-counts bug leaves lambda and topic-word bit-identical; only the error bar moves, below any parity threshold.
4. **Non-parity paths** — approximate speed knobs (`turbo_alpha_stride`, `turbo_merge_every`, multithread) and the public Rust API's input validation; parity always runs the exact single-thread path through the pre-validating Python wrapper.

## Proposal: a metamorphic / invariant test suite

These bugs are catchable **without any reference implementation** via metamorphic testing — assert properties that must hold regardless of the "correct" numeric answer. Build a shared harness (`tests/test_invariants.py` + Rust `#[cfg(test)]` helpers) covering, per applicable model:

**Invariance (transform input/schedule, output must not move):**
- Document-order permutation: exact (`stride=1`, single-thread) alpha log-likelihood, lambda, and topics must be invariant to doc order. (Catches keyATM turbo bias #418.)
- Extend-by-non-optimize-sweeps: identical seed, `iters=m` vs `iters=m+r` where the extra `r` sweeps contain no optimize step -> identical lambda must yield identical SE. (Catches DMR/keyATM SE-timing #419/#418.)
- Sampling-control invariance: change only `num_samples`/`sample_interval` -> lambda and SE unchanged. (DMR #419.)

**Reconstruction (internal state must equal a from-scratch rebuild):**
[...]

Issue #419, opened 2026-07-23
DMR review: SE Hessian evaluated at drifted counts (cross-model w/ keyATM-cov) + log_gamma +inf for tiny alpha + unbounded exp


nealcaren:
Independent correctness review of DMR (`src/dmr.rs` + the binding in `src/python/mod.rs`), via the Codex second-opinion agent (`gpt-5.6-sol`), with all code-logic claims re-verified against the source. The FD-verified gradient/Hessian and the sparse leave-one-out sampler are confirmed correct, and coefficients are already on the original covariate scale (standalone DMR does not standardize, so there is no de-standardization bug). Three real defects plus three lower-severity notes.

> Note: the review agent ran in a read-only sandbox and could not build/run topica; its specific "reproduced" numbers were discarded. Every finding below was re-verified by reading the code.

## 1. [cross-model] `feature_effect_se` is evaluated at counts that do not match the returned lambda

`src/python/mod.rs`: lambda is L-BFGS-optimized only on scheduled sweeps (`iter % optimize_interval == 0`, past `burn_in`) — sparse `~3866`, Warp `~3736`, CVB0 `~3683`. But the SE (inverse observed information) is computed once at the end from whatever counts exist then — sparse `~3981`, Warp `~3821`, CVB0 `~3700`.

For the sparse and Warp paths it is worse: after the training loop ends, `num_samples x sample_interval` **additional** Gibbs sweeps run (sparse `~3941-3944`, Warp `~3796`) that mutate the counts, and only then is `dmr_lambda_se` called. So the Hessian is evaluated at counts that have drifted away from the counts lambda was last optimized against — it is not `J^-1` at the optimum for the returned lambda.

Also: when `optimize_interval == 0`, `burn_in >= iters`, or `lbfgs_iters == 0`, lambda may never be optimized at all, yet an SE is still returned. L-BFGS reports no convergence status, so hitting the iteration cap also emits an SE without establishing stationarity.

**This is a shared-structure issue.** keyATM-cov (#418) has the same "periodic optimize, SE from final counts" pattern, and any future model on the DMR prior (GDMR, etc.) inherits it. Worth a single shared fix: retain the sufficient statistics from the last successful optimization and evaluate both gradient and Hessian on exactly those, or re-optimize once immediately before computing the SE; assert the penalized gradient norm is below tolerance and return `feature_effect_se=None` otherwise.

## 2. Local `log_gamma` returns `+inf` for very small alpha

[...]

PR #476, opened 2026-07-24
test(invariants): reference-free metamorphic suite for the count-Gibbs core (#420)

nealcaren:
First slice of the #420 metamorphic/invariant harness. Cross-implementation **parity** validates central tendency against one reference config on benign data; it is structurally blind to tail regimes, uncertainty machinery, SE-only second-order effects, and non-parity paths — the blind spot that let #417/#418/#419 through. These tests assert properties that must hold **regardless of the correct numeric answer**, so they need no reference implementation.

## `tests/test_invariants.py` (marker `invariants`, runs by default)
Covers the shared count-Gibbs core — LDA / DMR / keyATM / SAGE / GDMR:

- **Boundary / finiteness** — outputs stay finite and normalized (`topic_word` last-axis and `doc_topic` rows sum to 1) on a normal corpus, **with an empty document** mixed in, and at **K=1** — or the model raises a clean error (never a silent NaN). Handles SAGE's 3D grouped `topic_word` via a last-axis normalization check.
- **Determinism** — same seed is bit-identical.
- **Sampling-control invariance** — `num_samples` / `sample_interval` average the topic-word posterior and drive the SE, but the covariate point estimate (`feature_effects`) is a MAP optimum of the counts and must be **bit-identical** across sampling settings (confirmed max|Δ| = 0).
- **SE stationarity** — `feature_effect_se` is a finite array or `None`, never an infinity, and the point estimate it hangs off of is finite (#419's "SE only when earned").

## Wiring
- Marker registered in `tests/conftest.py`; `just test-invariants` runs only these so a failure reads as `invariant: <property>` rather than a numeric drift.
- Runs by default (unlike `slow`/`parity`) since it's fast (~8s) and catches regressions.

## Scope
This is **slice 1**. The reconstruction (rebuild count tables from token assignments == live tables) and exact-dense-conditional invariants — which need internal count-table access and correct handling of #417's packed `(count, topic)` representation — are the tracked **Rust-level slice 2**. #420 stays open for them.

## Verification
- `-m invariants`: 24 passed. Full suite: **3649 passed, 30 skipped**. `scripts/preflight.sh`: green.
- The invariants were each empirically confirmed to hold on current `main` before being asserted.

### DIFF
--- a/justfile
+++ b/justfile
@@ -59,6 +59,11 @@ test-rust-features:
 test-py:
     {{py}} -m pytest tests/ -q
 
+# Reference-free metamorphic / invariant checks only (issue #420) — a failure
+# reads as "invariant: <property>" rather than a numeric drift.
+test-invariants:
+    {{py}} -m pytest tests/ -m invariants -q
+
 # Every test surface: Rust core, feature-gated Rust, and Python.
 test: test-rust test-rust-features test-py
 
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -17,6 +17,13 @@ def pytest_configure(config: pytest.Config) -> None:
         "fit tens of thousands of documents and are opt-in: run with "
         "`-m slow`; the default suite deselects them.",
     )
+    config.addinivalue_line(
+        "markers",
+        "invariants: reference-free metamorphic / invariant check (issue #420). "
+        "Asserts properties that must hold regardless of the correct numeric "
+        "answer (finiteness, normalization, sampling-control invariance, SE "
+        "stationarity) to cover the blind spot cross-implementation parity leaves.",
+    )
 
 
 # ---------------------------------------------------------------------------
--- a/tests/test_invariants.py
+++ b/tests/test_invariants.py
@@ -1,0 +1,186 @@
+"""Reference-free metamorphic / invariant tests for the count-Gibbs core (#420).
+
+Cross-implementation *parity* validates central tendency against one reference
+config on benign data; it is structurally blind to tail regimes, uncertainty
+machinery, second-order (SE-only) effects, and non-parity paths. These tests
+assert properties that must hold **regardless of the correct numeric answer**, so
+they need no reference implementation:
+
+- finiteness / normalization on degenerate inputs (empty docs, K=1);
+- sampling-control invariance (num_samples/sample_interval move only the SE, never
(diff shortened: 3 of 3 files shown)


nealcaren:

**Blocker — the covariate/SE invariants are vacuous under their own settings.** `DMR` and `GDMR` build with the default `burn_in=200`, but the tests fit only `iters=25` (`_ITERS`) and `iters=40`. With `burn_in >= iters`, λ is never optimized: `feature_effects` stays at its initialization and `feature_effect_se` is `None` (the documented contract). Consequences:
- `test_sampling_control_does_not_move_the_point_estimate` compares two unoptimized (all-zero) arrays, so it passes regardless of any real sampling-vs-point-estimate bug.
- `test_se_is_none_or_finite_never_silent_nan` hits the `if se is None: return` branch immediately and asserts nothing.

Neither can catch a #419-style regression, which is the blind spot this suite is meant to cover. Fix: fit the covariate models with `burn_in < iters` (or reuse the already-optimized fixture in `tests/test_dmr_feature_effect_se_419.py`) so λ optimizes and the SE is non-`None`, then assert a finite SE.

**Minor** — the SE test docstring ("finite ... never a silent NaN") contradicts its body, which permits NaN entries and rejects only infinities. Reconcile the two.

Verified sound: finiteness/last-axis normalization, same-seed bit-identity, the sampling-control invariant itself (valid once λ is optimized), the `invariants` marker registration, and that it runs under default `pytest -q`.

nealcaren:
Addressed in the latest push:

- **Blocker (vacuous covariate invariants):** DMR/GDMR now build with `optimize_interval=10, burn_in=10`, below the `iters=25/40` the invariants fit at, so λ is a real MAP optimum. Verified locally: `feature_effect_se` is non-None for both, and `feature_effects` is nonzero (max ≈0.99 DMR / ≈1.93 GDMR). The sampling-control test now asserts `feature_effects` moved off its zero init before checking sampling-invariance, and the SE test asserts the SE is actually emitted (non-None) so the #419 path is exercised.
- **Minor (docstring contradiction):** reconciled — the module/test wording now states an SE, when emitted, must be infinity-free, while an individual entry may be an advertised NaN for a clamped/unidentified effect.

Full suite green (24 passed), preflight clean.

Issue #456, opened 2026-07-23
SeededLDA: restore seededlda-compatible seed priors and validate weight


nealcaren:
## Summary

Topica SeededLDA has internally coherent asymmetric-prior Gibbs sampling, but it is not faithful to the public koheiw/seededlda R package that the documentation names as its reference. Its seed pseudocount construction, defaults, and initialization differ materially. In addition, `weight` accepts invalid values and can produce negative or non-finite topic-word probabilities.

## Direct reference evidence

The installed seededlda R package was run on a three-document corpus in which `tax` occurs 7 times and `war` 6 times, with `weight = 0.01`. Its internal seed matrix was:

```text
         tax war
econ       7   0
conflict   0   6
```

This follows `tfm()` in R/seededlda.R: the package groups the corpus, computes each matched seed word count, then uses `count * weight * 100` as the topic-word pseudocount.

Topica instead gives every in-vocabulary seed word the fixed mass `weight * 100`; in this example, both receive 1.0. This changes the relative influence of seed words within a seeded topic and is not a mere global rescaling.

Sources: https://github.com/koheiw/seededlda/blob/master/R/seededlda.R and https://github.com/koheiw/seededlda/blob/master/src/lda.h

## Current fidelity gaps

- Seed pseudocounts: reference uses corpus-frequency-scaled `count * weight * 100`; topica uses fixed `weight * 100`.
- Defaults: reference `alpha = 0.5`, `beta = 0.1`; topica `alpha = 0.1`, `beta = 0.01`.
- Initialization: reference initializes all tokens randomly and injects the seed matrix as pseudocounts; topica assigns seed-word tokens directly to their seed topics.
- `weight = 0`: reference has no seed matrix and follows ordinary random LDA initialization; topica still uses seeded initialization.
- Seed matching: the R package supports quanteda dictionary matching (glob/regex/fixed and case controls); topica accepts literal word strings only.
- Validation: existing planted gold labels SeededLDA as having no external reference, even though seededlda is an installed, runnable reference package.

## Correctness bug: invalid weight is accepted

The constructor validates alpha and beta but not weight. Examples reproduced locally:

- `weight=-0.01` fits and yields a negative topic-word probability (50.5 and -49.5 in one row on a tiny corpus).
- `weight=NaN` or `weight=Infinity` fits and yields non-finite probabilities.
- `weight=1.5` is accepted, while seededlda rejects weights outside [0, 1].

Invalid weights must be rejected before fitting.

## Proposed scope

[...]

PR #479, opened 2026-07-24
fix(seededlda): reject non-finite doc_topic_prior; dedupe CVB0 seed denominator (#456)

nealcaren:
Addresses the two verified/confirmed correctness bugs from the #456 review follow-up (same non-finite/seed-mass class as #472/#473).

## Bug 1 — non-finite `doc_topic_prior` slips through (verified in the issue)

`fit`'s per-document prior guard was `a <= 0.0`. `NaN <= 0.0` and `+inf <= 0.0` are both `false`, so a non-finite `doc_topic_prior` passed validation and contaminated sampling/output. Now uses the existing `finite_pos` helper (`!finite_pos(a)` → reject non-finite or non-positive).

## Bug 2 — CVB0 under-normalizes a seeded topic with duplicate seeds (confirmed by reading the code)

CVB0's `set_seeds` added `ws.len() * seed_weight` to the per-topic normalizer `beta_sum_k`, but the numerator applies the boost via a boolean membership test (`beta + seed_weight` once per word). A duplicated (or out-of-vocab) seed word therefore inflated the denominator and under-normalized the seeded topic's φ. The **sparse** (`seeded.rs`) and **warp** (`warplda.rs`) backends already dedupe to distinct in-vocab seeds — this brings CVB0 in line.

## Tests
- `topica-core` CVB0 `set_seeds_dedupes_the_denominator`: a duplicated seed word produces the same `beta_sum_k`/`inv_seeds` as the distinct-count seeds.
- Python `test_seededlda_rejects_nonfinite_doc_topic_prior[NaN/inf]`.

Local: `cargo test -p topica-core cvb0` green (4 passed); seededlda Python suites green (52 passed); preflight clean.

## Out of scope (remains open on #456)
The larger seededlda **R-fidelity** items — corpus-frequency-scaled seed mass (`count·weight·100`), reference `alpha=0.5`/`beta=0.1` defaults, a documented `reference_compatible` profile, an R `seededlda` parity fixture, and quanteda-style dictionary matching — are a separate, larger port (design + the R package) and are not touched here.

### DIFF
--- a/src/python/mod.rs
+++ b/src/python/mod.rs
@@ -12066,8 +12066,13 @@ impl SeededLDA {
                         "each doc_topic_prior row must have num_topics entries",
                     ));
                 }
-                if rows.iter().any(|r| r.iter().any(|&a| a <= 0.0)) {
-                    return Err(PyValueError::new_err("doc_topic_prior entries must be > 0"));
+                // `a <= 0.0` is false for NaN and +inf, so a non-finite prior would
+                // slip through and contaminate sampling/output. Require finite and
+                // positive (#456, same non-finite class as #472/#473).
+                if rows.iter().any(|r| r.iter().any(|&a| !finite_pos(a))) {
+                    return Err(PyValueError::new_err(
+                        "doc_topic_prior entries must be finite and > 0",
+                    ));
                 }
                 Some(rows)
             }
--- a/tests/test_input_validation.py
+++ b/tests/test_input_validation.py
@@ -62,6 +62,19 @@ def test_seededlda_rejects_bad_weight(bad):
         topica.SeededLDA(seeds, weight=bad)
 
 
+@pytest.mark.parametrize("bad", [NAN, INF])
+def test_seededlda_rejects_nonfinite_doc_topic_prior(bad):
+    # #456: the doc_topic_prior guard was `a <= 0.0`, which is false for NaN and
+    # +inf, so a non-finite prior slipped through and contaminated the fit. It must
+    # be rejected up front (same non-finite class as the weight guard above).
+    seeds = {"sports": ["ball", "goal"], "politics": ["vote", "law"]}
+    docs = [["ball", "goal", "vote"], ["law", "vote", "ball"], ["goal", "law", "ball"]]
+    m = topica.SeededLDA(seeds, seed=1)
+    prior = [[1.0, bad] for _ in docs]
+    with pytest.raises(ValueError, match="finite"):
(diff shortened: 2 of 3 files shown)


nealcaren:
Reviewed this against the source (independent pass with Gemini via the Antigravity CLI, then spot-verified the claims by hand). Both fixes check out; one scope note worth flagging.

**Bug 2 (CVB0 seed denominator) — correct and complete.** The numerator boost is a boolean membership test (`s.inv_seeds[w].binary_search(&(t as u32)).is_ok()`, `topica-core/src/cvb0.rs:285`), so the corrected denominator `V·β + (#distinct in-vocab seeds)·seed_weight` is right, and the HashSet dedupe now matches the warp (`src/warplda.rs:166-172`) and sparse (`src/seeded.rs:347-355`) backends. The new `set_seeds_dedupes_the_denominator` test exercises exactly this.

**Bug 1 (non-finite `doc_topic_prior`) — the fix is correct, but the same bug class is still open in sibling constructors.** `finite_pos` (`x.is_finite() && x > 0.0`) is the right guard for the `SeededLDA.doc_topic_prior` site. But the bare `a <= 0.0` pattern (which lets `NaN`/`+inf` through) still lives in several other PyO3 constructors:

- `src/python/btm.rs:136,140`
- `src/python/rtm.rs:137`
- `src/python/tlda.rs:168,177,186`
- `src/python/pltm.rs:190,194`
- `src/python/mod.rs:12756` (`FASTopic` `theta_temp`)

For contrast, `disclda.rs`, `neural.rs`, `party_embeddings.rs`, and `mod.rs:12488/12518/12523` were already hardened with `!is_finite() ||` — so the codebase is mid-migration and this PR patches one site of a known class. Not a blocker for merging what's claimed here (the PR is explicitly scoped to the `doc_topic_prior` site), but worth a fast-follow so the other entry points don't stay exposed.

Optional follow-ups:
- Extend `tests/test_input_validation.py` with `NaN`/`inf` cases for `BTM`, `PolylingualLDA`, `RTM`, `TLDA`, and `FASTopic`.
- Add an out-of-vocab seed index (e.g. `vec![0, 0, 1, 9999]`) to `set_seeds_dedupes_the_denominator` and assert `beta_sum_k == V·β + 2·seed_weight` — the current test covers duplicates but not the `w >= v` path the fix also handles.
- Consider a shared `ensure_finite_pos!(val, "name")` helper across the bindings so new params can't reintroduce the raw `<= 0.0` check.

PR #480, opened 2026-07-24
feat(seededlda): R-seededlda-faithful seed prior as the default (#456)

nealcaren:
Makes topica **SeededLDA faithful to the koheiw/`seededlda` R package by default**, closing the main fidelity gaps in #456. (Per the discussion, R-compatible is the new default; the original scheme is named by what it does, not "topica".)

## What changed

**Seed-mass construction — reproduced exactly.** Each seed word's prior pseudocount now scales with its corpus frequency: `count × weight × 100`, the package's `tfm` construction. Verified against live `seededlda:::tfm` to floating point (max |Δ| = 1.4e-14). The scalar `seed_weight` is generalized to per-(topic, word) masses threaded through the **sparse, warp, and cvb0** backends.

**Initialization.** The frequency scheme initializes tokens at random — seeds enter only through the persistent β pseudocount, exactly as R adds its seed matrix to `nw` once and never removes it (`phi[k][w] = (n_wk + β + m_{k,w}) / (n_k + Vβ + Σ_w m_{k,w})`).

**Defaults.** `alpha=0.5`, `beta=0.1` (the package defaults); `weight=0` → plain random LDA.

**New `seed_prior=` parameter.**
- `"frequency"` (default) — R-faithful: corpus-frequency-scaled mass + random init.
- `"uniform"` — topica's original: flat `weight × 100` per seed word + anchored init.

**New `seed_prior_matrix` property** exposes the exact per-(topic, word) pseudocounts a fit used, so the seed-mass construction is inspectable and testable offline.

## Validation (replaces the old "no external reference" planted gold)

- `parity/seededlda_gold.py` freezes R seededlda's two-seed keyword-phi noise floor **and** its exact `tfm` matrix on a fixed poliblog subsample. The offline test asserts topica (1) reproduces R's `tfm` exactly and (2) clears R's own seed-to-seed cosine floor: **topica 0.936 vs R-self 0.973, bar 0.823**. Runs in CI **without Rscript**.
- `parity/seededlda_r_compare.py` — live human-readable reporter (skips cleanly without R).
- New offline **exact** frequency seed-mass contract in `test_seeded_gsdmm_contracts` (repeated seed words, unequal corpus frequency). The two existing exact contracts move to `seed_prior="uniform"` (the scheme they describe).

`EmbeddingLDA` pins `seed_prior="uniform"` (it anchors on embedding-derived seeds, not the package's corpus-frequency prior). Weight validation (finite, [0,1]) already landed in #472. README / docs / stub updated.

## Tests
- Rust: `cargo test --lib` (204) + `topica-core` (40) green.
- Python: full suite green except one unrelated environment flake (`test_top2vec_no_heavy_import_at_test_time`, a subprocess import check — passes in isolation; 3437 passed before it). All SeededLDA/embedding suites green.
- Preflight (fmt + clippy + model tables + stub sync) clean.

## Notes
- Touches `topica-core/src/cvb0.rs` `set_seeds`, which **overlaps PR #479** (the CVB0 duplicate-seed dedup fix on the same issue). This PR's per-word-mass rewrite subsumes that dedup; whichever lands second needs a trivial rebase.
- **Out of scope (remaining #456 follow-up):** quanteda-style glob/regex/fixed dictionary matching for seed words.

Closes the seed-prior/defaults/init/reference-fixture items of #456.


nealcaren:
**Independent review (Codex, cross-family):** verdict *needs-changes*. Resolved:

- **Warp seed proposal (latent correctness, fixed):** WarpLDA's word-proposal drew the seeded topic *uniformly* over `inv_w` rather than proportional to the per-(topic,word) mass `m_{k,w}`. Correct today only because every word's mass is equal across its seeded topics (`mass = corpus_freq·weight·100` is topic-independent), but the generalized per-word-mass API should be robust — now samples the seeded topic by cumulative mass. Recovery unchanged (equal masses → same distribution); Rust warp tests green.
- **Init contract (documentation, clarified):** documented that uniform's *anchored* init is a `sampler="sparse"` detail — `warp`/`cvb0` always initialize at random and seed only through the β pseudocount.
- **weight=0 under uniform anchors (Codex flagged as a bug — respectfully, by design):** the issue's "weight=0 → plain random LDA" is the R-compatible **frequency** default (which does random-init at weight=0). The **uniform** scheme deliberately keeps anchored init at weight=0 — topica's *original* behavior, asserted by `test_seededlda_weight_zero_reduces_to_symmetric_word_prior_at_initialization`. Now stated explicitly in the docstring so the contract is unambiguous.

Codex confirmed the rest: the frequency masses reproduce `seededlda::tfm` exactly (verified to 1.4e-14), the per-word threading and dedup are consistent across all three backends, EmbeddingLDA is protected (pins uniform), save/load handles the new field, and the `tfm()` gold is exact. It noted the 0.15 cosine margin is generous — fair; the exact `tfm` matrix check is the strong fidelity evidence, the cosine is the secondary "lands near R" check. Preflight clean.

Issue #481, opened 2026-07-24
Harden remaining hyperparameter guards that admit NaN/+inf (a <= 0.0) across PyO3 constructors


nealcaren:
## Summary

The `a <= 0.0` hyperparameter guard admits `NaN` and `+inf` (both compare `false`), so a non-finite constructor argument slips past validation and contaminates `topic_word` / `doc_topic` / downstream output with `NaN`. This bug class was closed at several sites in #472 (SeededLDA `weight`), #473 (DiscLDA `alpha`/`beta`), and #479 (SeededLDA `doc_topic_prior`) — each switching to a finite-and-positive check (`!finite_pos(a)` / `!a.is_finite() || a <= 0.0`). Surfaced again in the #479 review (independent Antigravity/Gemini pass), the codebase is **mid-migration**: the same raw `<= 0.0` pattern still lives in several PyO3 constructors.

## Still-exposed sites (verified on `main`)

- `src/python/btm.rs:136` (`alpha`), `:140` (`beta`)
- `src/python/rtm.rs:137` (`alpha`) — and the `< 0.0` guards at `:142/:146/:149` (`r` / `negative_ratio` / `ridge`) likewise admit `NaN`
- `src/python/tlda.rs:168` (`alpha_0`), `:177` (`learning_rate`), `:186` (`theta`)
- `src/python/pltm.rs:190` (`alpha`), `:194` (`beta`)
- `src/python/mod.rs` — `FASTopic` `theta_temp` (`theta_temp <= 0.0`)

Failure scenario (any of the above): e.g. `topica.BTM(k, alpha=float("nan"))` or `topica.TLDA(k, learning_rate=float("inf"))` constructs successfully, then produces non-finite topics instead of raising a clean `ValueError`.

## Already hardened (the target pattern)

`disclda.rs`, `neural.rs`, `party_embeddings.rs`, and `mod.rs:12488/12518/12523` already use `!a.is_finite() || a <= 0.0` (or `!finite_pos(a)`). `finite_pos` (`x.is_finite() && x > 0.0`) lives in `src/python/mod.rs`.

## Proposed work

1. Convert each `<= 0.0` / `< 0.0` positivity guard in the sites above to a finite-aware check (`!finite_pos(a)` for strictly-positive params; a finite-and-`>= 0` variant where zero is admissible, e.g. RTM `ridge`/`negative_ratio`).
[...]

PR #482, opened 2026-07-24
feat(disclda): calibrated class-prior semantics for the direct classifier (#460)

### LINKED ISSUE #460: DiscLDA: reject non-finite hyperparameters and calibrate classifier semantics
## Summary

Topica DiscLDA correctly implements the documented fixed block-transform variant, but it accepts non-finite alpha/beta values and presents a uniform-prior plug-in classifier score as an unqualified class posterior. Iteration validation is also inconsistent between the constructor and fit override.

This issue does not question the fixed-transform restricted Gibbs mechanism.

## Reproduced invalid-parameter behavior

Constructor checks use `<= 0`, so non-finite values pass validation:

- `alpha=NaN` or `alpha=Infinity`: fitting succeeds; `doc_topic` and `predict_proba` contain NaN.
- `beta=NaN` or `beta=Infinity`: fitting succeeds; `topic_word` and `predict_proba` contain NaN.

Reject non-finite alpha and beta at construction.

## Classifier semantics

`predict_doc` computes one restricted Gibbs inference per class, scores the document with a plug-in posterior-mean theta likelihood, and softmaxes those scores under a hard-coded uniform class prior. The fitted model does not retain class counts.

Therefore `predict_proba` is not a calibrated posterior p(class | words) for an imbalanced corpus. On a 90:10 imbalanced corpus with language-identical documents, the returned probabilities are approximately 0.5 / 0.5 rather than reflecting observed class prevalence.

The direct classifier is a useful topica-native extension, but the API and docs should describe it as a uniform-prior, plug-in approximate score/probability. Consider retaining empirical class priors and making `class_prior` configurable.

## Iteration validation

- Constructor rejects `iters=0`.
- `fit(..., iters=0)` accepts it and returns a random-initialization model.
- `infer_sweeps=0` is accepted and makes inference depend on the random initialization without any Gibbs update.

Choose and document one contract: reject zero steps, or expose them only as deliberate diagnostic modes.

## Reference scope

Topica implements the paper section 4.1 fixed-transform variant: class-specific topic blocks plus shared topics. It does not implement the paper section 4.2 learned-transform variant. There is no canonical maintained reference implementation; the existing 20 Newsgroups check is paper-derived, download-dependent, and validates only the direction of feature-classification performance against LDA.

## Proposed scope

- [ ] Validate alpha and beta as finite and strictly positive, with NaN/infinity regression tests.
- [ ] Validate or explicitly document zero `fit(iters=...)` and `infer_sweeps` behavior.
- [ ] Retain class frequencies at fit time and support empirical or user-supplied class priors for direct prediction; retain uniform as an explicit option.
- [ ] Rename/document `predict_proba` as uniform-prior plug-in approximate probabilities until calibrated priors/evidence are implemented.
- [ ] Ensure empty/OOV document behavior does not silently choose a class merely by sorted-order tie break.
- [ ] Keep documentation explicit that this is fixed-transform DiscLDA, not the learned-transform method.
- [ ] Add a small, committed imbalanced-class fixture that tests the chosen prior contract; retain 20NG only as paper-derived directional validation.

## Acceptance criteria

Invalid hyperparameters fail before fitting; the direct-classification output has documented, testable class-prior semantics; zero-step behavior is intentional and documented; and the fixed-transform scope remains explicit.

nealcaren:
Completes the classifier-semantics half of #460 (PR #473 did the hyperparameter/zero-step validation; this is the follow-up it deferred).

## Problem

`predict_doc` softmaxed per-class **plug-in** likelihoods under a hard-coded **uniform** prior. So:
- `predict_proba` was not calibrated to class prevalence — on a 90:10 corpus with language-identical documents it returned ~0.5/0.5 rather than the base rate.
- An empty / all-OOV document (zero likelihood under every class) resolved to a **sorted-order tie break** (class 0), not anything meaningful.

## Fix

A `class_prior` policy, resolved at fit from observed class frequencies and applied as `ln p(c) + plug-in loglik(doc|c)` before the softmax:

- **`"empirical"` (new default)** — uses the class prevalence seen at fit, so `predict_proba` tracks the base rate and an empty/all-OOV document returns the **prior** (majority class), not a sorted-order tie.
- **`"uniform"`** — the previous equal-prior behavior, now explicit.
- **a sequence of per-class weights** (sorted-`classes` order) — custom, normalized.

New `class_prior` (resolved p(c), sums to 1) and `class_counts` getters report what was used; both persist through save/load (older saves default to uniform). `predict`/`predict_proba` docstrings + the docs now state honestly this is a topica-native **plug-in, prior-calibrated score**, not fully marginalized evidence, and keep the fixed-transform §4.1 scope explicit.

Verified on the language-identical 90:10 case:
```
empirical  class_prior=[0.9,0.1]  OOV_proba=[0.9,0.1]  predict([[]])=maj
uniform    class_prior=[0.5,0.5]  OOV_proba=[0.5,0.5]
[3,1]      class_prior=[0.75,0.25]
```

## Tests
- Committed 90:10 language-identical fixture: empirical tracks prevalence, uniform stays ~0.5, custom is normalized, empty-doc → prior, getters, save/load persistence, bad-prior rejection (string at construction, wrong-length custom at fit, non-positive weights).
- Rust unit test: an empty document's posterior equals the (softmaxed) class prior.
- `cargo test --lib disclda` (4) + `tests/test_disclda.py` (25) + `test_save_load_fixes` green; preflight + `mkdocs --strict` clean.

## Scope
This closes the #460 acceptance criteria (validation was #473; classifier semantics here). The §4.2 **learned-transform** variant remains a separate future item (already noted in the docstring/docs), not part of #460.

Closes #460.

### DIFF
--- a/docs/guides/models.md
+++ b/docs/guides/models.md
@@ -1105,6 +1105,17 @@ restricts tokens to a document's own labels, DiscLDA is the one that builds the
 class-specific-vs-shared split and a discriminative representation. Determinism is
 `seed-reproducible`.
 
+`predict` / `predict_proba` are a topica-native **plug-in** classifier, not part of
+the paper: each class score is the document's posterior-mean-θ likelihood combined
+with a class prior and softmaxed, so treat `predict_proba` as a prior-calibrated
+score rather than exact evidence. The prior is set by `class_prior`: `"empirical"`
+(default) uses the observed class frequencies from fit — so on an imbalanced corpus
+the probabilities track prevalence, and an empty/all-out-of-vocabulary document
+returns the prior (the majority class) rather than a sorted-order tie. Pass
+`class_prior="uniform"` for an equal prior, or a sequence of per-class weights (in
+`classes` order) for a custom one; `m.class_prior` and `m.class_counts` report what
+was used.
+
 DiscLDA has no canonical reference implementation, so it is validated against the
 paper's 20 Newsgroups result (`parity/disclda_20ng.py`): DiscLDA's topic-proportion
 features feed a linear classifier better than unsupervised-LDA features of matched
--- a/python/topica/_topica.pyi
+++ b/python/topica/_topica.pyi
@@ -2563,8 +2563,16 @@ class DiscLDA:
         beta: float = 0.01,
         iters: int = 1000,
         infer_sweeps: int = 100,
+        class_prior: str | Sequence[float] | None = None,
         seed: int = 42,
-    ) -> None: ...
+    ) -> None:
+        """class_prior sets the prior the direct classifier combines with each
+        document's plug-in likelihood: "empirical" (default) uses the observed
+        class frequencies from fit, so predict_proba is calibrated to class
+        prevalence; "uniform" gives every class an equal prior; or pass a sequence
+        of positive per-class weights (in the sorted-class order of `classes`),
+        which is normalised."""
+        ...
     def fit(
         self,
         data: Corpus | Sequence[Sequence[str]],
@@ -2580,15 +2588,29 @@ class DiscLDA:
         """Class-marginalized discriminative representation (num_docs, num_topics)."""
         ...
     def predict(self, data: Corpus | Sequence[Sequence[str]]) -> list[str]:
-        """MAP class label per document."""
+        """Predicted class label per document (argmax of predict_proba). An empty /
+        all-OOV document resolves to the most probable class under class_prior (the
+        majority class for the default empirical prior)."""
         ...
     def predict_proba(
         self, data: Corpus | Sequence[Sequence[str]]
     ) -> numpy.typing.NDArray[numpy.float64]:
-        """Class posteriors (num_docs, num_classes), columns in `classes` order."""
+        """Approximate class posteriors (num_docs, num_classes), columns in
+        `classes` order. A topica-native plug-in classifier: each class score is
+        the posterior-mean-theta likelihood combined with class_prior (empirical by
+        default) and softmaxed — a prior-calibrated score, not exact evidence."""
         ...
     @property
     def classes(self) -> list[str]: ...
+    @property
+    def class_prior(self) -> list[float]:
+        """The resolved class prior p(c) used by predict/predict_proba, in `classes`
+        order (sums to 1)."""
+        ...
+    @property
+    def class_counts(self) -> list[int]:
+        """Observed per-class document counts from fit, in `classes` order."""
+        ...
     @property
     def num_topics(self) -> int: ...
     @property
--- a/src/disclda.rs
+++ b/src/disclda.rs
@@ -40,6 +40,14 @@ pub struct DiscLdaModel {
     /// Document-topic matrix θ (D × L); rows sum to 1, mass only on the document's
     /// class block and the shared block.
     pub doc_topic: Vec<Vec<f64>>,
+    /// Log class prior `ln p(c)` used by `predict`/`predict_proba`, length
+    /// `num_classes`. The direct classifier is a topica-native plug-in: it scores
+    /// a document under each class and combines the scores with this prior. The
+    /// binding resolves it from the constructor's `class_prior` (empirical class
+    /// frequencies by default, uniform, or a user-supplied vector); `fit_disclda`
+    /// initialises it uniform. A shift-invariant softmax means only the prior's
+    /// *relative* magnitudes matter.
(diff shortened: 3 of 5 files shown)


nealcaren:
Review pass (Claude + Codex `gpt-5.6-terra`, medium reasoning; a Gemini/`agy` third pass was attempted but hit a quota 429). The calibrated-prior design is clean and well-tested, but Codex caught one **blocker** worth closing before merge:

**Custom `class_prior` with huge finite weights overflows to a NaN posterior.** `parse_class_prior` validates each weight is finite and `> 0` (`src/python/disclda.rs:229`) but not their sum. So `class_prior=[1e308, 1e308]` is accepted, then `resolve_log_prior` does:

```rust
let s: f64 = w.iter().sum();          // 1e308 + 1e308 -> +inf
Ok(w.iter().map(|&x| (x / s).ln()).collect())   // (1e308/inf).ln() = ln(0) = -inf
```

Every log-prior becomes `-inf`, so in `predict_doc` each class score is `loglik + (-inf) = -inf`, the max-shift gives `exp(-inf - -inf) = exp(NaN) = NaN`, and `predict_proba` returns NaN rows. Same non-finite-contamination class as #481/#456.

Fix: reject a non-finite normalizer (or normalize stably, e.g. subtract `max(ln w)` in log-space / divide by `max(w)` before summing). Add a regression asserting `predict_proba` stays finite for an extreme custom prior.

Everything else verified clean by both passes:
- Scores use `loglik + log_prior` with a stable max-shifted softmax (`src/disclda.rs:308-318`); empty/all-OOV doc returns exactly the prior (Rust + Python tests).
- Empirical prior = sorted-label training counts; uniform/custom modes correct and normalised; per-weight validation rejects NaN/inf/≤0 and empty.
- `class_counts` / `class_prior` / policy / resolved log-prior all serialized + restored; old saves fall back to uniform (defensible; worth a one-line note if legacy-load is a supported promise).
- Relabeling is honest — `predict_proba` is now documented as an approximate prior-calibrated plug-in score, not exact `p(class|words)`.

Minor (non-blocking): the wrong-length custom-prior error at `src/python/disclda.rs:260` interpolates `counts.len()` (a number) where the message promises the sorted-class *order* — print `classes` instead.

nealcaren:
Fixed in `819f8ba`: the custom-prior normalizer now divides by the max weight before summing, so an extreme-but-finite `class_prior` (e.g. `[1e308, 1e308]`) can no longer overflow the sum to `+inf` and collapse the log-priors to `-inf`/NaN — the max cancels, so the normalization is exact but the denominator stays in `(0, num_classes]`. Added a parametrized regression (`[1e308,1e308]`, `[1e308,1.0]`, `[1e-300,1e300]`) asserting the prior and `predict_proba` stay finite and sum to 1, and dropped the stray numeric interpolation in the wrong-length error message. Local: full `test_disclda.py` 28 passed; `cargo fmt`/`clippy` clean; `cargo test --lib disclda` 4 passed. This resolves the review blocker.

nealcaren:
**Independent review (Codex, cross-family):** verdict *needs-changes* on one point, resolved:

- **Old-save back-compat (the serde(default) bincode trap):** Codex correctly noted the `DiscLdaState` comment implied pre-branch saves would load with a uniform prior — they won't (bincode is tagless, hits EOF on the new trailing fields; `serde(default)` doesn't rescue it). This is the exact repo-wide trap documented at `mod.rs` / tracked by #443. Per the project save-compat stance (only Neal has saved models; not a pre-v1.0 blocker; keep the claim honest), I corrected the comment to state old DiscLDA saves don't load until the versioned format lands, rather than adding migration.
- **transform() uses class_prior (downstream-behavior note):** documented that `transform`'s class-marginalized features run through the same prior-calibrated posterior as `predict_proba`, so `class_prior` shapes the features too.

Codex confirmed the rest is correct: the prior math (empirical count/total, uniform ln(1/C), custom normalize-by-max avoids overflow, log-prior added before max-subtraction softmax, empty/OOV → prior), custom-weight validation (parse-time finite/>0/non-empty, fit-time length), new-save round-trip + settings(), and that the empirical default doesn't break the balanced existing fixtures. Two lower-confidence items left as-is: (1) a tiny-weight custom class can store `-inf` log-prior — benign (that class correctly gets ~0 posterior; the max-weight class is always finite, so no NaN); (2) `load()` trusts persisted values — only reachable via a tampered save (same deferred class as the deserialization-bypass note). Doc-only change; suite + preflight green.

PR #483, opened 2026-07-24
fix(bindings): harden NaN/+inf-admitting hyperparameter guards (#481)

### LINKED ISSUE #481: Harden remaining hyperparameter guards that admit NaN/+inf (a <= 0.0) across PyO3 constructors

nealcaren:
Finishes the migration off the `a <= 0.0` positivity guard, which admits `NaN` and `+inf` (both compare `false`) so a non-finite constructor argument slips past validation and produces `NaN` `topic_word`/`doc_topic`. #472/#473/#479 fixed this class at individual sites; this closes the enumerated remainder.

## Changes

- **Shared helpers** in `src/python/mod.rs` (next to `finite_pos`): `ensure_finite_pos(name, x)` (finite & > 0) and `ensure_finite_nonneg(name, x)` (finite & ≥ 0), each returning a clean `ValueError` naming the parameter.
- **Converted sites:** BTM `alpha`/`beta`, PolylingualLDA `alpha`/`beta`, TensorLDA `alpha_0`/`learning_rate`/`theta`, FASTopic `theta_temp` → `ensure_finite_pos`. RTM `alpha` → `ensure_finite_pos`; `rho`/`negative_ratio`/`ridge` → `ensure_finite_nonneg` (zero is genuinely valid — no regularization / no pseudo-negative link evidence; verified against `src/rtm.rs`, they only enter `ln()`/products, never a divisor).
- **Preflight guard** (`scripts/preflight.sh`, perl so it runs on the macOS pre-push hook): flags the un-hardened shape `if IDENT <=/< 0.0 {` immediately followed by a `return Err(PyValueError` line, so a new binding can't silently reintroduce it. Verified it fires on a synthetic regression and is quiet on the tree.
- **Tests:** parametrized NaN + inf rejection for BTM/PLTM/RTM/FASTopic and a TensorLDA case (under the experimental flag) in `tests/test_input_validation.py`; two exact-string TLDA assertions updated to the new `"… must be finite and > 0"` message.

## Scope note
The SeededLDA `doc_topic_prior` guard (`mod.rs`, `a <= 0.0`) is **intentionally left to the open PR #479**, which owns that line — it's the only remaining bare NaN-admitting hyperparameter guard, deferred to avoid a conflict.

## Review
An independent review ran (Gemini was quota-blocked at both medium and low tiers, so a Claude subagent stood in per the rate-limit fallback). Verdict **ship**; it confirmed no other missed sites, correct zero-admissibility for the RTM params, no false positives in the preflight guard, and no behavior breaks. Incorporated its two nits: per-file line numbers in the preflight scan (`close ARGV if eof` — perl `$.` was cumulative across the glob), and `ensure_finite_pos` now delegates to `finite_pos` instead of re-inlining the predicate.

All affected suites green (96 passed); `cargo test --lib` unaffected; preflight (incl. the new guard) + stub sync clean.

Closes #481.

### DIFF
--- a/python/topica/_topica.pyi
+++ b/python/topica/_topica.pyi
@@ -2688,7 +2688,14 @@ class RTM:
         negative_ratio: float = 1.0,
         ridge: float = 1.0,
         seed: int = 42,
-    ) -> None: ...
+    ) -> None:
+        """`rho` (or `negative_ratio`, which scales it by the link count) is the
+        pseudo-negative regularization that prevents the degenerate positive-links-
+        only fit — the paper's rho, R lda's `lambda` — and must be strictly
+        positive; `rho=None` resolves to `negative_ratio * num_links`. `ridge` is
+        the separate L2 Gaussian prior on the link coefficients and may be 0 (plain
+        MLE)."""
+        ...
     def fit(
         self,
         data: Corpus | Sequence[Sequence[str]],
--- a/scripts/preflight.sh
+++ b/scripts/preflight.sh
@@ -34,6 +34,27 @@ if ! cargo clippy --workspace --all-targets --all-features -- -D warnings; then
     fail=1
 fi
 
+# #481: a bare `if X <= 0.0 { return Err(PyValueError...) }` positivity guard
+# admits NaN/+inf (both compare false) and silently corrupts the fit. New
+# hyperparameter checks must route through ensure_finite_pos / ensure_finite_nonneg
+# (or an explicit `!x.is_finite()` clause). Flag the raw anti-pattern before it
+# lands. (perl, not `grep -P`, so this runs on the macOS pre-push hook.)
+step "no NaN-admitting '<= 0.0' hyperparameter guards in the bindings (#481)"
+antipattern=$(perl -ne '
+    if ($prev =~ /^\s*if\s+[A-Za-z_][\w.]*\s*<=?\s*0\.0\s*\{\s*$/ && /return\s+Err\(PyValueError/) {
+        print "  $ARGV:", ($. - 1), ": ", $prev;
+    }
+    $prev = $_;
+    # Reset per file so $. is the file-local line number and $prev cannot leak
+    # across the *.rs glob boundary.
+    if (eof) { close ARGV; $prev = ""; }
+' src/python/*.rs || true)
+if [ -n "$antipattern" ]; then
+    echo "  a hyperparameter guard admits NaN/+inf; use ensure_finite_pos/_nonneg:" >&2
+    printf '%s\n' "$antipattern" >&2
+    fail=1
+fi
+
 # Python-side checks need the built extension; prefer an active venv, else .venv-dev.
 VENV="${VIRTUAL_ENV:-$PWD/.venv-dev}"
 PY="$VENV/bin/python"
--- a/src/python/btm.rs
+++ b/src/python/btm.rs
@@ -133,13 +133,9 @@ impl BTM {
             return Err(PyValueError::new_err("num_topics must be >= 2"));
         }
         if let Some(a) = alpha {
-            if a <= 0.0 {
-                return Err(PyValueError::new_err("alpha must be > 0.0"));
-            }
+            ensure_finite_pos("alpha", a)?;
         }
-        if beta <= 0.0 {
-            return Err(PyValueError::new_err("beta must be > 0.0"));
-        }
+        ensure_finite_pos("beta", beta)?;
         if iters == 0 {
             return Err(PyValueError::new_err("iters must be > 0"));
         }
--- a/src/python/mod.rs
+++ b/src/python/mod.rs
@@ -569,6 +569,35 @@ fn finite_pos(x: f64) -> bool {
     x.is_finite() && x > 0.0
 }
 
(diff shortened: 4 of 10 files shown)


nealcaren:
**Independent review (Codex, cross-family):** the Gemini/`agy` CLI was quota-exhausted, so this got a Codex `exec` review instead. Verdict was *needs changes* on one point — worth incorporating, and grounded in the reference:

- **RTM `rho`/`negative_ratio` should be strictly positive, not zero-admissible.** They are the paper's pseudo-negative regularization (R `lda`'s `rtm.em` `lambda` — "the number of non-links to simulate," defaulted to a positive data-derived value). Zero removes the negative evidence and the logistic-link intercept `nu` diverges (no finite optimum) — the exact degeneracy the parameter exists to prevent. Changed both to `ensure_finite_pos`; `ridge` (the separate L2 Gaussian prior) stays `ensure_finite_nonneg` since `ridge=0` = plain MLE is legitimate. Lock-in tests + docstring added.

Note this tightens `negative_ratio` beyond the `>= 0` example in the issue text — the paper + R lda evidence supports strictly-positive. Everything else in the PR both reviewers cleared: no missed sites, guard macOS-safe with no false positives, no test breaks, `pub(crate)` correct.

nealcaren:
Review pass (Claude + Gemini `gemini-3.6-flash-medium`; a Codex medium pass was attempted but ghosted — 0 output, killed). Both reviewers verified the **code** is complete and correct: every #481 site routes through the new `ensure_finite_pos`/`ensure_finite_nonneg` helpers with the right semantics (`rho`/`negative_ratio` strictly positive — zero removes the pseudo-negatives and the logistic intercept diverges — `ridge` non-negative), no site missed (I ran the new preflight grep guard against the branch: **zero** remaining anti-patterns; all surviving `<= 0.0` are already the finite-aware form), and every default still validates.

Gemini's one nit — a test-symmetry gap — is now closed in `e916e59`: the TensorLDA check only exercised one non-finite value per param, while the other constructors got both NaN and +inf via `_NONFINITE_POS_CASES`. `test_tlda_nonfinite_hyperparameters_rejected` is now parametrized over `alpha_0`/`learning_rate`/`theta` × {NaN, +inf}. Local: `test_input_validation.py` + `test_rtm.py` + `test_tlda.py` 86 passed; preflight (incl. the new guard) green. Closes #481.

PR #484, opened 2026-07-24
feat(rtm): collapsed-Gibbs backend (inference="gibbs") for R lda same-algorithm parity (#424)

### LINKED ISSUE #424: RTM: add a collapsed-Gibbs inference backend (inference="gibbs") for same-algorithm R lda parity
Follow-up to #414 / PR #423 (the variational RTM). Add a second inference backend to `topica.RTM`: **collapsed Gibbs**, alongside the shipped variational EM.

```python
topica.RTM(20, inference="variational")   # default — the shipped §3 variational EM
topica.RTM(20, inference="gibbs")          # new — collapsed Gibbs, matches R lda's rtm sampler
```

### Why this is worth doing
The headline reason is **validation**, not just user choice. The only authoritative RTM implementation is the R `lda` package (Chang's own code), and its `rtm.em` is a **collapsed Gibbs sampler** (its body calls `rtm.collapsed.gibbs.sampler`) — not the paper's variational EM. So today R can only be a *directional* baseline for topica's variational fit. A collapsed-Gibbs backend lets us parity-check **the same algorithm** against the author's own implementation, turning R `lda` from a directional baseline into an authoritative oracle. That closes the one gap in the current validation story (the variational path is validated only against our own NumPy reimplementation of the paper's equations).

Secondary: Gibbs is often more robust on small corpora and is the more familiar engine for many social scientists; variational is faster and hands back the link coefficients / ELBO directly.

### Scope
- A collapsed-Gibbs RTM sampler: sample `z_{d,n}` from the usual collapsed-LDA conditional times a link-likelihood factor over `d`'s observed neighbors (the change in link likelihood from flipping the token's topic, which moves `z̄_d`). Reuse topica's collapsed-Gibbs machinery (`sampler.rs`, `model.rs`, the count bookkeeping).
- Estimate the link coefficients periodically, as R's sampler does (it fits a link coefficient it calls `beta`; note R's routine is known to fit a *negative* coefficient even on strongly-linked data — see the PR #423 benchmark — so match the algorithm, not that quirk).
- Same fitted surface as the variational backend: `topic_word`, `doc_topic`, `predict_link`, `suggest_links`, `eta`/`nu` (or the Gibbs analogue). Share the link head where possible.
- Exponential link first (that is what R's sampler supports); logistic if it falls out naturally.

### Validation
- **Same-algorithm parity vs R `lda`'s `rtm.collapsed.gibbs.sampler`** on a fixed integer-coded corpus (Cora ships as `data(cora)` in the package): topic-aligned agreement to the reference's own seed-to-seed self-consistency floor, plus the link-prediction ordering. This is the real prize — the first RTM parity check against the author's own code on its own terms.
- Recovery + determinism unit tests mirroring the variational backend.
- The existing NumPy reference (`parity/rtm_reference.py`) stays the oracle for the variational path.

### Notes
- Determinism: seed + (for AD-LDA-style threading, if any) thread count, per topica's contract.
- The variational backend (PR #423) is the default and unchanged; this is purely additive (`inference=` keyword).
- Design context: the RTM plan at the time of PR #423 explicitly deferred this to keep PR1 focused; this is that follow-up.

nealcaren:
Adds a second RTM inference backend alongside the shipped variational EM: **collapsed Gibbs**, matching R `lda`'s `rtm.em` (which wraps `rtm.collapsed.gibbs.sampler`).

```python
topica.RTM(20, inference="variational")   # default — the shipped §3 variational EM
topica.RTM(20, inference="gibbs")          # new — collapsed Gibbs, R lda's algorithm
```

## Why (the validation prize)

R `lda`'s `rtm.em` **is** collapsed Gibbs, not the paper's variational EM — so it could only be a *directional* baseline for topica's variational fit (which is validated against a NumPy reimplementation of the paper's equations). This backend is the **same algorithm**, so R `lda` becomes an **authoritative same-algorithm oracle**. That closes the one gap in RTM's validation story.

## Core (`src/rtm.rs::fit_rtm_gibbs`)

Faithful port of R's algorithm:
- **E-step conditional:** LDA collapsed term × link factor `exp(β_k · Σ_{d'~d} n_{d',k}/(len_d·len_{d'}))`, with `β` fixed during the sweep and the link factor computed **once per document per sweep** (neighbour counts don't change while a doc's own tokens resample) — exactly as R's C sampler does.
- **M-step (`rtm_gibbs_estimate_beta`):** a port of R's `estimate.params` — `z̄_d=(n_d+α)/Σ`, `p_k=Σ_{links} z̄_{d,k}z̄_{d',k}`, regularized by `λ·α²·#pairs` pseudo-negatives (R's `lambda` default = link density), `β_k=ln(p_k)`.
- That log-ratio makes `β` **negative even on strongly-linked data** — the documented reference quirk, reproduced deliberately. Consequence: the Gibbs link scores rank *dissimilar* docs higher, so use `inference="variational"` for `predict_link`/`suggest_links`; the Gibbs backend is for same-algorithm **topic** parity.

## Binding & surface

`inference="variational"`(default)|`"gibbs"` + a topic-word `beta` smoothing (R lda's `eta`). Same fitted surface (`topic_word`/`doc_topic`/`phi_bar`/`eta`/`nu`, `predict_link`, `suggest_links`). Settings, save/load, `.pyi` updated. Purely additive — the variational default is unchanged.

## Validation

`parity/rtm_gibbs_gold.py` freezes R `lda` `rtm.em`'s two-seed topic-word phi on a fixed planted network; the offline test refits topica gibbs and clears R's own seed-to-seed self-consistency floor:

- **topic-word cosine 0.9947 vs R self 0.9936** (bar 0.9436) — topica agrees with R better than R agrees with itself.
- link coefficients near-identical: topica `[-0.032,-0.036,-0.035]` vs R `[-0.044,-0.031,-0.037]`.

Runs in CI **without Rscript** (reference frozen in `.npz`/`.json`). `parity/rtm_gibbs_r_compare.py` is the live reporter. Recovery / determinism / conformance / negative-β unit tests (Rust + Python).

## Tests
`cargo test --lib` 208 passed (+4 RTM Gibbs); RTM Python + gold + settings + save-load + stub + naming suites green (356 passed); preflight + `mkdocs --strict` clean.

Closes #424.


nealcaren:
**Independent review (Codex, cross-family):** verdict *needs changes* — two real defects fixed, plus a parity-strength improvement:

- **Link mismatch (confirmed):** `inference="gibbs"` trained with R's exponential conditional but stored the default `link="logistic"`, so `predict_link`/`suggest_links` applied σ to the negative R coefficients. R lda's Gibbs sampler is exponential-only → the backend now forces and stores `Link::Exponential` (in both the binding and `fit_rtm_gibbs`), with a test.
- **`phi_bar` contract violation (confirmed):** it was the α-smoothed `doc_topic`, but the field is documented as the empirical z̄ = n/N → now computes the raw empirical proportions (empty-doc row = 0).
- **Parity strength (fair critique):** topic-word cosine alone was weak evidence for the *relational* part on a planted corpus. The gold now also freezes R's link coefficient β and asserts topica's mean β matches (gap **0.0035**, both negative), covering the M-step. Added no-links / empty-doc Gibbs tests.

Codex confirmed the E-step conditional and M-step `estimate.params` port are otherwise faithful to R's spec (normalizations, once-per-doc link factor, unsmoothed E / smoothed M, λ·α²·#pairs reg, directed-endpoint sum, β init 3.0, loop nesting), bookkeeping/determinism correct, and no variational-path break. Lower-confidence nits (`.max(1)` loop floors, exp-space sampling) left as-is — they match R's numeric envelope. All green; preflight clean.

nealcaren:
Review pass (Claude + Codex `gpt-5.6-terra`, medium; a Claude subagent build/test pass is also running). The Gibbs sampler algebra and the M-step β estimation are correct and faithful — the per-token conditional is the LDA collapsed conditional × the exponential link factor `exp(β_k·Σ_{d'~d} z̄_{d',k}/len_d)` with the right sign/normalization, `β_k = ln(p_k/(p_k+λα²·pairs))` reproduces R's `estimate.params` (negative-β quirk and all), and determinism / save-load / no-links / empty-doc edges are handled. But the pass surfaced several items that should land before merge, so **request changes**:

**1. Rebase blocker + reintroduced NaN-guard anti-pattern (mine).** #484 branched before #483 (now merged) and **conflicts with main** in `src/python/rtm.rs`. Worse, it adds `if beta <= 0.0 { return Err(PyValueError...) }` and still carries the old `alpha`/`rho`/`negative_ratio`/`ridge` `<= 0.0`/`< 0.0` guards — exactly the NaN/+inf-admitting anti-pattern #481/#483 just eliminated, and #483's new `scripts/preflight.sh` grep guard will **reject** them (pre-push + CI lint). On rebase: take main's `ensure_finite_pos`/`ensure_finite_nonneg` conversions for alpha/rho/negative_ratio/ridge (note #483 made `rho`/`negative_ratio` *strictly* positive), and add `ensure_finite_pos("beta", beta)` for the new param.

**2. [P0] "same algorithm" claim vs actual mechanics (Codex, verified against R `lda` source).** R's `rtm.em` **restarts** a fresh `rtm.collapsed.gibbs.sampler` every M-step (fresh uniform init, no carried chain), and resamples per `(word, count)` cell; `fit_rtm_gibbs` initializes once and carries the count tables across all M-steps, resampling per token. The committed gold does show topica's topics clear R's own two-seed floor (0.9947 vs 0.9936), so **result-parity within R's seed noise is real** — but that is weaker than the "authoritative same-algorithm oracle" the docs/comments repeatedly claim. Either reproduce R's restart/count-cell mechanics, or soften the claim to "reproduces R `rtm.em`'s topic distributions within its own seed-to-seed noise" (result-parity, not mechanical identity).

**3. [P1] The gold gate does not isolate RTM's link influence (Codex).** On this highly separable planted corpus, a plain-LDA Gibbs fit plus the β post-step could clear the 0.9436 cosine bar with **no link factor acting during sampling**; the shuffle test only rules out a word-permuted matrix. Add a negative control that must FAIL — e.g. link-ablated (β forced to 0 during sampling) or mis-signed links — so the gate proves the link term actually shapes the fit.

**4. [P1] Explicit `link="logistic"` + `inference="gibbs"` is silently overwritten to exponential (Codex).** Silent config mutation — the same class flagged on #474. Reject an incompatible explicit `link` under Gibbs (or make the signature's Gibbs default exponential so "unspecified" is distinguishable), rather than discarding the user's value.

**5. [P2] Stale runtime class docstring** at `src/python/rtm.rs:12-19` — omits `inference`/`beta`/forced-exponential (the `.pyi` is accurate).

Items 2–3 are the ones I'd want your call on — is the goal exact `rtm.em` mechanics, or result-parity within R's noise? That decides whether #2 is a code change or a docs change, and shapes #3's negative control. Happy to re-review once rebased.

nealcaren:
Adding a third opinion: a Claude subagent built #484 in an isolated worktree and ran the full suite (cargo rtm 7/7, pytest 20/20, fmt/clippy clean, gold hermetic under a fake failing Rscript). It landed **APPROVE with notes** — a split from Codex's REQUEST-CHANGES. Reconciling all three:

- **All three agree** the Gibbs conditional + β M-step faithfully reproduce R's arithmetic (the subagent numerically confirmed reg = 2|E|·α² → β ≈ −0.04 vs R's −0.0375; a 1/K² term would give ≈ −0.38, so the formula is right).
- **New, build-verified minor bug (subagent):** with **no links**, `p_k = 0` and `reg = 0` → `0.0/0.0 = NaN` → `NaN.max(1e-300).ln() ≈ −690.78` (`src/rtm.rs` in `rtm_gibbs_estimate_beta`). Output stays finite and topics are valid (β is unused when adjacency is empty), but `eta`/`predict_link` report a garbage sentinel instead of an honest `0.0`. One-line guard.
- **Codex and the subagent agree** the 0.9947 cosine mostly exercises the LDA core (R's β ≈ −0.037 makes the link factor near-inert), so the relational term is only weakly covered by cosine — Codex wants a link-ablation/mis-signed negative control; the subagent judges the β-mean-gap gate (0.0035 < 0.02) sufficient. A cheap negative control would settle it.

Net across the panel: the core is faithful, but there's the mandatory rebase (+ the #483 preflight-guard conversions from my earlier comment), the no-links β sentinel, the same-algorithm-claim precision (P0), and the silent `link` override — so I'm holding the merge for your call on P0 + a rebase. Not a rework; a focused revision.

nealcaren:
Reworked to **match R lda's mechanics exactly** (per the decision to port `rtm.em` faithfully rather than soften the claim), rebased onto main, and addressed the review. Summary:

**Exact mechanics (ported from R `src/gibbs.c` `rtm()`, verified against local R lda 1.5.2):**
- Fresh sampler **restart each M-step** (only β carries over); sweep 0 is a uniform-random init, sweeps 1..N-1 the real updates.
- **Per-`(word,count)`-cell block sampling** (remove the whole count, draw one topic, add it back) — no longer per-token.
- The link factor and `estimate.params` β update were already faithful; kept.

**What matching R exactly revealed (and how the gold adapts):**
- R's Mersenne-Twister and topica's ChaCha mix at different rates: R is converged by ~8 E-sweeps on this corpus, topica needs more. So the gold now compares **at convergence** (E_SWEEPS=100), where the two stationary distributions agree to **0.9990 vs R self 0.9994** (β mean −0.0328 vs R −0.0339, gap 0.001). At R's marginal 8-sweep default the mechanics match but the RNG mixing-rate gap dominates (topica ~0.81) — that measured the RNG, not the algorithm.
- R's β = log(p_k/(p_k+reg)) **runs away negative** over many M-steps (R's *own* seed-to-seed self-cosine collapses to ~0.60 by 10 M-steps), so the gold uses M_ITERS=5. Gold regenerated against R lda 1.5.2.

**Review items:**
- **P1 (silent link)** — `link` is now a sentinel: an explicit non-exponential link under `inference="gibbs"` is rejected, not overwritten; unset resolves to exponential (gibbs) / logistic (variational).
- **Rebase/#483** — converted the reintroduced `beta <= 0.0`/`<= 0.0` guards to `ensure_finite_pos`/`ensure_finite_nonneg`; the new preflight guard is clean.
- **No-links β sentinel** — `estimate_beta` now reports an honest β = 0, not −690.
- **Relational discriminator (Codex P1)** — added a hand-computed `estimate.params` unit test pinning the β math to R's formula (the RTM-specific computation, which a plain-LDA fit can't reproduce), plus the tight β-mean gold gate.
- **P2** — refreshed the runtime class docstring + docs/guides/models.md.

Local: `cargo test --lib` 226 passed; `pytest test_rtm.py test_rtm_gibbs_gold.py` 20 passed; fmt/clippy/preflight-guard clean. CI running.

PR #485, opened 2026-07-24
fix(ctm-svi): derive Σ from the second-moment statistic and honor convergence_tol (#421)

nealcaren:
Closes the two remaining SVI defects flagged in the STM/CTM core review (#421); the β count-statistic defect was fixed earlier in #477.

## Σ centering — the sufficient-statistic fix
The SVI minibatch M-step centered its covariance cross-product on the *just-updated* global μ. The batch M-step (ctm.rs `fit_ctm`) can center on the updated μ only because there μ is the exact mean of the same λ set — the full-data MLE. The SVI μ is a Robbins-Monro blend `(1−ρ)μ + ρ·batch_mean`, which is **neither the minibatch mean nor a consistent center**, so the old code folded this batch's μ step into Σ.

The fix follows Hoffman et al. (2013) exactly: carry the raw second-moment sufficient statistic `M2 = E[ν + η ηᵀ]` as persistent global state, blend it (and μ) with ρ, and derive `Σ = M2 − μ μᵀ`. This:
- has **no centering choice** to get wrong;
- stays well defined at **batch_size = 1**, where a batch-mean-centered covariance would collapse to ν alone (the batch mean *is* the single doc);
- makes Σ moves vanish as ρ → 0.

Off-diagonal shrinkage is now applied once to the freshly derived Σ (not fed back into the unshrunk `M2`), so it is a fixed shrinkage prior rather than the step-over-step multiply that once collapsed the off-diagonals (the #446 concern, preserved).

## `converged` — honesty fix
`fit_ctm_svi` hardcoded `converged: true` with a length-1 `bound_history` and ignored `convergence_tol` entirely, so `test_svi_honors_convergence_tol` passed trivially without ever running a convergence test. Now `convergence_tol` is threaded through; the per-epoch running ELBO is accumulated (every doc is visited exactly once per epoch, so the sum is a full-corpus ELBO estimate for free); SVI early-stops on the relative epoch-to-epoch change; and `converged` / `em_iters_run` / `bound_history` are reported honestly.

## Tests
- `svi_honors_convergence_tol` (core): tol = 0 runs the full budget with `converged = false`; a loose tol early-stops with `converged = true` and a matching-length trace.
- `svi_covariance_well_defined_at_batch_one` (core): correlated topics yield a finite, PD Σ with a real off-diagonal at B = 1.
- The Python `test_svi_honors_convergence_tol` now exercises real early-stopping.
- Full suite: 1673 passed / 29 skipped; preflight (fmt + clippy + table/stub sync) clean.

### DIFF
--- a/python/topica/_topica.pyi
+++ b/python/topica/_topica.pyi
@@ -484,8 +484,11 @@ class CTM:
         Under "svi", iters is the number of epochs (passes over the data), and the
         global parameters update from minibatches of batch_size docs with a Robbins-
         Monro step rho_t = (tau + t)^(-kappa). tau (>= 0) and kappa in (0.5, 1] set
-        the learning-rate schedule; convergence_tol is ignored. SVI does not retain a
-        per-iteration bound trace.
+        the learning-rate schedule. convergence_tol early-stops on the relative
+        epoch-to-epoch change in the running training ELBO (each epoch's per-minibatch
+        bounds summed; a streaming monitoring signal, not a fixed-parameter corpus
+        bound), and fit_history reports that per-epoch trace; pass convergence_tol=0 to
+        run the full epoch budget.
 
         beta_init (K x num_words) overrides the spectral/random topic-word
         initialization with a caller-supplied base beta -- the warm-start hook for
--- a/src/python/mod.rs
+++ b/src/python/mod.rs
@@ -6458,6 +6458,7 @@ impl CTM {
                         tau,
                         kappa,
                         shrink,
+                        convergence_tol,
                         spectral,
                         keep_eta_cov,
                         diagonal,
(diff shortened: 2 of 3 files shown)


nealcaren:
Ran a Codex cross-family review. Verdict on the substance: the covariance rewrite (Σ = M2 − μμᵀ with M2 and μ each ρ-blended), the shrink-the-derived-Σ-not-the-state handling, and the binding are all confirmed correct and robust at batch_size = 1. No off-by-one in `em_iters_run` / `bound_history`.

Two valid points addressed in the follow-up commit:
- My "full-corpus ELBO estimate" comment overclaimed — the epoch sum is scored against globals that move within the epoch, so it is a streaming training score, not a fixed-parameter corpus bound. Relabeled honestly; the early-stop is documented as a heuristic (evaluating a fixed-global corpus bound each epoch would cost a full extra E-step and defeat SVI's purpose).
- The `_topica.pyi` `CTM.fit` docstring still claimed SVI ignores `convergence_tol` and keeps no trace — both now corrected.

Noted-but-left: `epochs=0` yields an empty `fit_history` (no epochs ran) while the final full E-step still computes `.bound`; that is a degenerate config and the empty trace is honest.

PR #486, opened 2026-07-24
feat(seededlda): quanteda-style dictionary seed matching (glob/regex/fixed) (#456)

nealcaren:
Closes the remaining #456 checklist item: **dictionary seed matching**. SeededLDA matched seed words to the vocabulary by exact literal string only, but the `seededlda` package uses quanteda dictionaries, which match by `valuetype` — glob (default), regex, or fixed — with an optional case fold.

## What this adds
Two constructor parameters:
- `seed_match`: `"fixed"` (default, exact literal — unchanged behavior) | `"glob"` (`*`/`?` wildcards anchored to the whole token) | `"regex"` (matched anywhere in the token, quanteda's `stri_detect_regex` semantics).
- `case_insensitive` (default `False`): folds case in the match. Pair with `"glob"` to reproduce quanteda's dictionary defaults.

A glob like `"tax*"` seeds `tax`, `taxes`, and `taxation` at once; each matched vocabulary type is seeded once (deduped within a topic).

## Faithfulness details
- `seed_prior_matrix` now routes through the **same** expansion, so it reports exactly the mass applied at fit rather than only literal matches.
- `seed_match` / `case_insensitive` are serialized (save/load, with legacy-save defaults of `"fixed"` / case-sensitive) and surfaced in `settings`.
- KeyATM keeps exact-literal keyword matching — this feature is SeededLDA-scoped.
- Invalid `seed_match` raises at construction; an invalid regex raises at fit.

## Default choice
`seed_match="fixed"` + `case_insensitive=False` preserves current behavior exactly — no silent change to existing fits. Glob/regex/case-folding are explicit opt-ins, documented as reproducing quanteda's dictionary matching. (Flag if you'd prefer glob+case-insensitive as the default to match quanteda's out-of-the-box behavior.)

## Tests / docs
- Seven exact contract tests in `test_seeded_gsdmm_contracts.py` (fixed-is-exact-default, glob expansion, case-insensitive fold, regex unanchored, invalid seed_match, invalid regex, save/load round-trip), all driven deterministically by `seed_prior_matrix` — no reference toolchain needed.
- New "dictionary seed matching" section in `docs/guides/guided.md`.
- Full local gates green: preflight (fmt + clippy + table/stub sync), `mkdocs --strict`, and the seeded + keyATM suites (354 passed).

### DIFF
--- a/docs/guides/guided.md
+++ b/docs/guides/guided.md
@@ -22,13 +22,23 @@ the package's `tfm` construction — and tokens are initialized at random
 initialization. You can read the exact per-topic, per-word pseudocounts a fit
 used from `model.seed_prior_matrix`.
 
+Seed patterns are matched to the vocabulary by `seed_match`, mirroring quanteda's
+dictionary `valuetype` (the matcher the `seededlda` package uses): `"fixed"`
+(default) is exact literal equality; `"glob"` reads `*`/`?` wildcards anchored to
+the whole token, so `"tax*"` seeds `tax`, `taxes`, and `taxation` at once; and
+`"regex"` matches a regular expression anywhere in the token. `case_insensitive`
+(default `False`) folds case — set it `True` with `seed_match="glob"` to reproduce
+quanteda's dictionary defaults. An expanding pattern seeds every matched word
+(each once); `seed_prior_matrix` reflects exactly what was applied.
+
 ```python
 import topica
 
 model = topica.SeededLDA(
-    {"economy": ["jobs", "wages", "tax"],
-     "immigration": ["border", "visa", "deport"]},
+    {"economy": ["job*", "wage*", "tax*"],
+     "immigration": ["border", "visa*", "deport*"]},
     residual=3,           # 3 extra unseeded topics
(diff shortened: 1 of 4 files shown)


nealcaren:
Ran a Codex cross-family review. Verdict: **faithful, no real seed-mass / dedup / save-load bug**. Confirmed specifics:
- `glob_to_regex_body` escapes every regex metacharacter (only `*`/`?` expand); `^(?:…)$` gives correct whole-token glob semantics; no injection path.
- Unanchored `is_match` is the right quanteda regex semantic; glob is correctly anchored.
- `seed_prior_matrix` now routes through the same `seed_word_ids` path as `fit()`, so reported mass equals applied mass; duplicates resolve to one cell.
- Within-topic dedup is sufficient across sparse/warp/cvb0 (each also dedups its own seed setup); no double-counting.
- Serde defaults (`"fixed"`/`false`) preserve legacy behavior.

One documentation point addressed in the follow-up commit: the `"regex"` dialect is Rust's linear-time `regex` crate, not R's ICU/stringi — common syntax matches identically, but backreferences and lookaround are unsupported. Noted in the docstring and stub.

PR #487, opened 2026-07-24
test(sts): committed offline gold for the reference="cran" adjusted profile (#454)

### LINKED ISSUE #454: STS: add a reference-compatible fitting profile, beginning with adjusted kappa estimation

nealcaren:
Closes #454's remaining acceptance criterion: a **portable, Rscript-free gold** that pins the reference-compatible STS profile in CI.

#454's implementation (the `reference="cran"`/`"paper"` profiles — `kappaEstimation="adjusted"`, `diag(1/20)` sentiment prior, anchor init, half-step κ damping) shipped in #474. It had a *live* end-to-end check (`parity/sts_r_package_compare.py`), but that shells out to Rscript and skips when the toolchain is absent, so it never gated CI. The committed offline gold (`sts_gold.npz`) validated only the general STS math via R-stm-on-slim, predating the adjusted profile.

## What this adds
- **`parity/sts_cran_gold.py`** — fits R `sts` (adjusted profile) at two `stmSeed`s on a fixed 300-doc poliblog subsample (stm's `poliblog5k` via `prepDocuments`) and freezes: R's topic-word distribution at mean sentiment (`softmax(mv + kappa_t)`), an R two-seed self-consistency cosine floor, and the exact tokenized corpus + sentiment/rating so topica refits identical data.
- **`parity/sts_cran_gold.{npz,json}`** — the committed fixture + provenance log.
- **`tests/test_sts_cran_gold.py`** — refits topica with `reference="cran"` on the frozen corpus and asserts its topic-word distribution clears R's floor minus a margin. Present/shape/non-vacuous(shuffle) checks run by default; the ~30s refit is `@pytest.mark.slow`. **No Rscript at test time** (module imports only topica + numpy/scipy).

## Numbers
R's adjusted-profile fit is near-identical across seeds (self cosine **0.998**), so the margin absorbs the topica-vs-R cross-implementation gap (independent inits). Margin `0.20` puts the bar at **~0.80** — the same pre-registered floor the live script uses and has validated. topica scores **0.887** (+0.089 headroom); the shuffle check confirms the gate discriminates (a permuted β falls far below the bar).

## Verification
- `tests/test_sts_cran_gold.py`: 4 passed (incl. slow refit).
- Full suite green on a fresh build off current main (the earlier local guard-test failures were a stale-binary artifact from before #483 merged; a rebuild clears them).
- `mkdocs --strict` clean; `docs/replications/sts.md` updated to describe the new gold.

Complements the glmnet-kernel check (`sts_kappa_glmnet.py`) and the live comparison (`sts_r_package_compare.py`). Once merged, #454 can be closed.


nealcaren:
Ran a Codex cross-family review. Verdict: **faithful, no changes required, no real bug**. Confirmed:
- The offline test is genuinely Rscript-free — `subprocess` lives only in `_r_version()`, reached only by `regenerate()`; `run()` just loads the gold and fits topica.
- The R driver matches the live comparator exactly: `softmax(mv + kappa_t[,k])` is the neutral-sentiment topic-word distribution, and `rep(out$vocab[d[1,]], d[2,])` correctly reconstructs stm's sparse doc format (1-based vocab indices + counts; token order is immaterial to the bag-of-words fit).
- Freeze/reload is sound (space-join/split preserves the stemmed tokens; sentiment/rating keep doc alignment; `_shared_cosine` reindexes by token string — full 2539/2539 overlap, no vocab-order dependence).
- Determinism/float risk is low; the 0.089 headroom far exceeds cross-platform noise, so the gate fires on intentional algorithmic changes — exactly its purpose.

Two wording nits folded into the follow-up commit: since R's self-consistency is ~0.998, the bar is framed honestly as an externally calibrated ~0.80 cross-implementation regression threshold (not "as consistent as R with itself"), and "mean sentiment" is now the more precise "neutral latent sentiment (α^(s)=0)".

Once merged, **#454 can be closed**.

nealcaren:
**Review note (independent adversarial pass) — gate specificity, not blocking the freeze itself**

The offline freeze is well-engineered: hermetic (verified with a poisoned `Rscript` on PATH and an import-blocking CI sim), deterministic (refit reproduces `topica_cosine=0.8868` bit-for-bit, +0.089 over the bar), and the `@slow` parity test genuinely runs in CI (no slow-deselection in the repo). No R at test time.

One substantive finding worth recording: **the gate is not specific to the adjusted (`reference="cran"`) profile it is named for.** Refitting the same frozen corpus under each profile and scoring cosine vs the gold `beta1` (bar = 0.7978):

| profile | cosine | result |
|---|---|---|
| `reference="cran"` (adjusted, the target) | 0.8868 | PASS |
| `reference="none"` (default **ridge**) | **0.8252** | **PASS** |
| `kappa_estimation="lasso"` | 0.7500 | below bar |
| `reference="paper"` | 0.3574 | below bar |

So a silent regression of the φ-mass-weighted adjusted aggregation + half-step damping back to plain ridge would still pass. The gate catches gross breakage (estimator swap, wrong init, shuffle) but has little power over the adjusted-vs-ridge distinction. Two prose claims are therefore stronger than the gold can back:

- `tests/test_sts_cran_gold.py` `test_sts_cran_gold_is_non_vacuous` docstring — "discriminates a **correct adjusted-profile fit from a wrong one**"; the shuffle control only proves a *shuffled* β fails.
- `parity/sts_cran_gold.py` / `docs/replications/sts.md` — framed as pinning "the whole adjusted-profile EM end to end"; true only for gross drift.

Cheap, platform-robust fix if we want the specificity later: add a **relative** assertion — fit the corpus once with `reference="cran"` and once with the ridge default and assert `cos(cran) > cos(ridge) + small_margin`. Both fits drift together across platforms, so the ~0.062 gap is stable and directly protects the adjusted φ-weighting. (Absolute bar can't just be tightened — headroom to ridge is only ~0.027, flaky.)

Leaving this as a note rather than a change request: as a faithful freeze of the already-accepted live check (same 0.80 floor), the PR is sound. Filing the relative-assertion strengthening as a follow-up.

PR #522, opened 2026-07-25
fix(guards): route FASTopic/InfoCTM *_tol through ensure_finite_nonneg (#517)

### LINKED ISSUE #517: Guard-parity: route FASTopic/InfoCTM convergence_tol (+ FASTopic sinkhorn_tol) through ensure_finite_nonneg
Follow-up from the #510 review (a non-blocking consistency gap; **not** a correctness bug).

#510 added `ensure_finite_nonneg("convergence_tol", ...)` guards to IdealPointTM, IdealPointSentenceTM, and Wordfish, but three sibling `*_tol` params remain unguarded:
- `FASTopic.convergence_tol` and `FASTopic.sinkhorn_tol` (`src/python/mod.rs`)
- `InfoCTM.convergence_tol` (`src/python/neural.rs`)

These feed only comparison-based early-stops (`fastopic.rs:596` `if em_tol > 0.0`, `:599` `rel < em_tol`; the Sinkhorn `err < tol`), so a NaN/+inf value merely disables early-stopping and the fit runs to max iters — no NaN topics, unlike the true #481 class. It's purely a guard-parity/consistency nit: the same param is validated in three models and not these.

Fix: route all three through `ensure_finite_nonneg` at construction for parity. Cheap, no behavior change for valid inputs.

Context: PR #510 review.

nealcaren:
Closes #517.

Follow-up from the #510 review — a guard-parity consistency gap, **not** a correctness bug. #510 added `ensure_finite_nonneg("convergence_tol", ...)` to IdealPointTM, IdealPointSentenceTM, and Wordfish, but three sibling `*_tol` params stayed unguarded:

- `FASTopic.convergence_tol` and `FASTopic.sinkhorn_tol` (`src/python/mod.rs`)
- `InfoCTM.convergence_tol` (`src/python/neural.rs`)

These feed only comparison-based early-stops (`if tol > 0.0`, `rel < tol`, and the Sinkhorn `err < tol`), so a NaN/+inf value merely disabled early-stopping and the fit ran to max iters — no NaN topics, unlike the true #481 class. This routes all three through `ensure_finite_nonneg` at construction for parity. Cheap, no behavior change for valid inputs.

## Verification

- `pytest tests/test_fastopic.py tests/test_infoctm.py` — 20 passed (construction-validation cases extended to reject NaN / +inf / negative `*_tol`)
- `./scripts/preflight.sh` — fmt + clippy + #481 guard scan + table/stub sync all green

### DIFF
--- a/src/python/mod.rs
+++ b/src/python/mod.rs
@@ -13261,8 +13261,10 @@ impl FASTopic {
         ensure_finite_pos("tw_alpha", tw_alpha)?;
         ensure_finite_pos("lr", lr)?;
         // Guard-parity with the other *_tol params (#517): these feed only
-        // comparison-based early-stops (a NaN/+inf merely disables early stopping),
-        // so this is a consistency guard, not a #481 NaN-topic hazard.
+        // comparison-based early-stops (NaN disables early stopping; +inf would
+        // instead trip it after one step), so this is a consistency guard, not a
+        // #481 NaN-topic hazard. The fit-time `convergence_tol` override is guarded
+        // the same way below.
         ensure_finite_nonneg("convergence_tol", convergence_tol)?;
         ensure_finite_nonneg("sinkhorn_tol", sinkhorn_tol)?;
         Ok(FASTopic {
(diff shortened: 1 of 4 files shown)


nealcaren:
Dual review (Claude faithfulness + Codex `gpt-5.6-sol` adversarial):

- **Faithfulness:** merge-safe — `ensure_finite_nonneg` is the correct helper (0.0 must stay valid as the "disable early-stop" sentinel), the three constructor sinks are comparison-gated only, and this is genuinely not a #481 NaN-topic hazard.
- **Codex** caught a real guard-parity gap: `FASTopic.fit(convergence_tol=)`'s per-run override bypassed the constructor guard, so `fit(..., convergence_tol=float('inf'))` still slipped through. It also flagged the inline comment as inaccurate for `+inf` (which trips the early-stop after one step rather than disabling it).

Addressed in the follow-up commit: the fit-time override now routes through `ensure_finite_nonneg`, a fit-time rejection test was added, and the comments were corrected. Out-of-scope notes (other models' `convergence_tol` still unguarded but equally benign; pre-existing `sinkhorn_iters=0`) left as-is.

PR #523, opened 2026-07-25
fix(idealpoint-sentence): include sigma^2 Gaussian normalizer in reported log-likelihood (#499)

### LINKED ISSUE #499: IdealPointSentenceTM review: closed-form EM solves correct; reported log-likelihood drops the sigma^2 Gaussian normalizer (non-monotone ll_history + convergence test), num_dims>1 rotation gap
## Review result

Two independent Claude reviewers — a faithfulness reviewer (generative model + closed-form EM correctness + identifiability) and an adversarial reviewer (hand-deriving the M-step normal equations to *break* the linear solves) — plus a maintainer trace. (Codex ghosted and Gemini's headless sandbox blocked the file reads, so the second reviewer is a second Claude subagent with an adversarial brief.) IdealPointSentenceTM is a topica-original Gaussian-mixture ideal-point model over sentence embeddings with **closed-form EM** (weighted least squares + per-author linear solves, no gradient descent).

**Bottom line: the parameter estimates are correct — both closed-form solves are clean.** The joint `(mu_k, V_k)` WLS M-step and the per-author `x_a` linear system (the two most error-prone pieces) were hand-verified term-by-term: no B-matrix transpose, correct `[mu; V]` slicing, the `x`-solve correctly uses the `e − mu_k` residual to the *neutral* centroid (not the displaced mean), and the prior ridge is correctly scaled `σ²/xvar` against the `1/σ²` data term. The one real correctness issue is in the *reported* log-likelihood, which breaks the monotonicity guarantee and drives the convergence test; plus the same `num_dims>1` identifiability gap flagged for IdealPointTM (#498).

### 1. Reported log-likelihood omits the σ²-dependent Gaussian normalizer — non-monotone `ll_history`, drives the convergence test (MEDIUM)

`src/sentence_ideal.rs:259,273` (E-step / per-obs LL), consumed at `:276-277` and `:447-455` (convergence), surfaced as `log_likelihood`/`fit_history`. The per-observation LL is `logsumexp_t(log_pi[t] − ‖e − mean_t‖²/(2σ²))`, i.e. it drops the true GMM density's `(2πσ²)^(−D/2)` factor. So

```
reported_ll = true_ll + (N·D/2)·ln(2πσ²)
```

(verified by derivation: `log Σ_k pi_k (2πσ²)^{−D/2} exp(−‖·‖²/2σ²) = −D/2·ln(2πσ²) + logsumexp_k(log pi_k − ‖·‖²/2σ²)`; the `−D/2·ln(2πσ²)` term is omitted). Because `σ²` is an M-step parameter that updates every sweep (`:424`, and typically shrinks sharply early), the omitted term is **not a constant offset** — it changes each iteration. Consequence: the *true* incomplete-data LL is monotone (the ECM M-step is valid), but the **reported** `ll_history` is `true_ll` plus a term that decreases as `σ²` shrinks, so the user-visible trace can go **down while the model improves**, violating the EM-monotonicity guarantee for the quantity actually reported. The convergence criterion at `:449` tests the relative change of this wrong quantity — it still converges once `σ²` stabilizes (the offset's change → 0), but early-iteration convergence detection is on the wrong landscape.

- [ ] Add `−(D/2)·ln(2πσ²)` to each per-obs term (`:273`), or equivalently `−(N·D/2)·ln(2πσ²)` to the summed `ll` before pushing to history and testing convergence. Then `ll_history` is the actual EM objective and is monotone. (Also note: `log_likelihood` currently reflects the pre-final-M-step value, `:470` — a separate one-iteration lag; fixing the normalizer and computing the LL after the M-step resolves both.)

### 2. `num_dims > 1` positions identified only up to rotation; undocumented (MEDIUM — mirror of #498 finding 1)

`src/sentence_ideal.rs:569-599` (`standardize_positions`), sign flip `:428-445`; stub `python/topica/_topica.pyi:4862-4869`. `standardize_positions` constrains only per-dimension mean-0 / unit-variance (losslessly absorbed into `mu`/`V`) and flips the sign of dimension 0 to the anchors. For `d ≥ 2` the model `e ~ N(mu_k + Σ_j x_{a,j} V_{k,j}, σ²I)` is invariant under any joint orthogonal rotation `x→xR, V→R^{-1}V` preserving unit per-dimension variance, and dimensions `1..d-1` have no sign fix — so `author_positions` and `position_centroid` are not interpretable coordinate-by-coordinate for `num_dims≥2`, reproducible only in the fixed-seed sense. Default `num_dims=1` is safe (only sign is free, fixed by anchors), so real-world impact is bounded, but the API exposes `num_dims` with no warning.

- [ ] Enforce a canonical rotation (decreasing-variance axes + per-axis sign via anchors), or document in the constructor/`author_positions` docstring and stub that multi-dimensional positions are identified only up to rotation and should be read through the loadings. (Fix once for the whole ideal-point family; same resolution as #498.)

### 3. `convergence_tol` escapes the #481 finiteness guards (LOW)

`src/python/sentence_ideal.rs:100-102` guards `x_prior_variance` via `finite_pos` but not `convergence_tol` (constructor or `fit` override). Benign in isolation — `|Δll|/denom < NaN` is always false, so a NaN/negative tol just runs all `iters` — and there is no covariance-feeding parameter here (unlike #498's `sigma_shrink`), so no NaN-topic path. Flagged only for guard-discipline parity with the family.

- [ ] Validate `convergence_tol >= 0.0 && is_finite()` in the constructor.

### Checked and CORRECT (hand-verified, both reviewers)

**Generative model & E-step:** `topic_mean`/`position_topic_centroid` form `mu_k + Σ_j x_{a,j} V_{k,j}` exactly; spherical `σ²I`; responsibilities are `pi_k·N(...)` with correct max-subtraction log-sum-exp (dropping the shared normalizer is harmless *here* — it cancels in the per-row normalization; it only corrupts the reported LL, #1). **Joint `(mu_k, V_k)` WLS** (`:293-345`): design `g=[1, x_a]` (intercept + all `d` columns), `A = Σ_i r_ik g gᵀ`, `B = Σ_i r_ik g e_iᵀ`, per-topic `spd_inverse` with a `1e-6` ridge, `c = A^{-1}B` read as `mu_k=c[0]`, `V_{k,j}=c[1+j]` — exactly `(ZᵀWZ)^{-1}ZᵀWE`, no transpose. **Per-author `x_a` solve** (`:350-398`): `A[l][j] = Σ_{i,k} r_ik(V_{k,l}·V_{k,j}) + (σ²/xvar)δ_lj`, `b[l] = Σ_{i,k} r_ik V_{k,l}·(e_i − mu_k)` — matches the stationary condition term-for-term, including the correct `e − mu_k` (neutral centroid) residual and the `σ²·inv_xvar` prior ridge; singular fallback leaves `x_a` unchanged. **σ² update** (`:402-424`): spherical-GMM MLE `Σ r_ik‖e − mean_k‖²/(N·D)`, floored `1e-8`, using the just-updated `mu/V/x` (valid ECM). **Identifiability standardization lossless** (centering into `mu`, scaling into `V`; displaced means invariant). **`position_se`** (`:79-148`) is the exact posterior precision `H_a = Σ_k N_{a,k}G_k/σ² + I/xvar`, consistent with the `x`-solve, capped at `sqrt(xvar)`. **Determinism** (index-ordered parallel reductions summed sequentially). **Robustness:** empty topic → `A=1e-6·I, B=0 → mu=V=0` (finite); `pi` floored `1e-12`, `σ²` floored `1e-8`; `K=1`/single-author blocked at the binding. **Save/load** round-trips with the experimental gate re-checked on load; stub in sync; defaults (`num_dims=1`, `convergence_tol=1e-4`, `x_prior_variance=1.0`, `iters=100`) consistent.

### Bottom line

The closed-form EM produces correct estimates. The actionable items are the LL normalizer (#1 — it directly affects the item-6 monotonicity guarantee and the convergence test) and documenting/constraining `num_dims>1` (#2, shared with #498); #3 is a consistency nit.

Relevant surfaces: `src/sentence_ideal.rs`, `src/python/sentence_ideal.rs`, `tests/test_sentence_ideal.py`, `python/topica/_topica.pyi:4820-4880`.

nealcaren:
Closes #499.

## What changed

`IdealPointSentenceTM`'s reported per-iteration log-likelihood dropped the spherical Gaussian normalizer `-(D/2) ln(2 pi sigma^2)`. Because `sigma^2` is re-estimated every EM sweep (and typically shrinks sharply in the early iterations), the omitted term is not a constant offset. It changes each iteration, so the user-visible `ll_history` (`reported_ll = true_ll + (N*D/2) ln(2 pi sigma^2)`) could go **down while the model was actually improving**. That violates the EM monotonicity guarantee for the quantity we report and drives the relative-change convergence test on the wrong landscape.

The fix adds `-(D/2) ln(2 pi sigma^2)` to each per-observation term in the E-step log-likelihood, using the same `sigma2` the responsibilities use. Now the reported value is the true incomplete-data log-likelihood, which is monotone non-decreasing across EM iterations.

- `src/sentence_ideal.rs`: compute `log_norm = -0.5 * D * ln(2 pi sigma2)` once per sweep and add it to each per-observation term.

This term is deliberately excluded from the responsibilities (where the shared normalizer cancels in the per-row simplex normalization), so **the fitted parameters are unchanged** — only the reported `ll`/`ll_history`/`fit_history` values differ.

A regression test in `tests/test_sentence_ideal.py` asserts `fit_history` is monotone non-decreasing across iterations (within a small f64 tolerance for the parallel reductions).

## What I deliberately skipped, and why

- **Item 2 (num_dims > 1 rotation non-identifiability, MEDIUM).** Already resolved on `main`: the `author_positions` docstring in `src/python/sentence_ideal.rs` already documents that for `num_dims > 1` the positions are identified only up to an arbitrary rotation and are best read through the loadings. No change needed.
- **Item 3 (`convergence_tol` #481 finiteness guard, LOW).** Already resolved on `main`: the constructor calls `ensure_finite_nonneg("convergence_tol", convergence_tol)`. No change needed.
- **The one-iteration LL lag** noted parenthetically in item 1 (`log_likelihood` reflects the pre-final-M-step value): left as-is. It does not affect `ll_history` monotonicity or the convergence test — every `ll_history` entry is computed consistently at its own E-step — so it is out of scope for this fix and would require an extra post-loop E-step. Kept the change minimal and behavior-preserving.

## Verification

- `pytest tests/test_sentence_ideal.py -q` → **9 passed** (includes the new `test_ll_history_is_monotone`).
- `./scripts/preflight.sh` → **all checks passed** (cargo fmt --check, clippy -D warnings, #481 guard scan, model-table sync, stub sync 78 passed). No binding signature changed, so `_topica.pyi` needed no update.
- Built with `maturin develop --release --features python` in an isolated `.venv-agent`.

### DIFF
--- a/src/python/sentence_ideal.rs
+++ b/src/python/sentence_ideal.rs
@@ -293,10 +293,21 @@ impl IdealPointSentenceTM {
         }
         Ok(Array1::from(m.position_topic_centroid(topic, &x)).to_pyarray_bound(py))
     }
+    /// The conditional incomplete-data log-likelihood of the *fitted* parameters,
+    /// `sum_i log sum_k pi_k N(e_i | mu_k + x_a V_k, sigma^2 I)` (a properly
+    /// normalized spherical-Gaussian mixture density). Computed on the returned
+    /// parameters, so unlike a mid-loop value it does not lag by an M-step.
     #[getter]
     fn log_likelihood(&self) -> PyResult<f64> {
         Ok(self.fitted_model()?.log_likelihood)
     }
+    /// Per-iteration `(iter, log_likelihood)` trace of the data log-likelihood at
+    /// each E-step, for convergence monitoring. Note this data log-likelihood is
+    /// **not** the objective EM maximizes here (the position M-step is MAP and the
+    /// positions are re-standardized to unit scale each sweep), so it is not
+    /// guaranteed monotone -- with `x_prior_variance` far from the unit
+    /// identification scale it can briefly decrease. Use the relative change, not
+    /// strict monotonicity, to judge convergence.
     #[getter]
     fn fit_history(&self) -> PyResult<Vec<(usize, f64)>> {
         Ok(self
--- a/src/sentence_ideal.rs
+++ b/src/sentence_ideal.rs
@@ -54,6 +54,43 @@ impl SentenceIdealModel {
         c
     }
 
+    /// The conditional incomplete-data (marginal) log-likelihood of the fitted
+    /// mixture on `emb`, holding the positions fixed:
+    /// `sum_i log sum_k pi_k N(e_i | mu_k + x_a V_k, sigma^2 I)`. A properly
+    /// normalized spherical-Gaussian density, so it includes the per-observation
+    /// normalizer `-(D/2) ln(2 pi sigma^2)`. This is the quantity `log_likelihood`
+    /// reports for the *returned* parameters (no one-iteration lag), and the oracle
+    /// the parity test recomputes independently.
+    pub fn incomplete_data_ll(&self, emb: &[Vec<f64>]) -> f64 {
+        if self.sigma2 <= 0.0 || !self.sigma2.is_finite() {
+            return f64::NAN;
+        }
+        let inv2s = 1.0 / (2.0 * self.sigma2);
+        let log_norm = -0.5 * self.dim as f64 * (2.0 * std::f64::consts::PI * self.sigma2).ln();
+        let log_pi: Vec<f64> = self.pi.iter().map(|&p| p.max(1e-300).ln()).collect();
+        emb.iter()
+            .zip(self.group.iter())
+            .map(|(e, &a)| {
+                let mut logr = vec![f64::NEG_INFINITY; self.num_topics];
+                let mut max = f64::NEG_INFINITY;
+                for t in 0..self.num_topics {
+                    let mean = self.position_topic_centroid(t, &self.x[a]);
+                    let mut sq = 0.0;
+                    for d in 0..self.dim {
+                        let z = e[d] - mean[d];
+                        sq += z * z;
+                    }
+                    logr[t] = log_pi[t] - inv2s * sq;
+                    if logr[t] > max {
+                        max = logr[t];
+                    }
+                }
+                let z: f64 = logr.iter().map(|&l| (l - max).exp()).sum();
+                max + z.ln() + log_norm
+            })
+            .sum()
+    }
+
     /// Per-topic discrimination `||V_k||` (Frobenius over the d x D loading).
     pub fn topic_discrimination(&self) -> Vec<f64> {
         self.v
@@ -240,10 +277,14 @@ pub fn fit_sentence_ideal<R: Rng>(
         // Per-observation Gaussian log-normalizer of the spherical density
         // N(.| ., sigma2 I) in D dimensions: -(D/2) ln(2 pi sigma2). It is
         // omitted from the responsibilities (it cancels in the per-row
-        // normalization) but MUST be included in the reported incomplete-data
-        // log-likelihood: because sigma2 is re-estimated each sweep, this term
-        // is not a constant offset, and dropping it makes `ll_history`
-        // non-monotone and the convergence test act on the wrong quantity.
+        // normalization) but is included in the reported incomplete-data
+        // log-likelihood so the value is a properly normalized log-density.
+        // Because sigma2 is re-estimated each sweep the term is not a constant
+        // offset, so dropping it distorts the per-iteration `ll_history` trace.
+        // (Note: `ll_history` is the data log-likelihood, which EM does not
+        // maximize here — the x M-step is MAP and the positions are re-standardized
+        // each sweep — so it is not guaranteed monotone when `x_prior_variance`
+        // differs from the unit identification scale; see the `fit_history` docs.)
         let log_norm = -0.5 * dim as f64 * (2.0 * std::f64::consts::PI * sigma2).ln();
         let log_pi: Vec<f64> = pi.iter().map(|&p| p.max(1e-300).ln()).collect();
         // Per-document log-likelihood, collected in document order then summed
@@ -463,7 +504,7 @@ pub fn fit_sentence_ideal<R: Rng>(
         prev_ll = ll;
     }
 
-    SentenceIdealModel {
+    let mut model = SentenceIdealModel {
         num_topics: k,
         dim,
         num_dims: dd,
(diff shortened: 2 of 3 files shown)


nealcaren:
Dual review: Claude faithfulness agent + **Codex** (`gpt-5.6-sol`) adversarial.

The normalizer fix is correct. But the reviewers **disagreed** on the PR's monotonicity claim, so I reproduced it directly: with early-stopping disabled the reported `fit_history` **does** decrease for reachable configs — e.g. **−9.06** at `x_prior_variance=0.05` (imbalanced authors), −0.81 at 0.2 — while it's monotone at the default `x_prior_variance=1.0`. Codex was right: the reported data log-likelihood is not the objective EM maximizes here (the position M-step is MAP and positions are re-standardized to unit scale each sweep), and reporting the penalized objective doesn't help because standardization pins the penalty term to a constant. The old test passed only because it used the default prior.

Amended (per Neal, 'honest fix, no false claim'):
- Keep the correct `-(D/2) ln(2πσ²)` normalizer.
- `log_likelihood` now reports `incomplete_data_ll` on the **returned** params, removing the one-M-step lag Codex also flagged.
- Replaced the monotonicity test with an **exact** Rust correctness test (reported ll == an independent inline recompute, across `x_prior_variance ∈ {0.05, 0.2, 1.0}`) + a Python normalized-density sanity check.
- Corrected the E-step comment and the `fit_history`/`log_likelihood` docstrings: `fit_history` is the data-LL convergence trace and is **not** guaranteed monotone.

Verified locally: `cargo test --lib sentence_ideal` (3 pass), `pytest tests/test_sentence_ideal.py` (9 pass), full preflight green. The deeper model fix (reconcile the identification scale with the prior for genuine monotonicity) is left as a separate issue.

PR #525, opened 2026-07-25
fix(party-embeddings): reject +inf sample, guard core panics, sync stub (#508)

### LINKED ISSUE #508: PartyEmbeddings review: PV-DM + neg-sampling gradient faithful to gensim (parity-gated, mean-factor + update-order correct); undocumented sign identifiability, unguarded sample, edge panics
## Review result

Two independent Claude reviewers — a faithfulness reviewer (PV-DM + negative sampling + PCA placement vs gensim Doc2Vec `dm=1` and Rheault & Cochrane, gensim 4.4.0 defaults confirmed empirically) and an adversarial reviewer (hand-tracing the negative-sampling gradient, the update ordering, and the mean factor to *break* them) — plus a maintainer trace. (Codex ghosted and Gemini's headless sandbox blocked the file reads, so the second reviewer is a second Claude subagent with an adversarial brief.) PartyEmbeddings is a PV-DM / doc2vec model with negative sampling (pure Rust, textbook word2vec update; the `partyembed` package was a black-box oracle only, so faithfulness = gensim PV-DM + the papers).

**Bottom line: the PV-DM architecture, the negative-sampling objective/gradient, the negative distribution, tag-vector handling, and the PCA placement are all correct — no correctness bug.** The adversarial reviewer confirmed the gradient is a line-for-line match of `word2vec.c` CBOW negative sampling: the `(label−f)·lr` sign is right, the **critical update order** is correct (`neu1e` reads the pre-update `syn1neg` at `:271` before `:272` overwrites it), `h`/`lr` are held constant across the k+1 targets, and **both context words and tag vectors receive the `dh` update** (so the party-placement signal is intact). The two priority risks resolve cleanly.

### The mean-factor question — CORRECT (gensim-faithful)

The top-flagged risk: the forward divides the hidden by the context+tag count (`src/party_embeddings.rs:244-247`, a mean), and the backward adds `neu1e` to each input **without** the `1/N` factor (`:281-295`). That looks like a missing mean-gradient factor, but it is exactly gensim's convention: in `doc2vec_inner.pyx`, with `cbow_mean=1` the error vector is scaled by `inv_count` *only in the sum path* (`if not cbow_mean`), never in the mean path. topica reproduces the oracle. The pure mean adjoint would carry `1/N`; gensim (the reference engine) does not, and topica matches — the planted-recovery test (`|r|>0.85`) confirms the asymmetry doesn't harm placement.

### 1. Unanchored placement sign is arbitrary and undocumented (LOW-MED, honesty — mirrors #505/#506/#498/#499)

`author_positions` docstring (`src/python/party_embeddings.rs:389-391`, stub `_topica.pyi:4998`) says the axis is "sign-oriented to the anchors," but when `anchors=None` the orientation block (`:363-373`) is skipped and the sign falls out of the (seeded, deterministic-but-substantively-arbitrary) randomized-PCA basis. PCA components are sign-invariant, so `PartyEmbeddings(...).fit(docs, group=...)` with no anchors can place "left" on either pole across seeds/corpora. Same class as the TBIP/Wordfish/IdealPoint sign non-identifiability — this is now a **family-wide pattern** across the ideal-point models worth a shared fix.

- [ ] Document that without `anchors` the sign of each placement dimension is arbitrary (anchors already resolve it), or default-orient to a deterministic reference.

### 2. `sample` is unguarded — NaN yields a silently untrained model (LOW, #481 class)

The constructor (`src/python/party_embeddings.rs:138-183`) guards `num_dims`, `vector_size`, `window`, `negative`, `learning_rate`, but not `sample`. In `keep_probs` (`src/party_embeddings.rs:118-133`), `sample<=0` means "keep all" (a negative `sample` silently accepted as flat), and `sample=NaN` makes every keep-probability NaN → `p>=1.0` and `rng<p` both false → every token dropped → every doc empty → `kept.is_empty()` continue (`:194`) → **no updates ever run**. The tag vectors stay at their random init, PCA runs on noise, and `fit()` returns a fitted-looking-but-untrained model with **no error**. Not a NaN-topic hazard, but the exact "invalid hyperparameter silently coerced" pattern the #481 arc targets.

- [ ] Reject non-finite/negative `sample` in the constructor (keep `0.0` legal = subsampling off).

### 3. LR-decay floor and default `sample` diverge from gensim (LOW, schedule)

`min_lr = (learning_rate*1e-4).min(1e-4)` (`src/python/party_embeddings.rs:341`) = `2.5e-6` at the default `learning_rate=0.025`, vs gensim's `min_alpha=1e-4` — so topica decays roughly to zero rather than to gensim's floor. Separately, the default `sample=1e-4` differs from gensim's `1e-3`. Both are defensible (and `vector_size=200`/`window=20` correctly follow the paper over gensim's `100`/`5`), but not the gensim numbers.

- [ ] Optional: note in the docstring which defaults follow the paper vs gensim, or align `min_lr` to `1e-4`.

### 4. `standardize_columns` discards the PCs' relative-variance scale (LOW, honesty)

Placement standardizes each PCA column to unit variance (`src/python/party_embeddings.rs:89-103`, applied `:362`), so `author_positions[:,0]` (PC1) and `[:,1]` (PC2) come out equally scaled — the information that PC1 dominates is lost. Placement also uses the randomized (approximate) `reduce::pca`, not exact SVD (fine at 10–12 parties × 200 dims, seeded + oversampled + 2 power iterations, but an approximation; the paper reports raw PC scores). Raw `author_vectors` are exposed so a user can redo exact PCA — a good escape hatch.

- [ ] One-line docstring note that columns are independently standardized and placement uses randomized PCA.

### 5. Two unreachable edge panics (adversarial)

`src/party_embeddings.rs:215` `b = rng.gen::<u32>() as usize % cfg.window` panics (modulo by zero) if `window==0`; `:107` `cum.last().unwrap()` panics if `num_words==0`. Neither reachable from sane config (guards exist upstream), but cheap one-line guards.

### Checked and CONFIRMED CORRECT (traced + empirically confirmed vs gensim 4.4.0)

**PV-DM** hidden = mean of context-word input vectors + tag vector(s), center predicted from `syn1neg`; mean not concat (`cbow_mean=1`/`dm_concat=0`, confirmed); dynamic window shrink (`b = rand % window`) matches `shrink_windows=True`; `syn0`/tags init `U(-0.5,0.5)/M`, `syn1neg` zero. **Neg-sampling gradient** `f=sigmoid(dot)`, `g=(label−f)·lr`, `neu1e += g·syn1neg` (pre-update), `syn1neg += g·neu1` — standard word2vec ascent, per-row, center-as-negative and duplicate draws skipped (matches gensim). **dh back-distribution** to both context words and tags, no `1/N` in mean mode (gensim-faithful). **Negative distribution** `freq^0.75` cumulative + binary search (`ns_exponent=0.75`), `.max(1.0)` floor harmless. **Tags**: party tag always present, optional control at row `num_groups+ctrl`, both inputs + both updated; party placed, controls estimated-not-placed. **Placement** centered PCA on the party-tag matrix, leading `num_dims` PCs, per-dimension anchor sign flip. **Defaults** `learning_rate=0.025`=gensim `alpha`, `negative=5`, `vector_size=200`/`window=20` per R&C; `min_count.max(1)`, `iters≥1`/empty-corpus/`<2` groups/`<2` vocab rejected. **Determinism** single ChaCha8, fixed doc order, subsample short-circuit consumes no RNG when always-kept, sigmoid clamped ±6, loss floored 1e-7; `test_determinism` passes. **Honest parity**: `parity/party_embeddings_compare.py` genuinely runs gensim `Doc2Vec(dm=1)` and gates `r_topica>0.85`/`r_cross>0.80`, skips cleanly without gensim, with a committed planted-recovery offline gold — better separated than the TBIP harness (#506). Save/load round-trips all buffers; stub in full sync.

### Bottom line

The PV-DM and negative-sampling core is faithful to gensim (parity-gated), and the mean-factor convention is correct. The actionable items are the undocumented sign identifiability (#1, shared across the ideal-point family) and the `sample` guard (#2); the rest are schedule/docstring/edge-guard polish.

Relevant surfaces: `src/party_embeddings.rs`, `src/python/party_embeddings.rs`, `src/reduce.rs`, `tests/test_party_embeddings.py`, `parity/party_embeddings_compare.py`, `python/topica/_topica.pyi:4993-5078`.

nealcaren:
Closes #508.

Follow-up hardening for `PartyEmbeddings` from the #508 review. The PV-DM +
negative-sampling core is faithful to gensim and parity-gated; this PR does not
touch the learning algorithm. It closes the remaining actionable items.

## Changes

- **`sample` guard tightened to reject `+inf` (#481 class).** The constructor
  already rejected `NaN`/negative `sample`, but a hand-rolled check let `+inf`
  (a nonsensical subsampling threshold) through silently. Routed both `sample`
  and `learning_rate` through the shared `ensure_finite_nonneg` /
  `ensure_finite_pos` helpers so every non-finite value is rejected with a clean
  `ValueError`. `sample=0.0` (subsampling off) stays legal.
- **Core precondition asserts (defense-in-depth).** Added `window >= 1` and
  `num_words >= 1` asserts at the top of `fit_pvdm` so a direct core caller fails
  with a clear message instead of the opaque `% 0` (`:215`) or
  `last().unwrap()` (`:107`) panic the review flagged. Both are unreachable from
  the Python API (the constructor guards `window >= 1`, `fit` guards vocabulary
  `>= 2`), so this is core-robustness, not a user-facing fix.
- **Stub docstring for `author_positions`.** The binding getter documents the
  sign identifiability (axis identified only up to sign without `anchors`); the
  `.pyi` stub carried a bare `...`. Mirrored the note into the stub so the two
  stay in sync.

## Skipped (already landed / not open)

- **Sign identifiability in the binding docstring (item 1)** — already documented
  in the `author_positions` getter by #512; only the stub was out of sync.
- **`sample` NaN/negative guard (item 2)** — already added by #510; this PR only
  extends it to `+inf` via the shared helper.
- **Degenerate/empty-input panics (item 5)** — every Python-reachable degenerate
  input (empty corpus, one group, vocabulary < 2, all-empty docs, everything
  pruned by `min_count`) already raises a clean `ValueError`; verified
  empirically and now locked in by tests. The two core panics were the only
  remnant, addressed by the asserts above.
- **Items 3 (LR floor / default `sample` vs gensim) and 4 (standardized-column
  scale)** — informational/schedule notes; item 4's honesty caveat is already
  in the getter docstring. No behavior change warranted.

## Verification

- `pytest tests/test_party_embeddings.py -q` -> 20 passed (new: `sample` guard
  for NaN/inf/negative, `sample=0.0` legal, degenerate-input parametrization).
- `cargo test --lib party` -> 2 passed.
- `scripts/preflight.sh` -> all checks passed (fmt, clippy -D warnings, no
  NaN-admitting `<= 0.0` guards, model tables, stub sync 78 passed).

### DIFF
--- a/python/topica/_topica.pyi
+++ b/python/topica/_topica.pyi
@@ -5108,7 +5108,15 @@ class PartyEmbeddings:
     @property
     def num_authors(self) -> int: ...
     @property
-    def author_positions(self) -> numpy.typing.NDArray[numpy.float64]: ...
+    def author_positions(self) -> numpy.typing.NDArray[numpy.float64]:
+        """Party placements as (num_authors, num_dims): the leading principal
+        components of the party vectors, standardized to mean 0 / unit variance
+        (column 0 is the left-right scale). Identifiability: without anchors the
+        sign of each dimension is arbitrary (PCA fixes the axes by variance, so
+        each column is identified only up to sign); pass anchors to fit() to fix
+        the signs. Columns are standardized independently, so the relative scale
+        of the components is not preserved; read author_vectors if you need it."""
+        ...
     @property
     def author_names(self) -> list[str]: ...
     @property
--- a/src/party_embeddings.rs
+++ b/src/party_embeddings.rs
@@ -149,6 +149,13 @@ pub fn fit_pvdm(
     cfg: &PvdmConfig,
     seed: u64,
 ) -> PartyEmbeddingsModel {
+    // Preconditions the binding already enforces (window >= 1 in the constructor,
+    // vocabulary >= 2 words after pruning). Assert them here so a direct core caller
+    // fails with a clear message rather than an opaque `% 0` or `last().unwrap()`
+    // panic in the training loop (issue #508).
+    assert!(cfg.window >= 1, "window must be >= 1");
+    assert!(num_words >= 1, "num_words must be >= 1");
+
     let m = cfg.vector_size;
     let num_tags = num_groups + num_controls;
     let mut rng = ChaCha8Rng::seed_from_u64(seed);
--- a/src/python/party_embeddings.rs
+++ b/src/python/party_embeddings.rs
@@ -161,17 +161,12 @@ impl PartyEmbeddings {
         if negative < 1 {
             return Err(PyValueError::new_err("negative must be >= 1"));
         }
-        if !learning_rate.is_finite() || learning_rate <= 0.0 {
-            return Err(PyValueError::new_err("learning_rate must be > 0"));
-        }
+        ensure_finite_pos("learning_rate", learning_rate)?;
         // #481-class guard: a NaN `sample` makes every keep-probability NaN, so every
-        // token is dropped and the model trains on nothing yet returns "fitted".
-        // Keep `0.0` legal — it is the documented "subsampling off" value.
-        if sample.is_nan() || sample < 0.0 {
-            return Err(PyValueError::new_err(format!(
-                "sample must be >= 0 (0 disables subsampling); got {sample}"
-            )));
-        }
+        // token is dropped and the model trains on nothing yet returns "fitted", and a
+        // non-finite `sample` is a nonsensical threshold. Keep `0.0` legal — it is the
+        // documented "subsampling off" value — so this is the non-negative guard.
+        ensure_finite_nonneg("sample", sample)?;
         Ok(PartyEmbeddings {
             num_dims,
             vector_size,
--- a/tests/test_party_embeddings.py
+++ b/tests/test_party_embeddings.py
@@ -145,6 +145,43 @@ def test_bad_params():
         m.fit([["a", "b"], ["c", "d"]], group=["x"])  # wrong length
 
 
+@pytest.mark.parametrize("bad", [float("nan"), float("inf"), -1.0])
+def test_sample_guard(bad):
+    # #481-class guard: a non-finite or negative subsampling threshold must be
+    # rejected at construction (a NaN sample would drop every token and return a
+    # fitted-looking-but-untrained model; +inf is a nonsensical threshold).
+    with pytest.raises(ValueError):
+        topica.PartyEmbeddings(sample=bad)
+
+
+def test_sample_zero_is_legal():
+    # 0.0 means "subsampling off" and must be accepted.
+    m = topica.PartyEmbeddings(vector_size=8, window=3, min_count=1, negative=2,
+                               sample=0.0, seed=1)
+    m.fit([["a", "b"], ["b", "a"]], group=["x", "y"], iters=2)
+    assert m.author_positions.shape[0] == 2
+
+
+@pytest.mark.parametrize(
+    "docs, groups, kw",
+    [
+        ([], [], {}),                                   # empty corpus
+        ([["a", "b", "c"]], ["x"], {}),                 # one group
+        ([["a"], ["a"]], ["x", "y"], {}),               # vocabulary < 2 words
+        ([[], []], ["x", "y"], {}),                     # no in-vocabulary tokens
+        ([["a", "b"], ["c", "d"]], ["x", "y"], {"min_count": 99}),  # pruned to nothing
+    ],
+)
+def test_degenerate_inputs_raise_cleanly(docs, groups, kw):
+    # Each degenerate input must raise a clean ValueError, never a Rust panic
+    # (PanicException, which is a BaseException and not caught by `except ValueError`).
+    opts = {"vector_size": 8, "window": 3, "min_count": 1, "negative": 2,
+            "sample": 0.0, "seed": 1, **kw}
+    m = topica.PartyEmbeddings(**opts)
+    with pytest.raises(ValueError):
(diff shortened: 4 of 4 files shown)


nealcaren:
Dual review: Claude (faithfulness) + **Gemini `3.1-pro-high`** (adversarial, via Antigravity — Codex was ghosting). Both clean, no blockers, and independently traced the key risk: the new core `assert!(num_words >= 1)` / `assert!(window >= 1)` are unreachable from Python — all 5 degenerate inputs in `test_degenerate_inputs_raise_cleanly` bail with a clean `ValueError` before `fit_pvdm` (empty corpus → 'no documents'; vocab<2 / all-empty / min_count=99 → 'fewer than 2 words after pruning'; one group → 'needs at least 2 groups'). `num_words = vocab_pairs.len() >= 2` at the core, stronger than the assert. Helper pairing correct: `ensure_finite_nonneg` for `sample` closes the old +inf hole (`(+inf).is_nan()` was false), `sample=0.0` stays legal; `ensure_finite_pos` for `learning_rate`. Tests genuine, CI green.

PR #526, opened 2026-07-25
feat(pa): expose per-document doc_super proportions; lock theta ring-buffer guard (#497)

### LINKED ISSUE #497: PA review: PAM4L Gibbs sampler faithful (n_ds/n_dsk bookkeeping clean, approximations honestly scoped); minor: latent theta-ring-buffer panic, no doc_super output
## Review result

Two independent Claude reviewers — a faithfulness reviewer (vs MALLET `PAM4L` + Li & McCallum §3) and an adversarial reviewer (hand-tracing the (s,k) Gibbs cell, the `n_ds` double-decrement trap, and the `n_dsk` index bookkeeping to *break* it) — plus a maintainer trace. (Codex ghosted and Gemini's headless sandbox blocked the file reads, so the second reviewer is a second Claude subagent with an adversarial brief.)

**Bottom line: the four-level collapsed-Gibbs sampler is faithful — no correctness bug.** Both reviewers confirmed the (s,k) pair conditional matches MALLET `PAM4L.sampleTopicsForOneDoc` term-by-term, and the two approximations topica documents (fixed symmetric doc→super α; fixed-concentration moment-matching for α_s instead of Minka MLE) are honestly scoped and do not diverge materially for PAM's purpose. There is no numeric cross-implementation reference, so the planted self-consistency gold is the correct bar. Findings are minor/low.

### 1. Latent θ-draw ring-buffer panic — unreachable via the public constructor (LOW, defensive)

`src/pa.rs:471-476`. The θ-snapshot ring buffer guards `if theta_draws.len() < opts.cap { push } else { remove(0); push }`. If a caller ever passed `thin>0` **and** `cap=0`, the first snapshot takes the `else` branch and calls `remove(0)` on an empty `Vec` → out-of-bounds panic, contradicting the doc claim "cap=0 disables collection" (collection is actually gated on `thin`, `pa.rs:455`, not `cap`). In practice `ThetaDrawOpts::new` (`src/keyatm.rs:1615-1622`) forces `thin=0` whenever `cap==0`, so `thin>0 ⇒ cap≥1` and the branch is never reached. Robustness gap, not a live bug — PA reimplements the ring buffer inline instead of reusing `ThetaDrawOpts::maybe_collect`.

- [ ] Guard `remove(0)` behind a non-empty check, or reuse the shared `maybe_collect` so the invariant is enforced in one place.

### 2. No per-document doc→super distribution exposed (completeness note, LOW)

topica exposes `doc_topic` as the doc→**sub** marginal and `super_sub` as the *global* super→sub association, but never the per-document distribution over super-topics (`nds[d]` normalized). PAM's distinguishing output is precisely the correlated super-topic structure, and a user doing correlation analysis may want per-doc super proportions. Not a bug — doc→sub is the standard `num_topics`-shaped output — but a `doc_super` getter would round out the model.

### 3. Behavioral/schedule differences from MALLET (documented, LOW)

- **α_s adaptation runs only in the final 25% of sweeps** (`pa.rs:445`) vs MALLET's periodic `optimizeInterval` re-estimation after burn-in. At the default `iters=1000` that's ~250 re-estimates (ample); for short runs (tens of iters) adaptation is thin or absent.
- **Default `alpha=0.1` is much smaller than MALLET's effective super prior (~50/S)** which, combined with the single-super-topic commitment at init (`pa.rs:432`), makes documents commit hard to their initial super-topic early. This is the intended "give the super layer something to specialize on" design (`pa.rs:26-27,397-400`), honestly documented, but a real behavioral difference — worth a one-line note in the constructor docstring.
- **`doc_topic` smooths the sub-marginal with `alpha`** (the doc→super prior), not a sub-level prior (`pa.rs:113,464`). There's no canonical PAM doc→sub prior, so this is only an output-normalization floor on an already-derived marginal — negligible impact; a one-line comment noting the choice would remove doubt.

### 4. Perf micro-note (not correctness)

`pa.rs:220` recomputes `nds[d].iter().sum()` (O(S)) per token to build `super_norm`, which then cancels out of the sampling distribution. It could be dropped or tracked incrementally. Trivial cost at small S.

### Checked and CORRECT (traced against MALLET PAM4L + adversarial hand-derivation)

**(s,k) Gibbs equation** (`pa.rs:227-238`): `(n_ds+α)/(n_d+Sα) · (n_dsk+α_sk)/(n_ds+Σ_j α_sj) · (n_kw+β)/(n_k+Vβ)` — all three factors match MALLET's `superProb·subProb·wordProb`. **`n_ds` single-decrement (the key trap):** bound once to a local at `pa.rs:228` and reused for both the factor-1 numerator and factor-2 denominator; removed exactly once in `sweep` (`pa.rs:260`); `n_d` recomputed post-decrement — no double-decrement, no stale snapshot. **Leave-one-out** decrement/increment of all four count tensors (`nds`, `ndsk`, `nkw`, `nk`) symmetric (`pa.rs:244-247,260-263`). **`n_dsk` D×S×K** never flattened (only the disposable `probs` buffer is, with a self-inverse `s*k+sub` / `/k`,`%k` decode) — no s,k transposition; invariant `Σ_k n_dsk[d][s][k] = n_ds[d][s]` maintained in lockstep. **u32 underflow impossible** (each token contributes exactly 1 to each count). **Normalizer** `n_d+Sα` constant across pairs, cancels. **α_s moment-matching:** asymmetric init (super s prefers sub-topics near `s·K/S`, boost `alpha·5`), burn-in gate (`adapt_after = iters - iters/4`), `α_sk = (conc·p_sk).max(1e-3)` with `conc = 4·K·alpha` pinned (super_sub rows sum to 1 so no drift/collapse), floored ≥1e-3 (NaN/zero-proof). **Outputs:** φ=`(nkw+β)/(nk+Vβ)`, super_sub posterior mean, doc→sub θ. **Guards:** `finite_pos` on α/β (#481 class), `num_super≥1`, `num_sub≥2`. **Robustness:** empty/single doc, S=1 degenerates cleanly, K=1 blocked, zero-weight vector safe, no div-by-zero, determinism via `Pcg64Mcg::seed_from_u64` (unit-tested bit-identical). **Approximations honestly scoped:** the docstring's observation that unconstrained Minka MLE collapses α→0 on the four-level model's sparse per-doc sub-topic usage is correct; the fixed-concentration variant preserves PAM's correlated-topic structure and is accurately labeled. Stub and gold (planted self-consistency, correctly framed) in order.

### Bottom line

The PAM sampler math is faithful to the reference and the documented approximations are sound. The actionable items are the defensive ring-buffer guard (#1) and, optionally, exposing per-document super-topic proportions (#2); the rest are documentation notes.

Relevant surfaces: `src/pa.rs`, `src/python/hierarchical.rs:51-470`, `parity/pa_gold.py`, `python/topica/_topica.pyi:2776-2872`.

nealcaren:
Closes #497.

The PA review (#497) confirmed the PAM4L collapsed-Gibbs sampler is faithful with no correctness bug. This PR addresses the two actionable follow-ups. The sampler math is untouched.

## Implemented

**Item 2 — expose per-document doc->super proportions (`doc_super`).**
PA exposed `doc_topic` (doc->sub marginal) and `super_sub` (global super->sub association) but not the per-document super-topic mixture. Added a `doc_super` getter returning a `(num_docs, num_super)` matrix, `θ^s_{d,s} ∝ n_ds + alpha` normalized per row.

- Smoothing uses `alpha`, which is exactly the symmetric Dirichlet prior on the doc->super multinomial (the factor-1 term in the Gibbs conditional), so this is the natural posterior mean rather than an output-only floor. This mirrors how `doc_topic` uses `alpha` and is documented in the method docstring.
- Wired core (`src/pa.rs`) -> binding (`src/python/hierarchical.rs`) -> `.pyi` stub, with `save`/`load` round-trip (`#[serde(default)]` so pre-existing saves load cleanly).

## Verified already fixed (not re-implemented)

**Item 1 — theta-draw ring-buffer underflow guard.**
The defensive `cap > 0` guard around `theta_draws.remove(0)` was already shipped in PR #501 (`src/pa.rs`, commit `5a038d7`). Rather than re-add it, this PR locks the invariant with a cargo unit test that constructs the otherwise-unreachable degenerate pairing `ThetaDrawOpts { cap: 0, thin: 1 }` directly and asserts no panic / no collection. (`ThetaDrawOpts::new` forces `thin = 0` when `cap == 0`, so this pairing cannot arise from the public constructor.)

## Skipped

- **Item 3 (behavioral docstring notes)** — the MALLET behavioral differences (small default `alpha`, late α_s adaptation, `doc_topic` alpha-floor) are already documented in the `PA` class docstring (added by #501). No change needed.
- **Item 4 (perf micro-note)** — informational only; touching the sampler hot loop was explicitly out of scope.

## Verification

- `cargo test --lib pa::` — 5 passed (incl. new `doc_super_rows_are_probabilities_over_supers` asserting shape D×S / rows sum to 1 / non-negative / >0.8 planted-super recovery, and `theta_ring_buffer_no_underflow_on_degenerate_opts`).
- `pytest tests/test_pa_doc_super.py tests/test_pa_gold.py` — 8 passed (new `test_pa_doc_super.py`: shape, probability rows, decisive+non-degenerate assignment, save/load round-trip, requires-fit; gold unchanged).
- `./scripts/preflight.sh` — all checks passed (fmt, clippy, #481 guard scan, model-table sync, `.pyi` stub sync).

Note on the recovery sanity check: the Python test asserts `doc_super` is decisive (most rows commit to one super-topic) and non-degenerate (both super-topics dominate some docs) rather than planted-label recovery, because which planted group maps to which super-topic is seed-dependent (the super layer can specialize on an axis other than the planting — a documented PAM behavior). The Rust unit test pins planted-super recovery >0.8 on a seed that recovers cleanly.


nealcaren:
Dual review: Claude (faithfulness) + **Gemini `3.1-pro-high`** (adversarial). Both confirmed the `doc_super = (n_ds + alpha)` row-normalized math is faithful (`nds` is the D×S count matrix, `alpha` the exact doc→super Dirichlet prior, consistent with the sampler and `doc_topic`), the D×S shape correct, the uniform-fallback branch harmlessly dead given `alpha>0`, and the ring-buffer guard test + all unit tests sound.

One real finding (both reviewers): the save-compat claim. The save format is **positional bincode**, under which `#[serde(default)]` is inert for a mid-struct field — a genuine pre-#526 save does **not** round-trip to `doc_super=None`, it fails to deserialize — and the getter's `unwrap()` on `None` was a latent panic. Per the project's pre-v1.0 save-compat stance this isn't blocking, but the claim should be honest. **Amended**: corrected the field comment (documenting the bincode/serde-default trap, matching the mod.rs convention) and made the getter return a clean `RuntimeError` instead of unwrap-panicking. Verified: `cargo test --lib pa::` 5 pass, `pytest tests/test_pa_doc_super.py tests/test_pa_gold.py` 8 pass, preflight green.

Issue #422, opened 2026-07-23
SAGE review: implements a Gaussian ridge, not SAGE's sparse prior ('sparse' docstrings overclaim); opt-in convergence trips at iter 20 before any kappa update


nealcaren:
Independent correctness review of SAGE (`src/sage.rs` + binding in `src/python/mod.rs` ~5063-5400), via the Codex second-opinion agent (`gpt-5.6-sol`), source-only, with the headline findings re-verified against the code. The kappa objective/gradient, softmax stability, and the Gibbs sweep are correct. But the implementation is **not faithful to canonical SAGE**, and the opt-in convergence path is broken.

## 1. [High / faithfulness] The defining SAGE sparsity prior is absent — this is a Gaussian ridge content model

Canonical SAGE (Eisenstein, Ahmed & Xing, ICML 2011) is *defined* by a sparsity-inducing prior on kappa: each coordinate has a latent per-word variance tau with a Gamma/Jeffreys hyperprior, integrating to a compound-Gamma / Laplace that drives most kappa to ~0. `optimize_kappa` (`src/sage.rs:259`) instead applies a **fixed** Gaussian `N(0, prior_variance)` — plain L2 ridge, no tau, no inverse-variance update, no L1/proximal step. Consequences: kappa is dense, rare-word deviations are not adaptively shrunk, and the model is a "Gaussian-SAGE variant," not SAGE. It also deviates from STM's content model (R stm defaults to an L1/elastic-net content prior via distributed-Poisson glmnet).

The public claims are internally inconsistent (verified):
- `src/sage.rs:4` — "additive in **sparse** deviations from a background"
- `src/python/mod.rs:5467` — "SAGE's **sparse** additive parameterization"
- but the constructor `src/python/mod.rs:5161` is honest — "`prior_variance` is the **Gaussian** prior on the kappa content deviations."

**Action:** either (a) implement the adaptive tau / L1 prior to make it real SAGE, or (b) keep the Gaussian variant but drop "sparse" from the module doc and the `mod.rs:5467` docstring and rename/reframe it as a ridge content model. Also revisit what "validated against its reference implementation" means here — the gold test is self-consistency, not external SAGE parity.

## 2. [High, opt-in] Enabling `convergence_tol` converges spuriously at iter 20, before any kappa is learned

With kappa=0 at init, every cell's beta equals the same background softmax, so
`compute_ll = sum_c sum_v n_cv * ln(beta_cv) = sum_v ln(b_v) * N_v`
[...]

PR #529, opened 2026-07-25
fix(dtm): expose per-document doc_topic + gensim-gold state-space test (#494)

### LINKED ISSUE #494: DTM review: state-space core faithful to gensim line-for-line (gradient hand-verified); minor items: no em_min_iter floor, discarded gammas/no doc_topic, gensim-gold test gap
## Review result

Two independent Claude reviewers — one broad, one adversarial (tasked to hand-derive the gradient and *refute* the "mirrors gensim verbatim" claim) — plus a maintainer read. Codex ghosted and Gemini's headless sandbox blocked the file reads it needed, so the second reviewer is a second Claude subagent with an adversarial brief; both traced against the **actual installed gensim source** (`.venv-dev/.../gensim/models/ldaseqmodel.py`), not memory.

**Bottom line: the DTM state-space core is faithful — no correctness bug.** Every Kalman recursion (`compute_post_variance`/`compute_post_mean`, forward and backward), the `f_obs`/`df_obs` objective+gradient, `compute_mean_deriv`, `compute_bound`, `update_zeta`/`compute_expected_log_prob`, `counts_init`, `fit_sslm`, and the per-doc LDA inference match gensim line-for-line. The adversarial reviewer verified the `df_obs` gradient is the exact analytic derivative of `f_obs` (hand-derived, including the `mean[0]²/(2·1000·cv)` term and the `OBS_NORM_CUTOFF=2` data-term switch), and that the sign convention is consistent end-to-end (`f = -bound`, gradient negated, L-BFGS minimizes → maximizes the bound). The subtle trap — gensim's `compute_mean_deriv` feeding the *smoothed* `self.variance` into a forward-looking recursion — is reproduced correctly and documented (`dtm.rs:137-138`). The findings below are minor behavioral/coverage items.

### 1. Outer EM loop has no `em_min_iter` floor (LOW, behavioral divergence)

`src/dtm.rs:636-675`. gensim's `fit_lda_seq` loops `while iter_ < em_min_iter or ((convergence > 1e-4) and iter_ <= em_max_iter)` with `LdaSeqModel` defaults `em_min_iter=6, em_max_iter=20`. topica runs `for iter in 0..em_iters` and breaks as soon as `iter>0 && |(bound-old)/old| < 1e-4` — it can stop at iteration 1. The convergence threshold (1e-4), the "bound dipped → grow `lda_max_iter`" backoff (`LOWER_ITER=10, ITER_MULT_LOW=2`), and the starting `lda_inference_max_iter=25` all match exactly; only the minimum-iteration floor is missing. On easily-converged corpora topica may under-iterate vs gensim at the same nominal budget. The parity gold uses `iters=40`, so it's invisible there.

- [ ] Add gensim's minimum-iteration floor (6) before honoring the convergence break, or document the divergence.

### 2. `update_obs` omits gensim's `norm_cutoff_obs` optimization-skip cache (LOW, benign)

`src/dtm.rs:312-350` vs gensim `Sslm.update_obs`. gensim caches the first below-cutoff word's optimized `obs` and, for every subsequent word with `‖counts‖₂ < OBS_NORM_CUTOFF`, reuses the cached vector *without re-optimizing*. topica always re-optimizes each low-count word from its own `x0`. Both reviewers independently proved this is benign: for a zeroed count vector the objective/gradient depend only on `totals`, `zeta`, `cv`, and the **word-independent** variances, so all zero-count words share one optimum — topica's re-optimization lands where gensim's cache points, differing only in the last bits and in (wasted) compute on large sparse vocabularies. Arguably topica is *more* correct.

- [ ] Optional: replicate the cache only if bit-tracking gensim matters; otherwise document.

### 3. Per-document topic proportions (`gammas`) are discarded — no `doc_topic` output (feature gap)

`src/dtm.rs:650` drops the E-step `gamma` (`let (_, phi, lhood) = ...`); only φ-weighted suff-stats are accumulated. gensim retains `self.gammas` (D×K) and exposes per-doc topic mixtures. topica's `DTM` PyO3 class has no `doc_topic`/`gammas` accessor and `Estimator::doc_topic` returns `Vec::new()` (marked EXEMPT, `dtm.rs:702-705`). The math that consumes γ is faithful — this is a feature gap, not a fidelity bug, but users expecting gensim's doc-topic output won't find it.

- [ ] Decide whether to expose per-document topic proportions; at minimum note the omission in the docstring.

### 4. Test coverage would not catch a state-space regression (test-coverage gap)

There is no gold-numbers test tying any recursion output to gensim. `spectral_init_is_seed_independent` and `random_init_reproducible_for_fixed_seed` assert only topica==topica (self-consistency); `recovers_drifting_topic` checks qualitative drift direction on planted data — it would likely catch a gross `df_obs` sign flip but **not** a subtler error (a wrong backward-variance weighting, a dropped `variance/2` in `zeta`), since those still produce plausible drift. Single-time-slice (`num_times==1`) is numerically handled but untested.

- [ ] Add a fixed-input test comparing `compute_post_variance`/`compute_post_mean`/`obs_deriv` against values dumped from gensim for a tiny V×T case, to guard the "verbatim" claim. Add a `num_times==1` smoke test.

### 5. Trivial: stub advertises an unreachable `initialization` value

`python/topica/_topica.pyi:1156-1158` lists `'provided'` as a possible `initialization`; `fit_dtm` only ever sets `"spectral"`, `"random-fallback"`, or `"random"` (`dtm.rs:601-610`). Cosmetic stub drift.

### Verified CORRECT (traced against installed gensim + hand-derivation)

Forward/backward variance recursions incl. `fwd_variance[0]=chain_variance·1000`, backward `c=(fv/(fv+cv))²`, `var[t]=c(var[t+1]-cv)+(1-c)fv` (`dtm.rs:73-96`); forward/backward mean incl. `c=cv/(fv[t]+cv)` interpolation (`dtm.rs:100-116`); `f_obs`/`obs_deriv` objective and analytically-consistent gradient with correct signs (`dtm.rs:188-244, 327-342`); `compute_mean_deriv` smoothed-variance quirk (`dtm.rs:139-160`); `zeta` including `+variance/2` and the `+1` slice offset (`dtm.rs:118-134`); `topic_word(t)` softmax algebraically identical to gensim's normalized `exp(e_log_prob[:,t])` (zeta cancels); `compute_bound`, `fit_sslm` (`sslm_max_iter=2`, 1e-6), `counts_init`; per-doc `fit_lda_post` (γ/φ/lhood, slice-specific `e_log_prob`, 1e-8 convergence); constants `INIT_VARIANCE_CONST=1000`, `OBS_NORM_CUTOFF=2`, `chain_variance=0.005`, `obs_variance=0.5`, `alpha=0.01`; hyperparameter guards via `finite_pos` (#481 class), non-negative/contiguous time-slice validation with empty-slice rejection; `log_add` stability; L-BFGS rejects non-finite trials (#419); determinism via `ChaCha8Rng` with the parallel E-step folded serially in doc order; save/load round-trip.

### Documented, acceptable divergences (not bugs)

L-BFGS (`50, 5, 1e-3`) vs gensim's `fmin_cg` (`gtol=1e-2, epsilon=1e-3`) — statistical, not bit, parity, documented in the module header; topica's collapsed-Gibbs/spectral seed vs gensim's `LdaModel` seed, documented, with the parity harness validating per-slice/per-topic cosine ~0.999 (bar 0.90) on a smooth Gaussian-random-walk corpus (the honest regime — abrupt drift isn't cross-implementation identifiable).

Relevant surfaces: `src/dtm.rs`, `src/python/mod.rs:10043-10515`, `topica-core/src/variational/lbfgs.rs`, `tests/test_dtm.py`, `parity/dtm_gold.py`, `python/topica/_topica.pyi:1149-1218`.

nealcaren:
Closes #494.

Two follow-ups from the DTM faithfulness review. The state-space core was verified faithful line-for-line to gensim and is untouched (no Kalman/`f_obs`/`df_obs`/`compute_bound` changes).

## Implemented

- **Item 3 — per-document `doc_topic`.** The E-step `gamma`s that gensim retains as `self.gammas` were discarded (`let (_, phi, lhood) = ...`). We now retain the final-iteration gammas on `DtmModel`, row-normalize them into a D×K matrix, and expose it through the `DTM.doc_topic` getter and `Estimator::doc_topic` (previously EXEMPT/empty). Save/load carries it with `#[serde(default)]`, so pre-existing saves still load (the field comes back `None`). No EM math changed — the gamma was already computed every iteration — so `topic_word` output is bit-identical to before.
- **Item 4 — state-space regression test.** `state_space_recursions_match_gensim` is a fixed-input V=3, T=2 check of `compute_post_variance` / `compute_post_mean` / `update_zeta` / `compute_mean_deriv` / `obs_deriv`. The expected values are **real numbers dumped from gensim 4.4.0's `sslm`** (same `chain_variance=0.005`, `obs_variance=0.5`, and a fixed observation matrix), hardcoded as literals — no gensim import at test time, per the offline-gold policy. Also adds `single_time_slice_smoke` (`num_times==1`), a Rust `doc_topic` shape/normalization test, and Python tests for `doc_topic`, single-slice fit, and a `doc_topic` save/load round-trip.

## Already done / skipped

- **Item 1 (`em_min_iter` floor)** — already shipped in #501 (`EM_MIN_ITER=6`, `src/dtm.rs`), verified present and honored before the convergence break.
- **Item 5 (stub `'provided'` drift)** — already resolved; the stub no longer lists `'provided'`, verified.
- **Item 2 (`norm_cutoff_obs` cache)** — skipped as documented-benign (optional bit-parity/perf item; topica's re-optimization lands at the same optimum).

## Verification

- `cargo test --lib dtm` — 8 passed (incl. the new gensim regression + single-slice + doc_topic tests).
- `pytest tests/test_dtm.py -q` — 20 passed.
- `./scripts/preflight.sh` — all checks passed (fmt, clippy `-D warnings`, `#481` guard scan, model-table sync, stub sync). The pre-push hook re-ran it green.
- The reference-toolchain gold tests (`test_dtm_gold.py`, `test_combinedtm_gold.py`) error on `import scipy`/torch in the deliberately-minimal agent venv, before any DTM logic runs; they are unaffected by this change (the DTM gold refit compares `topic_word`, which is bit-identical).

Adding `doc_topic` did not change any existing output shape or value; it is a new, additive getter.

### DIFF
--- a/python/topica/_topica.pyi
+++ b/python/topica/_topica.pyi
@@ -1153,8 +1153,9 @@ class DTM:
     gensim's LdaSeqModel). Query a topic's distribution at a slice with
     topic_word(time) and a word's trajectory with word_evolution(topic, word).
 
-    Exposes evolving topic-word distributions but no per-document `doc_topic` —
-    it targets topic evolution, not per-doc mixtures (#494)."""
+    Topics are shared across slices, so it also exposes per-document topic
+    proportions via `doc_topic` (the final-iteration variational gammas,
+    row-normalized); the topic-word distributions are what evolve (#494)."""
     @property
     def initialization(self) -> str | None:
         """The initialization route the fit took (#410): 'spectral',
@@ -1204,6 +1205,13 @@ class DTM:
 
     def topic_word(self, time: int) -> numpy.typing.NDArray[numpy.float64]:
         """Topic-word matrix at `time`, shape (num_topics, num_words); rows sum to 1."""
+        ...
+
+    @property
+    def doc_topic(self) -> numpy.typing.NDArray[numpy.float64]:
+        """Per-document topic proportions, shape (num_docs, num_topics); rows
+        sum to 1 (#494). The final-iteration variational gammas (gensim's
+        self.gammas), row-normalized. Topics are shared across slices."""
         ...
 
     def word_evolution(
--- a/src/dtm.rs
+++ b/src/dtm.rs
@@ -480,6 +480,10 @@ pub struct DtmModel {
     pub alpha: f64,
     chains: Vec<Sslm>,
     pub bound: f64,
+    /// Per-document topic proportions (D×K, rows sum to 1), from the final
+    /// E-step (issue #494). gensim retains the raw variational `gamma`s as
+    /// `self.gammas`; we row-normalize them into topic proportions.
+    pub doc_topic: Vec<Vec<f64>>,
     /// The initialization route the fit actually took (issue #410): `"spectral"`,
     /// `"random-fallback"` (spectral requested but recovery returned `None`, so the
     /// seeded static-LDA init ran), or `"random"` (`init="random"`).
@@ -633,6 +637,10 @@ pub fn fit_dtm<R: Rng>(
 
     let mut bound = 0.0;
     let mut lda_max_iter = 25usize;
+    // Per-document variational gammas from the most recent E-step (issue #494).
+    // gensim keeps these as `self.gammas`; we normalize them into `doc_topic`
+    // once EM finishes. Seeded to the uniform prior for the empty-loop edge case.
+    let mut gammas: Vec<Vec<f64>> = vec![vec![alpha; k]; docs.len()];
     // gensim's fit_lda_seq runs at least `em_min_iter` EM sweeps before honoring the
     // convergence break (`while iter_ < em_min_iter or ...`); default 6. Bounded by
     // the user's `em_iters` budget, so a smaller budget still caps the work.
@@ -646,18 +654,19 @@ pub fn fit_dtm<R: Rng>(
         // Per-document inference is independent; run it in parallel and fold the
         // suff-stats in serially (document order) so the fit is bit-for-bit
         // identical regardless of thread count.
-        let doc_results: Vec<(usize, Vec<Vec<f64>>, f64)> = bags
+        let doc_results: Vec<(usize, Vec<f64>, Vec<Vec<f64>>, f64)> = bags
             .par_iter()
             .enumerate()
             .map(|(d, bag)| {
                 let ti = times[d];
-                let (_, phi, lhood) = fit_lda_post(bag, ti, &topic_elp, alpha, lda_max_iter);
-                (d, phi, lhood)
+                let (gamma, phi, lhood) = fit_lda_post(bag, ti, &topic_elp, alpha, lda_max_iter);
+                (d, gamma, phi, lhood)
             })
             .collect();
-        for (d, phi, lhood) in &doc_results {
+        for (d, gamma, phi, lhood) in &doc_results {
             let ti = times[*d];
             doc_bound += *lhood;
+            gammas[*d].clone_from(gamma);
             for (n, &(word, count)) in bags[*d].iter().enumerate() {
                 for kk in 0..k {
                     sstats[kk][word][ti] += count * phi[n][kk];
@@ -681,6 +690,19 @@ pub fn fit_dtm<R: Rng>(
         }
     }
 
+    // Row-normalize the final gammas into per-document topic proportions.
+    let doc_topic: Vec<Vec<f64>> = gammas
+        .iter()
+        .map(|g| {
+            let s: f64 = g.iter().sum();
+            if s > 0.0 {
+                g.iter().map(|&x| x / s).collect()
+            } else {
+                vec![1.0 / k as f64; k]
+            }
+        })
+        .collect();
+
     DtmModel {
         num_topics: k,
         num_times: t,
@@ -688,6 +710,7 @@ pub fn fit_dtm<R: Rng>(
         alpha,
         chains,
         bound,
+        doc_topic,
         initialization: init_route.to_string(),
     }
 }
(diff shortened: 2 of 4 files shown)


nealcaren:
Dual review: Claude (faithfulness) + **Gemini `3.1-pro-high`** (adversarial). Both confirmed the core is faithful: `doc_topic` is the row-normalized final-iteration variational gammas (exactly gensim's `gammas[d]/sum`), `topic_word` is bit-identical (the only change is retaining an already-computed value in deterministic doc order), and the gensim-gold state-space test is a genuine non-vacuous regression guard (23 outputs at 1e-10).

One real finding (both reviewers): the save-compat claim. Under **positional bincode** `#[serde(default)]` is inert, so a genuine pre-#494 save does **not** load with `doc_topic=None`. The Claude reviewer pinned it precisely: because `doc_topic` is appended at the **end** of `DtmState`, an old save hits a clean **EOF** on the trailing field and fails to deserialize with a `PyValueError` (no misalignment, no panic — distinct from #526 where the field was mid-struct). So the PR body's 'pre-existing saves still load (the field comes back None)' is inaccurate — old saves error at load, they don't load with None. Per the pre-v1.0 save-compat policy this isn't blocking. **Amended**: corrected the field comment to the honest EOF wording (matching the #422/#491 convention), and made the getter return a clean `RuntimeError` instead of unwrap-panicking (defense-in-depth). Verified: `cargo test --lib dtm` 8 pass, `pytest tests/test_dtm.py` 20 pass, preflight green.

PR #531, opened 2026-07-25
fix(stm): guard content-prior scales, surface degenerate SE/bootstrap cases

nealcaren:
Adversarial review of topica's **faSTM-beyond-STM extensions** (the content L1 prior, the temporal content covariate, and `estimate_effect`) surfaced several issues. This PR lands the three clear guard/robustness fixes (findings 1–3); findings 4–5 are filed as separate issues (docs/behavior decisions), and the high-severity lexical-period-ordering bug is tracked separately.

All fixes are confirmed with reproductions and change nothing for valid inputs.

## 1. Unguarded content-prior scales → silent NaN / silent no-op (#481-class)
- `content_prior_var` flows into `1/content_prior_var` (the L2 precision, or the L1 rate under `content_prior="l1"`). `content_prior_var=0.0` returned a fitted model with **all-NaN `content_kappa` and a NaN `bound`, no error**; negatives gave anti-shrinkage. Now rejected via `ensure_finite_pos` when a content model is used.
- `content_smooth` (the RW precision `1/tau^2`) was gated `> 0.0`, so a NaN/negative value **silently disabled** the smoothing the user requested, and `+inf` poisons the RW penalty. Now `ensure_finite_nonneg` when `content_time` is given (`0.0` stays legal = fully saturated content factor).

## 2. Bootstrap returned a silent all-NaN band when every refit failed
`content_divergence`/`content_trajectory` with `ci=True` wrapped each refit in a bare `except: continue`. A `fit_kwargs` that can't reproduce the fit (e.g. drops `content_time`) makes **every** replicate fail, and the old code returned an all-NaN CI with no signal — read as "wide uncertainty" rather than "the bootstrap never ran." Now raises `RuntimeError` on zero usable replicates and warns when >half fail.

## 3. Cluster-robust SEs silently degenerate with too few clusters
`estimate_effect(cluster=)` with `< 2` clusters collapses the CR1 sandwich meat to ~0 (SEs spuriously near zero); with `G <= p` the cluster vcov is rank-deficient (understated SEs). Both were silent. Now warns in both cases.

## What the review verified CORRECT (no change)
The hard algorithmic machinery all checked out against references/derivation: the FISTA prox + penalize-mask, the **entire elastic-net gamma solver**, the RW penalty/gradient (correct discrete Laplacian, within-base tying, saturated/pooled limits), the **CR1 sandwich / WLS / REML / AME / Rubin pooling** (cross-checked against statsmodels to machine precision), and the **bootstrap topic-alignment across replicates** (anchor re-resolution — the concern that would have been most damaging).

## Verification
- `pytest tests/test_content_trajectory.py tests/test_standard_errors.py` — 44 passed (new: `test_content_prior_var_guard`, `test_content_smooth_guard`, `test_bootstrap_all_failures_raise_not_silent_nan`, `test_cluster_robust_warns_on_too_few_clusters`)
- `pytest tests/test_stm*.py tests/test_content.py` — 190 passed
- `./scripts/preflight.sh` — fmt + clippy + #481 scan + table/stub sync all green

### DIFF
--- a/python/topica/content.py
+++ b/python/topica/content.py
@@ -28,6 +28,7 @@ topic-word distribution ``topic_word_at(level)`` at a few sentiment levels
 
 from __future__ import annotations
 
+import warnings
 from dataclasses import dataclass
 
 import numpy as np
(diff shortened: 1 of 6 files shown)


nealcaren:
Dual review: Claude (faithfulness) + **Gemini `3.1-pro-high`** (adversarial). Both clean, no blockers, all three fixes verified: (1) guards gated on `content.is_some() || content_time.is_some()` so plain STM is 100% unaffected, `ensure_finite_pos` for `content_prior_var` and `ensure_finite_nonneg` for `content_smooth` (0.0 stays legal); (2) bootstrap counts survivors, raises `RuntimeError` only on zero usable replicates, warns at >half, and `nanpercentile` keeps a partial-success CI valid; (3) cluster-robust `G<2`/`G≤p` conditions — the Claude reviewer noted `G≤p` rank-deficiency is *exact* (cluster scores sum to zero), non-spurious for G>p,G≥2. The core `ctm.rs` asserts are a backstop for faSTM's extendr front-end (topica users get the clean ValueError first). Composes cleanly with #536: `content_prior_var` sets the Laplace rate under l1, and guarding it >0 is what prevents `content_l1 = 1/0 = +inf` poisoning FISTA.

Rebased onto main (over merged #535 + #536); only conflict was the additive test file (both sets kept). Verified on the merged tree: 91 STM content/SE/L1 tests pass, preflight green. Non-blocking nits (dead `shape` param, cosmetic B=0 message) left for a future cleanup.

PR #535, opened 2026-07-25
fix(stm): order content_time periods numerically, not lexically (#534)

### LINKED ISSUE #534: BUG: STM content_time random-walk orders periods lexically, not numerically — smooths wrong chronological neighbors
From the adversarial review of the temporal content covariate. **Confirmed HIGH-severity correctness bug — silent wrong smoothing.**

## Defect
The saturated group index and, critically, the **temporal adjacency the random walk smooths over** are derived from a **lexical string sort** of the period labels.

`src/python/mod.rs` ~7459-7476:
```rust
let mut periods: Vec<String> = times_str.iter().cloned()
    .collect::<HashSet<_>>().into_iter().collect();
periods.sort();          // lexical sort on String
```
Period labels arrive as strings even when the user passes integers (`parse_groups` stringifies `Vec<i64>` via `to_string()`). So `content_time=[1,2,3,10,11,12]` becomes `["1","2","3","10","11","12"]` and `periods.sort()` orders them `1,10,11,12,2,3`. The RW in `topica-core/src/ctm.rs` (~659-687) then ties consecutive cells **in that lexical order** — pulling period **12 toward period 2** and **1 toward period 10**.

## Impact
Any integer-like period labels of non-uniform digit width are smoothed across chronologically non-adjacent cells: month numbers 1–12, congressional session numbers, day-of-study 1–30, quarter indices past 9, etc. **4-digit years (1990–2020) are accidentally safe** because lexical and numeric order coincide — which is exactly why the existing test (`tests/test_content_trajectory.py`, periods `[1990,2000,2010,2020]`) never catches it.

The bug is **fully silent**: no error, no warning. Worse, the Python reader `_parse_content_time_groups` (`python/topica/content.py` ~366-369) sorts periods **numerically** (`key=lambda p: float(p)`), so `content_trajectory(...).periods` shows a clean chronological axis and the user cannot tell the underlying surface was smoothed across the wrong adjacencies.

## Confirmed empirically
Fit STM with `content_time=[1,2,3,10,11,12]`, `content_smooth=20`, vs an unsmoothed (`content_smooth=0`) fit; measured how much each pair's gap shrinks:

| pair | kind | gap ratio (cs20/cs0) |
|---|---|---|
| (1,2) | numeric-adjacent | 1.21 (did NOT shrink) |
| (3,10) | numeric-adjacent | 0.95 (did NOT shrink) |
| (1,10) | lexical-adjacent | **0.84 (shrank)** |
| (12,2) | lexical-adjacent | **0.76 (shrank)** |

The RW is tying the wrong periods.

## Suggested fix
Sort periods **numerically when all labels parse as numbers** (mirroring `content.py`), else lexically; keep the Rust binding and the Python reader on identical ordering logic. Alternatively accept an explicit ordered `content_time_levels=` list. Add a test with non-uniform-width integer periods (e.g. `[1,2,3,10,11,12]`) so the years-only blind spot is closed.

## Note
The RW penalty/gradient itself, the within-base tying (no cross-base leak), the saturated/pooled limits, and the flatten-index round-trip are all verified correct — this is purely the period-ordering step. The `content_smooth` finiteness guard from the same review shipped in #531.

nealcaren:
Closes #534.

The `content_time` random-walk tied periods in the order of a **lexical** sort of the (stringified) period labels. Integer-like labels of non-uniform width were ordered wrong — `[1,2,3,10,11,12]` became `1,10,11,12,2,3`, so the RW smoothed the wrong chronological neighbors (period 12 pulled toward 2, period 1 toward 10). Four-digit years happened to be safe (lexical == numeric), which is why the existing tests missed it, and the bug was **silent**: the Python trajectory reader sorts numerically, so the read-back axis looked correct while the underlying fit had smoothed the wrong cells.

## Fix
Sort periods numerically when every label parses as a number, else fall back to lexical — matching the numeric ordering `content.py` already uses for the read layer.

## Verification
- New `test_content_time_periods_order_numerically_not_lexically`: fits with periods `[1,2,3,10,11,12]` and asserts the group axis orders `1,2,3,10,11,12` (was `1,10,11,12,2,3`).
- `pytest tests/test_content_trajectory.py` — 11 passed
- `./scripts/preflight.sh` — green

Found by the adversarial review of topica's faSTM-beyond-STM extensions. Note: faSTM's own R path orders periods via R `factor()`, which sorts numeric covariates numerically, so faSTM's documented numeric usage is unaffected (only string-labeled periods would hit the same class there).

### DIFF
--- a/src/python/mod.rs
+++ b/src/python/mod.rs
@@ -7454,15 +7454,30 @@ impl STM {
                         num_docs
                     )));
                 }
-                // Ordered periods: sorted unique labels (pass sortable labels, e.g.
-                // years or zero-padded strings, so the order is chronological).
+                // Ordered periods: unique labels sorted chronologically. When every
+                // label parses as a number, sort numerically so integer-like periods
+                // (1,2,3,10,11,12) order 1<2<...<10<... rather than lexically
+                // (1,10,11,12,2,3) — the latter would tie the wrong chronological
+                // neighbors in the random walk (#534). Falls back to lexical order for
+                // non-numeric labels. This matches the numeric ordering the Python
+                // trajectory reader (content.py) already uses.
                 let mut periods: Vec<String> = times_str
                     .iter()
                     .cloned()
                     .collect::<HashSet<_>>()
                     .into_iter()
                     .collect();
-                periods.sort();
+                if periods.iter().all(|p| p.trim().parse::<f64>().is_ok()) {
+                    periods.sort_by(|a, b| {
+                        let (na, nb) = (
+                            a.trim().parse::<f64>().unwrap(),
+                            b.trim().parse::<f64>().unwrap(),
+                        );
+                        na.total_cmp(&nb).then_with(|| a.cmp(b))
+                    });
+                } else {
+                    periods.sort();
+                }
                 let pindex: HashMap<&str, usize> = periods
                     .iter()
                     .enumerate()
--- a/tests/test_content_trajectory.py
+++ b/tests/test_content_trajectory.py
@@ -133,3 +133,18 @@ def test_to_frame(fit):
     df = tr.to_frame()
     assert list(df.columns) == ["word", "period", "estimate"]
     assert len(df) == 4  # 1 word x 4 periods
+
+
+def test_content_time_periods_order_numerically_not_lexically():
+    # Integer-like periods of non-uniform width must order chronologically
+    # (1,2,3,10,11,12), not lexically (1,10,11,12,2,3), or the random walk ties the
+    # wrong neighbors (#534). 4-digit years happen to be safe; these labels are the
+    # regression case.
+    docs, grp, per = [], [], []
+    for t in [1, 2, 3, 10, 11, 12]:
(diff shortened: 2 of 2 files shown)


nealcaren:
Dual review: Claude (faithfulness) + **Gemini `3.1-pro-high`** (adversarial). Both confirmed the fix end-to-end: sort order → `pindex` → the RW ties adjacent `pindex` positions, so numeric sorting genuinely re-wires which periods are chronological neighbors; `total_cmp` with a lexical tiebreak is a safe total order (no panic on the degenerate `nan`/`inf` parse), and the fit-time ordering now matches `content.py`'s numeric read-layer ordering (the silent fit/read mismatch this closes). Test is a genuine regression guard (fails under the old lexical sort). Non-blocking edge cases both noted: Rust vs Python parser asymmetry on pathological labels (`"1_000"`, `"nan"`) — strictly better than the pre-fix all-lexical state; worth a future validation follow-up, not a blocker. CI green.

PR #536, opened 2026-07-25
fix(stm): make content_prior="l1" a pure Laplace prior, not L1+L2 (#532)

### LINKED ISSUE #532: STM content_prior="l1" is elastic-net (L1+L2), not the documented pure Laplace; l1→l2 limit unreachable
From the adversarial review of the faSTM-beyond-STM extensions (content sparse prior). **Not a crash — a behavior-vs-documentation mismatch that needs a decision.**

## Finding
Under `content_prior="l1"`, the Gaussian (L2) prior is applied to **every** content coordinate in the smooth objective (`topica-core/src/ctm.rs` ~647-650: `value -= 0.5*inv_var*xi*xi` for all `i`), and the FISTA prox then adds L1 on top of the penalized `kappa_cov`/`kappa_interaction` blocks. So those deviation blocks carry **both** an L2 penalty (rate `inv_var = 1/content_prior_var`) and an L1 penalty (rate `content_l1 = 1/content_prior_var`) — an elastic net at a single coupled rate.

But the docstrings advertise a pure Laplace prior:
- core (`ctm.rs` ~574-577): "swaps the L2 solve for a FISTA solve with a sparse (Laplace/L1) prior ... keeping the topic baseline κ_topic on L2" (i.e. the deviations lose their L2)
- Python (`src/python/mod.rs` ~7500-7503): advertises a "rate 1/content_prior_var" Laplace.

Two consequences:
1. The effective regularization on the deviations differs from the docs (shrunk more than a pure-Laplace MAP), which matters for anyone tuning `content_prior_var` or citing "sparse Laplace / SAGE".
2. The advertised "l1 → l2 limit as rate → 0" is **unreachable**: `content_l1` and `inv_var` are the same knob, so sending the L1 rate to 0 (var → ∞) simultaneously kills the L2, landing on the unregularized MLE rather than the L2 solution at a fixed variance.

The identical rates suggest the residual L2 on the penalized blocks is an unintended leftover rather than a deliberately-balanced elastic net.

## Decision needed
- **(a) Make it truly Laplace:** drop the L2 term on the penalized `kappa_cov`/`kappa_interaction` blocks under `content_prior="l1"` (keep L2 only on `kappa_topic`/background), matching the docstrings and stm's sparse content model. OR
- **(b) Keep the elastic net:** relabel the docstrings/stub as elastic-net (L1+L2 at rate `1/content_prior_var`) and fix/remove the "l1→l2 limit" claim.

Also (related, lower priority): `content_prior_var` denotes a Gaussian *variance* under l2 but a Laplace *scale* (`1/var` rate) under l1 — same knob, different statistical quantity; worth a docstring note whichever way (a) / (b) goes.

## Verified correct (context)
The penalize-mask (only deviation blocks, never baseline/background), the soft-threshold prox, FISTA backtracking, and the l2 path (bit-exact) all check out; this is purely the residual-L2/labeling question.

_Source: adversarial review of faSTM extensions; guard fixes landed in #531._

nealcaren:
Closes #532.

## Problem
Under `content_prior="l1"`, the smooth objective applied the Gaussian (L2) prior to **every** content coordinate (`topica-core/src/ctm.rs`), and the FISTA prox then added L1 on the deviation blocks — so `kappa_cov` / `kappa_interaction` carried **both** penalties at the same coupled rate `1/content_prior_var`. That is a (coupled) elastic net, while the docstrings advertised a pure sparse Laplace / SAGE prior, and the "l1→l2 limit" was unreachable (one knob drove both rates).

## Fix
Skip the L2 term on the penalized deviation coordinates when `content_l1 > 0`, so the deviations carry the Laplace (L1) prior **only**; the baseline `kappa_topic` keeps its L2 either way. `content_prior="l2"` and the no-content path are untouched (bit-exact).

Because this is the **shared `topica-core` fitting core**, faSTM inherits the corrected prior automatically (its `content_prior="l1"` calls the same `fit_ctm`).

## Effect (verified)
On a strongly group-separated corpus at `content_prior_var=0.5`:
- l1 deviations: sparse (77/168 nonzero), mean |nonzero| ≈ **2.10** — survivors no longer ridge-shrunk
- l2 deviations: dense (168/168), mean |dev| ≈ **1.09**
- l1 baseline `kappa_topic`: dense (keeps L2) ✓

## Docs
Clarified that `content_prior_var` is the Gaussian **variance** under `"l2"` and the Laplace **scale** (rate `1/content_prior_var`) under `"l1"`.

## Verification
- New `test_l1_content_prior_is_pure_laplace_not_elastic_net` (sparsity, dense baseline, l1 survivors > l2 magnitudes)
- `cargo test --lib -p topica-core` — 43 passed
- `pytest` STM + content + gold + R-compare — 149 + 71 passed, no pinned-value regressions
- `./scripts/preflight.sh` — green

Found by the adversarial review of topica's faSTM-beyond-STM extensions.

### DIFF
--- a/src/python/mod.rs
+++ b/src/python/mod.rs
@@ -7253,13 +7253,17 @@ impl STM {
     /// random walk, the temporal generalization of `content`. `content_smooth`
     /// controls that random-walk penalty strength (``1/tau^2``); larger values tie
     /// adjacent periods more tightly. `content_prior` selects the prior on the
-    /// content (SAGE κ) deviation blocks: ``"l2"`` (default) is a Gaussian ridge that
-    /// keeps every `kappa_topic`, while ``"l1"`` puts a sparse Laplace prior (FISTA,
-    /// exact zeros) that recovers sparse content contrasts, matching R `stm`'s sparse
-    /// content model. `content_prior_var` is the L2 prior variance on those content
-    /// deviations (default ``0.5``); larger loosens regularization (more group-driven
-    /// contrast), smaller tightens it toward the shared baseline. The ``"l2"`` path
-    /// with `content_time=None` is bit-for-bit identical to the prior release.
+    /// content (SAGE κ) deviation blocks: ``"l2"`` (default) is a Gaussian ridge on
+    /// the deviations, while ``"l1"`` puts a *pure* sparse Laplace prior (FISTA, exact
+    /// zeros) on them — recovering sparse content contrasts, matching R `stm`'s sparse
+    /// content model. Under ``"l1"`` the deviation blocks carry the Laplace prior only
+    /// (not an additional ridge); the topic baseline `kappa_topic` keeps its L2 either
+    /// way. `content_prior_var` (default ``0.5``) sets the deviation-prior scale: under
+    /// ``"l2"`` it is the Gaussian prior variance, under ``"l1"`` the Laplace scale (the
+    /// L1 rate is ``1/content_prior_var``). Larger loosens regularization (more
+    /// group-driven contrast), smaller tightens it toward the shared baseline. The
+    /// ``"l2"`` path with `content_time=None` is bit-for-bit identical to the prior
+    /// release.
     /// `convergence_tol` is the relative-bound tolerance for EM early
     /// stopping — the run stops when the relative change in the variational evidence
     /// bound falls below it (the criterion R `stm` uses). `beta_init` is an optional
--- a/tests/test_stm_content.py
+++ b/tests/test_stm_content.py
@@ -548,3 +548,40 @@ class TestSTMContentRecomputeEtaCov:
         # Identical under the fix (both use the current fit's β); under the bug the
         # reused instance recomputes against the stale first-fit β and diverges.
         npt.assert_allclose(a, b, rtol=0, atol=1e-5)
+
+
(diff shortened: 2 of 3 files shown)


nealcaren:
Dual review: Claude (faithfulness) + **Gemini `3.1-pro-high`** (adversarial). Both confirmed the math: the L2 mask (`0..n_t`, the `kappa_topic` baseline) is the **exact complement** of the FISTA L1 penalize mask (`n_t..total`, the `kappa_cov`/`kappa_interaction` deviations) — no overlap (double-penalize) and no gap (un-penalized), so deviations carry Laplace-only while the baseline keeps L2. `content_l1 = 1/content_prior_var` only under `"l1"` (0.0 under `"l2"`), so the l2 and no-content paths are bit-exact (`content_l1 > 0` gates both the L2-skip and the FISTA vs L-BFGS route). The κ prior is outside the monitored ELBO in both old and new code, so there's no bound-vs-gradient inconsistency; value and grad drop L2 in lockstep. The sparsity test's discriminating assertion (l1 survivors > l2 magnitudes) genuinely fails under the old coupled elastic net. Both noted the same pre-existing nit — no `content_prior_var` positivity guard — which #531 closes. CI green.

PR #537, opened 2026-07-25
fix(estimate_effect): GLM robust SEs use HC0, not HC1 (#533)

### LINKED ISSUE #533: estimate_effect: GLM (logit/log) robust SEs apply the HC1 n/(n-p) factor; GLM/Papke-Wooldridge convention is HC0
From the adversarial review of `estimate_effect`. **Low severity — a small-sample convention divergence, not a correctness error.**

## Finding
`_sandwich` (`python/topica/stm.py` ~186) applies the finite-sample factor `n/max(n-p,1)` unconditionally. For a linear model this matches statsmodels HC1 exactly (verified). But for a **GLM** (fractional logit / quasi-Poisson), statsmodels treats HC0 and HC1 identically (no df factor), and the classic Papke–Wooldridge fractional-logit robust SE is HC0. So topica's GLM robust SE is a constant `sqrt(n/(n-p))` larger than the reference (observed 1.0127× at n=120, p=3; identical factor for logit and Poisson).

Defensible as a small-sample choice, but it is undocumented and diverges from the reference the docstring implicitly invokes (Papke & Wooldridge).

## Options
- Drop the `n/(n-p)` factor on the GLM (non-identity link) path so it matches HC0 / statsmodels / Papke–Wooldridge, OR
- Document that topica applies an HC1-style df correction to GLM robust SEs (and note the ~`sqrt(n/(n-p))` difference vs statsmodels).

## Verified correct (context)
β for both links matches statsmodels GLM to ~1e-12; the sandwich score/bread are analytically correct; the identity-link (OLS/WLS) HC1 and cluster-robust CR1 paths match statsmodels exactly. This is purely the GLM df-factor convention.

_Source: adversarial review of faSTM extensions; related guard fixes in #531._

nealcaren:
Closes #533.

## Problem
`_sandwich` (`python/topica/stm.py`) applied the finite-sample factor `n/max(n-p,1)` unconditionally. For the linear model that matches statsmodels **HC1** (verified). But for a **GLM** (fractional logit / quasi-Poisson), statsmodels treats HC0 and HC1 identically (no df factor), and the classic **Papke–Wooldridge** fractional-logit robust SE is **HC0**. So topica's GLM robust SE was a constant `sqrt(n/(n-p))` too large (≈1.0127× at n=120, p=3).

## Fix
Add an `hc1=` flag to `_sandwich` (default `True`) that gates the `n/(n-p)` factor on the heteroskedasticity-robust (non-cluster) meat, and pass `hc1=False` from the GLM branch of `_fit_one`. 

- Identity-link **HC1** and cluster-robust **CR1** paths are unchanged (both already matched statsmodels exactly).
- No change to point estimates; only the GLM non-cluster robust SE changes (removes the spurious inflation).
- Documented the convention in the `estimate_effect` docstring: GLM heteroskedasticity-robust SE is HC0, identity is HC1, cluster is CR1.

## Verification
- New `test_sandwich_hc1_flag_toggles_df_factor` (HC0 == bread·meat·bread, HC1 == HC0·n/(n-p)) and `test_fit_one_glm_reports_hc0_not_hc1` (the GLM path reports HC0, reconstructed from the same IRLS fit; guards against the HC1 regression).
- `pytest -k "stm or content or standard_error or effect or estimate"` — 1791 passed, 29 skipped, no pinned-value regressions.
- `./scripts/preflight.sh` — green.

### DIFF
--- a/python/topica/stm.py
+++ b/python/topica/stm.py
@@ -544,8 +544,9 @@ def estimate_effect(
       bounds where OLS can wander outside them (Papke & Wooldridge). Non-identity
       links report heteroskedasticity- or cluster-robust standard errors; the
       heteroskedasticity-robust GLM SE is HC0 (no ``n/(n-p)`` factor), matching
-      statsmodels' GLM sandwich and the Papke–Wooldridge convention, while the
-      identity link uses HC1 and the cluster path uses CR1.
+      statsmodels' GLM sandwich and the Papke–Wooldridge convention. The identity
+      link without ``cluster`` reports classical OLS standard errors; under
+      ``cluster`` both links use the CR1 cluster-robust sandwich.
     - ``weights`` — a length-``num_docs`` array of (survey) weights, or a column
       name in ``data``. Switches to weighted least squares: documents enter the
       regression in proportion to their weight, so a weighted sample (e.g. a
--- a/python/topica/stm.py
+++ b/python/topica/stm.py
@@ -173,18 +173,25 @@ def _link_inv(eta, link):
     return eta
 
 
-def _sandwich(X, bread, score_resid, groups, n, p, weights=None):
+def _sandwich(X, bread, score_resid, groups, n, p, weights=None, *, hc1=True):
     """Robust covariance ``bread · meat · bread``. With `groups` (a list of index
-    arrays) the cluster-robust CR1 estimator; otherwise heteroskedasticity-robust
-    HC1. `score_resid` is the estimating-equation residual (y−μ). With `weights`
+    arrays) the cluster-robust CR1 estimator; otherwise heteroskedasticity-robust.
+    `score_resid` is the estimating-equation residual (y−μ). With `weights`
     (survey weights), the score rows become ``w_i · X_i · resid_i`` and `bread` is
     the weighted ``(X'WX)^-1`` the caller supplies — matching faSTM's weighted
-    cluster-robust meat."""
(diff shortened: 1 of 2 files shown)


nealcaren:
Dual review: Claude (faithfulness) + **Gemini `3.1-pro-high`** (adversarial). Both confirmed the fix is statistically correct and faithful: the GLM robust covariance is the textbook quasi-MLE sandwich `(X'WX)^-1 · Σ ψψ' · (X'WX)^-1` with canonical-link score `ψ_i = X_i(y_i-μ_i)`, and HC0 (no `n/(n-p)` factor) is exactly the statsmodels-GLM / Papke–Wooldridge convention. Wiring verified: only two `_sandwich` call sites, GLM→`hc1=False`, identity default; the flag gates only the non-cluster factor so cluster CR1 is untouched (correct for GLM+cluster too); point estimates unchanged; Rubin-pooled draws get per-draw HC0 with no double-count. Both new tests are non-tautological (the `_fit_one` test would fail on a ~5% inflation if the wiring regressed).

One low-severity doc fix from the review: the docstring said 'identity link uses HC1', but identity+non-cluster actually reports classical OLS SEs (`_ols`) and identity+cluster uses CR1 — corrected in a follow-up commit. Expected downstream effect: non-clustered GLM SEs shrink by `sqrt((n-p)/n)` — the intended correction.

PR #538, opened 2026-07-25
fix(topicgpt): robustness pass for #509 (zero-doc topic, brace-safe prompts, parse drops, rare-topic pruning, open-model examples)

### LINKED ISSUE #509: TopicGPT review: 3-stage pipeline faithful to Pham et al.; zero-doc topic breaks topic_word invariant + NaN coherence, custom-prompt brace crash, silent parse drops, 'prune rare topics' unimplemented, non-open-model examples
## Review result

Two independent Claude reviewers — a faithfulness reviewer (3-stage protocol + prompts vs Pham et al. 2024 / chtmp223/topicGPT, open-model convention) and an adversarial reviewer (malformed-LLM-output parsing + orchestration edge cases, every finding empirically reproduced against the real module) — plus a maintainer trace. (Codex ghosted and Gemini's headless sandbox blocked the file reads, so the second reviewer is a second Claude subagent with an adversarial brief.) TopicGPT is a Python LLM-orchestration model — there is no math, so parsing robustness *is* the correctness surface.

**Bottom line: the 3-stage generate → refine → assign pipeline is faithful to Pham et al. — sequential generation with a running taxonomy, the `[level] Label: Description` parse, and code-side label verification against the hierarchy are all correct — but the LLM-output parsing has several confirmed robustness bugs and two documentation overclaims.** The document-text `str.format` injection hypothesis is genuinely **safe** (substituted values are not re-parsed).

### 1. Zero-document topic breaks the `topic_word` sum-to-1 invariant and yields NaN coherence (MEDIUM, reachable correctness bug)

`_class_tfidf` (`topicgpt.py:719-750`), `coherence` (`:827-834`). When the LLM discovers K topics but assigns documents to only some of them, a zero-doc topic gets an **all-zero** `topic_word` row. The `rowsum[rowsum==0]=1.0` guard (`:749`) prevents a div-by-zero, but the row still sums to 0 — so the documented "L1-normalized rows sum to 1" invariant (asserted by `test_topic_word_is_valid_descriptor`) is **false**, and `coherence()` returns **NaN** for those rows (violating `test_coherence_runs`'s finiteness expectation), with `top_words` empty. Empirically confirmed (3 topics, docs only on 1 → row sums `[1, 0, 0]`, coherence `[0.289, nan, nan]`). Reachable in normal use; the tests miss it because the fake backend spreads docs evenly across topics.

- [ ] Drop zero-doc topics before building `topic_word` (renumbering), or give them a uniform row and make `coherence` return a finite sentinel (e.g. skip/`0.0`) for empty rows; add a test with an uneven assignment.

### 2. Custom prompt templates with stray braces crash at fit time; validation gives false confidence (MEDIUM)

`_merge_prompts` (`topicgpt.py:204-212`), `.format()` at `:605/638/680/855`. The **document** path is safe. But the same `.format()` is applied to **user-supplied custom prompt templates**, and `_merge_prompts` validates only that the *required* `{field}` placeholders are present — it does not reject *extra* `{...}`. A custom prompt containing any stray brace (a pasted JSON or LaTeX example — a very common user action) is **accepted** by the validator, then crashes deep in `fit` with an opaque `KeyError`/`IndexError`:

```python
TopicGPT(backend=be, prompts={"generation":
    'Analyze {taxonomy} then {document}. Return JSON like {"topic":"name"}.'})
# _merge_prompts accepts it → fit() → KeyError: '"topic"'
# 'cost O(n^{2})' → IndexError: Replacement index 2 out of range
```

The field-presence check actively signals "your template is fine," so the late failure is doubly surprising.

- [ ] Validate custom templates at construction by trial-formatting with the exact stage kwargs (or scanning for unexpected replacement fields), so a bad template fails fast with a clear message.

### 3. Silent partial-drop of malformed / bolded topic lines in a mixed generation reply (LOW-MED)

`_TOPIC_LINE` (`topicgpt.py:306`), `_parse_topic_lines` (`:312-337`). The regex requires a literal `[n]` at line start. Markdown `**[1] Politics: ...**`, numbered `1.`, bulleted `- [1]`, and bracket-less lines are all silently dropped. The sharper bug: the JSON fallback only fires when *no* bracketed line matched (`if out: return out`, `:327`), so a **mixed** reply keeps the one clean line and **silently discards** the bolded ones (confirmed: `[1] Politics + **[1] Sports** + **[1] Tech**` → only Politics). An LLM that bolds every line yields zero topics → `fit` falls back to a single synthetic `topic_0`. Sub-finding: `[2]` subtopics are correctly skipped in generation (`:611`) but **not** in `_refine` (`:686`, no level guard).

- [ ] Strip surrounding markdown/bullets before matching (or match `[n]` anywhere in the line), and apply the JSON fallback when the bracketed parse is *suspiciously sparse*, not only when empty; add the level-1 guard to `_refine`.

### 4. Declined / fully-hallucinated assignments are silently forced onto topic 0 (LOW-MED, both reviewers)

`fit` (`topicgpt.py:642-643`, `if not chosen: chosen = [(0, "")]`), `transform` (`:858`). Hallucinated labels are correctly dropped (`name_to_id.get → None → continue`, so no garbage topic id can be injected — good). But a document whose assignments are *all* hallucinated or that the LLM declined ("None") is deterministically pinned to **topic 0** with an empty quote (confirmed), contaminating topic 0's document set, c-TF-IDF descriptor, and coherence with unrelated docs. Unlike BERTopic's `-1` outliers, there is no way to see which assignments were fallbacks. The reference re-prompts (an assignment-correction pass) rather than coercing.

- [ ] Record fallbacks as `topic_id = -1` / an `unassigned` flag on `Assignment` (or at minimum log a coerced-count to `stage_log`) instead of silently forcing topic 0.

### 5. "Prune rare topics" refinement is claimed but not implemented (LOW-MED, honesty)

Module docstring (`topicgpt.py:5-6`, "refines it by merging near-duplicates **and pruning rare topics**") and `fit` docstring (`:586`, "prune, cap at `max_topics`"). In chtmp223/topicGPT, generation tracks how many documents evoke each topic and refinement removes topics below a frequency threshold. topica's generation dedups into a `seen` dict (`:612-617`) and **never records per-topic frequency**, so the pruning signal is discarded; `_refine` (`:673-692`) does only the LLM merge, and the only "pruning" is `max_topics` *order*-truncation (`topics[:max_topics]`), not frequency-based removal. So a one-off spurious topic survives unless the LLM happens to merge it. The docstrings overclaim reference fidelity.

- [ ] Either track a per-topic document count in generation (increment on the dedup hit) and drop below a `min_count` in `_refine`, or correct the two docstrings to stop claiming rare-topic pruning (a fourth documented deviation).

### 6. Open-model repo convention not honored in examples/routed default (LOW, honesty)

`topicgpt.py:516` (`TopicGPT(model="gpt-4o-mini")`), `:542` (error-message example), routed through `labeling.py:80` (`llm_backend(model="gpt-4o-mini")`). TopicGPT itself has no hard default model (it requires `backend=`/`model=`, good), but every worked example names the proprietary `gpt-4o-mini`, against the repo's "LLM features: open models only" convention (showcase ollama/openrouter qwen).

- [ ] Change the docstring/example model strings to an open model (e.g. `model="ollama/qwen3"` or an openrouter qwen id) and add one sentence pointing at open models as the showcased default.

### 7. Secondary (PLAUSIBLE)

- **Refinement "None" false-positive** (`:682`): `reply.strip().lower().startswith("none")` fires on `"Nonetheless, here is the merged list: ..."` (confirmed True), discarding the parsed merge and silently keeping the pre-refinement topics. Use an exact/anchored match.
- **Quote is never verified against the document** (`:340-355`, stored `:651`): labels are code-verified, quotes are not — a hallucinated quote is stored verbatim and surfaced as if grounded, despite the "verbatim, do not invent" prompt. Optionally substring-check the quote (normalized) and blank/flag it if absent, or document that quote grounding is prompt-enforced only.
- **Quote-span regex over-captures** (`_QUOTE_SPAN`, `:309`): with parentheses in the reasoning, the end-anchored non-greedy body stretches to the final `)` (`... (see note) ("quote")` → captures `see note) ("quote"`). Display-only. Prefer the last parenthesized span.
- **Non-string backend coerced** (`:555`): `str(backend(prompt))` turns a `None` return into the literal `"None"` → parsed as a no-op → zero topics → synthetic `topic_0`, no error. A raising backend aborts fit mid-loop with no partial-progress recovery over a long/expensive run.

### Checked and CORRECT

**3-stage pipeline** in the right order and data flow (generation `:599-618` → refinement over the full taxonomy `:620-628` → assignment over the refined taxonomy `:630-654`). **Sequential generation with running taxonomy** — the taxonomy is re-rendered inside the per-doc loop (`:604`) and fed back into each prompt; new topics parsed and appended (`:610-617`) — the reference's incremental accumulation, not a fixed empty taxonomy. **Topic format parse** `_TOPIC_LINE` faithfully generalizes the reference regex; malformed lines skipped, empty labels guarded, tolerant JSON fallback, level-1-only in generation. **Label verification** against `name_to_id` (`:694-701`) — genuine code-side enforcement, better than prompt-only, and it prevents hallucinated topic ids from entering the matrices. **Hard/soft `{n_label}`** documented as deviation #3, default `hard` = one-hot matching the reference single-assignment. **Caching** per-instance sha256 on `(model-tag, prompt)` (`:546-558`), the three distinct stage templates prevent cross-intent collisions; determinism honestly labeled `llm-bounded`. **Prompt override** `_merge_prompts` rejects unknown keys and enforces required placeholders; `with_prompt` refuses post-fit changes. **Principled declines** — `estimate_effect`/`posterior_theta_samples`/`ensemble` raise clear "no theta posterior" errors, consistent with the BERTopic/Top2Vec cluster-style tier. **c-TF-IDF `topic_word`** documented throughout as a descriptor, not `P(w|topic)`. Stub/exports in sync.

### Bottom line

The pipeline structure is faithful, but for an LLM-output model the parsing robustness is the correctness surface, and it has real gaps: the zero-doc-topic invariant/NaN bug (#1) and the custom-prompt brace crash (#2) are the two worth fixing first; #3–#4 are silent-data-loss robustness; #5–#6 are documentation/convention overclaims.

Relevant surfaces: `python/topica/topicgpt.py`, `python/topica/labeling.py`, `tests/test_topicgpt.py`, `python/topica/__init__.pyi`.

nealcaren:
Closes #509.

A robustness and honesty pass on `TopicGPT` (pure-Python LLM orchestration; the parsing surface *is* the correctness surface). The faithful 3-stage generate -> refine -> assign core is untouched. Each item was re-verified against the current module before fixing.

## Implemented

**(a) Zero-doc topic breaks `topic_word` invariant + NaN coherence -- FIXED.** `_class_tfidf` gave a topic with no assigned documents an all-zero row: `rowsum[rowsum==0]=1.0` avoided a div-by-zero but left the row summing to 0, so the documented "rows sum to 1" descriptor invariant was false and `coherence` returned NaN (empty `top_words`). Zero-doc rows now get a uniform `1/V` row, restoring the invariant, and `coherence` `nan_to_num`-guards any degenerate row to a finite `0.0`.

**(b) Custom prompt with literal braces crashes at fit time -- FIXED.** The document path was already safe (substituted values are not re-parsed), but a user-supplied custom template with a stray brace (pasted JSON, LaTeX `O(n^{2})`, a stray placeholder) passed the field-presence check and then crashed deep in `fit` with an opaque `KeyError`/`IndexError`. `_merge_prompts` now trial-formats each overridden template with its exact stage kwargs, failing fast at construction with a message that says to double the braces.

**(c) Silent parse drops -- SURFACED (and the root cause fixed).** Two parts:
- Root cause of the mixed-reply loss (#3): markdown/bullet decoration (bolded, `- [1] ...`, `1. [1] ...`) is now stripped before matching, so decorated lines are parsed instead of silently discarded.
- Any remaining non-noise line the parser cannot read (a bare `None` reply is not counted) is collected into the new `parse_drops` attribute, logged to `stage_log` as `("parse_drops", n)`, and reported via a `warnings.warn` once per fit.
- Also anchored the `_refine` "None" check so `"Nonetheless, ..."` is no longer mistaken for a bare `"None"` (#7), and added the level-1 guard to `_refine` that generation already had.

**(d) "Prune rare topics" claimed but unimplemented -- IMPLEMENTED.** Generation now tracks a per-topic document count (incrementing on the dedup hit), and the new `min_topic_count` constructor param (default `1`, a no-op) prunes topics evoked by fewer than that many documents before refinement -- the reference's frequency-based rare-topic removal, making the existing docstring claim true rather than removing it. Pruning never removes the last topic. Exposed in `settings`; no `.pyi` change needed (`TopicGPT` is a pure-Python re-export).

**(e) Non-open example models -- SWAPPED.** The `with_prompt` docstring example, the no-backend error message, the class docstring, and the LLM guide now showcase an open model (`ollama/qwen3`) with a sentence pointing at open models as the showcased default, per the repo convention.

## Skipped / not changed (with reasons)

- **#4 (declined/hallucinated assignments coerced onto topic 0):** left as-is. The forced-to-0 fallback and coerced count are now indirectly visible through `parse_drops`/`stage_log`; a full `-1`/`unassigned` sentinel changes the `doc_topic`/`Assignment` contract and is a larger design change better scoped separately.
- **#7 quote-not-verified / quote-span over-capture / non-string backend coercion:** display-only or prompt-enforced concerns, out of scope for this robustness pass (the "Nonetheless" false-positive from #7 *was* fixed).
- **`labeling.py` `llm_backend(model="gpt-4o-mini")` default:** a shared functional default for the labeling adapter, not a TopicGPT example; changing it is a separate concern.
- Verified the 3-stage pipeline, label verification, caching, and declines are already faithful -- left untouched.

## Verification

- `pytest tests/test_topicgpt.py -q` -- **41 passed** (30 existing unchanged + 11 new: zero-doc health, brace-safe/positional/escaped prompts, markdown-line parsing, parse-drop surfacing + warning, anchored `None`, prune behavior, `min_topic_count` guards).
- `./scripts/preflight.sh` -- **all checks passed** (fmt + clippy + #481 guard + model tables + stub sync; 78 stub-sync tests green). Also ran as the pre-push hook.
- `pytest tests/ -k "conformance or settings or manifest or naming"` -- **742 passed, 21 skipped**.
- `mkdocs build --strict` -- clean.

### DIFF
--- a/docs/guides/llm.md
+++ b/docs/guides/llm.md
@@ -17,7 +17,8 @@ without topica taking an API dependency.
 ```python
 import topica
 
-# Bring a model: a callable, or model="gpt-4o-mini" via the topica[llm] adapter.
+# Bring a model: a callable, or model="ollama/qwen3" via the topica[llm] adapter.
+# topica showcases open-source models; any callable str -> str backend works.
 model = topica.TopicGPT(backend=my_callable, assignment="hard")
 model.fit(docs)                       # docs: a Corpus, raw strings, or token lists
 
@@ -33,6 +34,12 @@ model.transform(new_docs)             # assign held-out docs to the taxonomy
 `model.hierarchy`. `sample=` restricts the generation stage to the first *n*
 documents as a cost control (assignment still covers every document), and
 `model.estimated_calls(docs)` reports the backend-call budget before you run.
+`min_topic_count=` prunes topics evoked by fewer than *n* documents before
+refinement (the reference's rare-topic removal; the default of 1 keeps every
+topic), and `max_topics=` caps the count carried past refinement. When a backend
+reply does not match the requested format, the unparsed lines are recorded in
+`model.parse_drops` and a warning reports the count, so a malformed backend is
+not silent.
 
 ### What composes, and the caveats
 
--- a/python/topica/topicgpt.py
+++ b/python/topica/topicgpt.py
@@ -50,6 +50,7 @@ import hashlib
 import json
 import pickle
 import re
+import warnings
 from collections import Counter
 from dataclasses import dataclass
 from typing import Callable, Optional, Sequence
@@ -209,6 +210,23 @@ def _merge_prompts(overrides: Optional[dict]) -> dict[str, str]:
                 f"{missing} (expected placeholders: "
                 f"{['{' + f + '}' for f in PROMPT_FIELDS[stage]]})"
             )
+        # Fail fast on a template that ``str.format`` cannot fill. The
+        # orchestration substitutes only the stage's required fields via
+        # ``str.format``, so any *other* brace (a pasted JSON snippet like
+        # ``{"topic": "x"}``, a LaTeX/exponent ``O(n^{2})``, or a stray
+        # ``{placeholder}``) makes the real fit crash deep in ``fit`` with an
+        # opaque ``KeyError``/``IndexError``. Trial-format here with the exact
+        # stage kwargs so the failure surfaces at construction with a clear
+        # message and a fix (escape literal braces as ``{{`` / ``}}``).
+        try:
+            template.format(**{f: "" for f in PROMPT_FIELDS[stage]})
+        except (KeyError, IndexError, ValueError) as exc:
+            raise ValueError(
+                f"custom {stage!r} prompt is not brace-safe: str.format() raised "
+                f"{type(exc).__name__}: {exc}. The orchestration fills only "
+                f"{list(PROMPT_FIELDS[stage])} via str.format(), so escape every "
+                "other literal brace by doubling it ({{ and }})."
+            ) from exc
         merged[stage] = template
     return merged
 
@@ -308,24 +326,51 @@ _TOPIC_LINE = re.compile(r"^\s*\[(\d+)\]\s*([^:\]]+?)\s*:\s*(.*)$")
 # e.g. ``[1] Trade: reasoning ("...the quote...")``.
 _QUOTE_SPAN = re.compile(r"\(\s*[\"“‘]?(.*?)[\"”’]?\s*\)\s*$", re.DOTALL)
 
+# Leading list/markdown decoration an LLM often wraps a topic line in:
+# ``- ``, ``* ``, ``1. ``, ``1) ``, or a leading ``**`` bold marker. We strip
+# these before matching so a decorated ``**[1] Sports: ...**`` is not silently
+# dropped (github issue #509 finding #3).
+_LINE_DECORATION = re.compile(r"^\s*(?:[-*+•]\s+|\d+[.)]\s+)?\*{0,2}\s*")
 
-def _parse_topic_lines(text: str) -> list[tuple[int, str, str]]:
+
+def _strip_decorations(line: str) -> str:
+    """Remove leading bullet/number markers and surrounding ``**`` bold marks so a
+    markdown-wrapped topic line still matches :data:`_TOPIC_LINE`."""
+    s = _LINE_DECORATION.sub("", line)
+    s = re.sub(r"\*{1,2}\s*$", "", s)  # trailing bold close
+    return s.strip()
+
+
+def _is_noise_line(raw: str) -> bool:
+    """A blank line or a bare ``None`` reply (the reference's "no change / no
+    topic" signal) is expected, not a dropped topic line."""
+    s = raw.strip().rstrip(".")
+    return not s or s.lower() == "none"
+
+
+def _parse_topic_lines(text: str) -> tuple[list[tuple[int, str, str]], list[str]]:
     """Parse the reference ``[level] Label: Description`` format into
-    ``(level, name, description)`` triples. A bare ``None`` reply (the reference's
-    "no change / no topic" signal) yields an empty list. Falls back to a tolerant
-    JSON parse if no bracketed line is present (a backend that ignored the format).
+    ``(level, name, description)`` triples, plus the list of non-noise lines that
+    did not parse (so the caller can surface silent drops, issue #509 finding #3).
+
+    A bare ``None`` reply yields no triples and no drops. Markdown/bullet
+    decoration is stripped before matching. Falls back to a tolerant JSON parse if
+    no bracketed line is present (a backend that ignored the format); a reply that
+    parses cleanly as JSON reports no drops.
     """
     out: list[tuple[int, str, str]] = []
+    dropped: list[str] = []
     for line in str(text or "").splitlines():
-        m = _TOPIC_LINE.match(line)
-        if not m:
-            continue
-        name = m.group(2).strip()
+        raw = line.strip()
+        m = _TOPIC_LINE.match(_strip_decorations(line))
+        name = m.group(2).strip() if m else ""
         if not name:
+            if not _is_noise_line(raw):
+                dropped.append(raw)
             continue
         out.append((int(m.group(1)), name, m.group(3).strip()))
     if out:
-        return out
+        return out, dropped
     # Tolerant fallback: {"topics": [{topic, description}, ...]} or {topic, description}.
     obj = _extract_json(text)
     if isinstance(obj, dict) and isinstance(obj.get("topics"), list):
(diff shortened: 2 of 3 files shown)


nealcaren:
Dual review: Claude (faithfulness) + **Gemini `3.6-flash-high`** (adversarial). Both confirmed the five fixes are correct: (a) zero-doc topics get a uniform `1/V` row restoring the rows-sum-to-1 invariant + `coherence` `nan_to_num`-guard (only maps NaN/inf, real values pass through); (b) `_merge_prompts` trial-formats each override with its exact stage kwargs — no false positive on a legitimately-braced template, fails fast on a stray brace; (c) markdown stripping + anchored `None` check (rejects 'Nonetheless') + `parse_drops` surfacing; (d) `min_topic_count` pruning with correct `>=`/no-op-at-1/never-empty guards; (e) open-model examples.

One real semantic gap both flagged in the new `min_topic_count` feature: it counted topic **mentions** (a topic listed twice in one reply counted as 2) rather than **documents**. **Amended**: dedup topic keys per document (a `doc_keys` set) so it's a true document-frequency threshold, plus a regression test. Other low-severity notes (chatty-backend over-warn; `_refine` skipping on a leading blank line) left as non-blocking follow-ups.

PR #540, opened 2026-07-25
fix(sentence-ideal): monotone EM objective under non-unit x_prior_variance (#530)

### LINKED ISSUE #530: IdealPointSentenceTM: reconcile unit-variance identification with x_prior_variance for a monotone EM objective
Follow-up from #499 / PR #523 review.

## Problem

`IdealPointSentenceTM` standardizes author positions to **unit variance** every EM sweep for identification (`standardize_positions`), but the positions carry a Gaussian prior with variance `x_prior_variance`. When those scales differ, the identification step fights the prior, and the reported incomplete-data log-likelihood is **not monotone** across iterations.

Reproduced (early-stopping disabled, imbalanced authors), worst per-step change in `fit_history`:

| x_prior_variance | worst step |
|---|---|
| 0.05 | −9.06 |
| 0.2  | −0.81 |
| 0.5 / 1.0 / 2.0 | monotone |
| 5.0 | −1.7e-5 |

At the default `x_prior_variance=1.0` everything is monotone, which is why it went unnoticed.

## Why PR #523 didn't fix this

#523 correctly makes the reported value a properly normalized incomplete-data log-likelihood and stops claiming monotonicity. Reporting the *penalized* (MAP) objective instead does **not** restore monotonicity, because the unit-variance standardization pins the penalty term `(1/2·x_prior_variance)·Σ‖x_a‖²` to the constant `a_n·dd` at every reported point, so the penalized objective reduces to `data_LL + const`.

## Options for a real fix (a modeling decision)

1. Standardize positions to the **prior's** scale (`sqrt(x_prior_variance)`) instead of unit variance, so identification and prior agree; then report the penalized objective (genuinely monotone), evaluated after the full M-step.
2. Drop the per-iteration standardization and identify only once post-fit.
3. Make `x_prior_variance` a fixed unit and remove it as a knob.

All change fitted `x` for non-unit priors, so this needs re-validation of the (experimental) model. Not urgent — the model is experimental and #523 makes the reported quantity honest.

nealcaren:
Closes #530.

## Problem

`IdealPointSentenceTM` re-standardized author positions to unit sample variance every EM sweep for identification. That rescale is lossless for the data likelihood, but it (a) changes the positions the `(mu, V)` M-step uses as regressors and (b) changes the position prior penalty `sum_a ||x_a||^2 / (2 x_prior_variance)`. Because the position M-step is MAP, the objective EM actually maximizes is the *penalized* incomplete-data log-likelihood, and the mid-loop standardization broke its monotonicity whenever `x_prior_variance` differed from the unit identification scale (issue's reproduction: worst step -9.06 at `x_prior_variance=0.05`). At the default `x_prior_variance=1.0` unit variance already agrees with the prior, which is why it went unnoticed.

## Fix

Issue Option 2: identify once, after the loop. With the Gaussian position prior the scale of `x` is identified during fitting, so `standardize_positions` + anchor sign orientation now run a single time *after* the EM loop instead of every sweep. `ll_history` / `fit_history` now reports the penalized MAP-EM objective (data log-likelihood minus the position prior penalty), computed at the same positions the M-step optimized. That is the quantity EM maximizes here, and it is monotone non-decreasing (up to the tiny ridge in the least-squares solves). Convergence is judged on that objective.

`log_likelihood` is unchanged: it stays the properly normalized *unpenalized* incomplete-data log-density from #499/#523. Docstrings (Rust module + struct field, Python `fit_history`) now say plainly that `x_prior_variance` sets shrinkage strength during fitting, not the scale of the returned positions (which are always unit variance).

## Why this over the alternatives

Keeping unit-variance standardization inside the loop and merely subtracting the right penalty cannot yield a monotone objective: the standardization perturbs the M-step's `(mu, V)` regressors, so the penalized objective genuinely dips. This was confirmed empirically: a first attempt that kept in-loop standardization and paired the pre-standardization penalty still showed a -0.010 step at `x_prior_variance=0.05`. Making the objective monotone therefore requires removing the in-loop rescale. This does change the fitted positions for non-unit priors (as the issue notes), so the change was re-validated below.

## Verified already done (not touched)

- #499/#523: `log_likelihood` is the normalized incomplete-data log-density and lag-free. Left as is; the Rust `reported_ll_matches_recompute` and Python `test_reported_ll_is_a_normalized_density` still guard it and pass.
- The closed-form E-step and the `(mu, V)` / `x` / `sigma^2` M-step solves are unchanged.

## Verification

- `cargo test --lib sentence_ideal` - 4 passed (recovery, position-SE-shrinks, reported-ll-matches-recompute, and the new `fit_history_objective_is_monotone`).
- `pytest tests/test_sentence_ideal.py -q` - 12 passed (including the new parametrized `test_fit_history_objective_is_monotone` over `x_prior_variance` in {0.05, 0.2, 5.0}).
- The monotonicity tests use `x_prior_variance` in {0.05, 0.2, 0.5, 1.0, 5.0} (Rust) and disable early stopping, so the full trace is checked; the position-recovery test (`r > 0.8`) re-validates the changed fit for the experimental model.
- `./scripts/preflight.sh` - all checks passed (fmt, clippy `-D warnings`, #481 guard scan, model-table sync, `.pyi` stub sync). The `.pyi` needed no change: the `fit_history` signature is unchanged (only its docstring semantics changed).

### DIFF
--- a/src/python/sentence_ideal.rs
+++ b/src/python/sentence_ideal.rs
@@ -301,13 +301,15 @@ impl IdealPointSentenceTM {
     fn log_likelihood(&self) -> PyResult<f64> {
         Ok(self.fitted_model()?.log_likelihood)
     }
-    /// Per-iteration `(iter, log_likelihood)` trace of the data log-likelihood at
-    /// each E-step, for convergence monitoring. Note this data log-likelihood is
-    /// **not** the objective EM maximizes here (the position M-step is MAP and the
-    /// positions are re-standardized to unit scale each sweep), so it is not
-    /// guaranteed monotone -- with `x_prior_variance` far from the unit
-    /// identification scale it can briefly decrease. Use the relative change, not
-    /// strict monotonicity, to judge convergence.
+    /// Per-iteration `(iter, objective)` trace of the penalized incomplete-data
+    /// objective that MAP-EM maximizes: the data log-likelihood minus the position
+    /// prior penalty `sum_a ||x_a||^2 / (2 x_prior_variance)`. It is monotone
+    /// non-decreasing across sweeps (up to the tiny ridge in the least-squares
+    /// solves), so convergence can be judged from it directly. The bare data
+    /// log-likelihood is not the EM objective here (the position M-step is MAP) and
+    /// is not monotone when `x_prior_variance` differs from the unit identification
+    /// scale; that is why the penalty is subtracted. The reported `log_likelihood`
+    /// remains the (unpenalized) incomplete-data log-likelihood of the fitted model.
     #[getter]
     fn fit_history(&self) -> PyResult<Vec<(usize, f64)>> {
         Ok(self
--- a/src/python/sentence_ideal.rs
+++ b/src/python/sentence_ideal.rs
@@ -45,6 +45,13 @@ struct SentenceIdealState {
     ll_history: Option<Vec<f64>>,
     converged: Option<bool>,
     iters_run: Option<usize>,
+    // Per-dimension identification scale for a correctly scaled `position_se`.
+    // Positional bincode makes serde(default) inert for a genuine old save (it
+    // fails to deserialize), but the default keeps same-version round-trips and a
+    // self-describing reader working; an empty scale makes `position_se` fall back
+    // to sd = 1 (the pre-fix behavior).
+    #[serde(default)]
+    std_scale: Option<Vec<f64>>,
 }
 
 impl IdealPointSentenceTM {
@@ -373,6 +380,7 @@ impl IdealPointSentenceTM {
                 ll_history: m.map(|m| m.ll_history.clone()),
                 converged: m.map(|m| m.converged),
                 iters_run: m.map(|m| m.iters_run),
+                std_scale: m.map(|m| m.std_scale.clone()),
             },
         )
     }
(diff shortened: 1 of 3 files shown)


nealcaren:
Dual review: Claude (faithfulness) + **Gemini `3.1-pro-high`** (adversarial), plus my own Monte-Carlo verification.

The **monotone-EM core fix is confirmed correct** by both reviewers and my Rust-test run: J = data-LL − Σ‖x‖²/(2·x_prior_variance) is the true MAP-EM objective (x M-step is the exact MAP update whose ridge = the penalty), it's evaluated at the M-step positions, and the deferred once-after-loop identification is lossless. The new monotonicity test fails under the old in-loop standardization.

Gemini also caught a **real `position_se` scaling bug** for non-unit priors that Claude's pass didn't cover. I quantified it by Monte Carlo (data-poor, 3 obs/author): reported/true SE ratio decayed 0.78→0.46 as x_prior_variance went 1.0→0.05 — a >2x understatement, worsening with prior strength. Root cause: with positions now fit in unstandardized space and standardized once (x_std = x_fit/sd), the prior block of Cov(x_std) is diag(sd²)/xvar, but `position_se` used diag(1)/xvar. **Fixed**: `standardize_positions` returns the per-dim scale, stored as `std_scale` (serde-default; old saves fall back to sd=1), and `position_se` adds `inv_xvar·sd_j²`. Re-running the Monte Carlo, the ratio is now **flat across priors (0.82/0.83/0.84)** — the understatement is gone. Added a Rust regression test recomputing the sd²-scaled Hessian. Both reviewers' doc/comment notes were already addressed in the PR (verified).

PR #541, opened 2026-07-25
fix(tbip): honest parity claim, sign-id stub note, clamp-region prior gradient (#506)

### LINKED ISSUE #506: TBIP review: full-VI faithful to keyonvafa/tbip, LogNormal-Gamma KL + gradients FD-verified; parity harness never runs its claimed PyTorch reference, undocumented sign identifiability, clamp-region prior gradient
## Review result

Two independent Claude reviewers — a faithfulness reviewer (generative model + ELBO/KL + full-VI-not-MAP vs the paper and `keyonvafa/tbip`, defaults/priors confirmed by fetching `tbip.py`) and an adversarial reviewer (hand-deriving the reparameterized gradients, the LogNormal-Gamma KL, and auditing whether the FD test fixes the MC sample) — plus a maintainer trace. (Codex ghosted and Gemini's headless sandbox blocked the file reads, so the second reviewer is a second Claude subagent with an adversarial brief.)

**Bottom line: the port is mathematically faithful and correct — no correctness bug.** TBIP is a Poisson factorization with full mean-field VI (LogNormal factors for `theta`/`beta`, Normal for `eta`/`x`, reparameterized SVI + Adam). Both reviewers confirmed the generative model, the genuinely-full VI (not a MAP shortcut — every factor learns both a location and a scale), the trickiest term (the LogNormal-vs-Gamma ELBO term and its reparam gradient), the analytic Gaussian KL, the Poisson likelihood gradients, and the SVI global-vs-local scaling. Critically, the FD harness **properly fixes the MC sample** (`eps` drawn once, `elbo_and_grads` draws no internal RNG, both `±h` evaluations reuse it) and covers all 8 variational buffers over the full ELBO at 1e-3 — so the in-range gradients are genuinely verified. The findings are honesty/documentation overclaims plus one benign clamp-region gradient inconsistency.

### 1. The parity harness claims a "PyTorch reference" it never runs (MEDIUM, honesty)

`parity/tbip_parity.py:1-11` docstring: "against an independent PyTorch reference of the SAME model + the SAME mean-field VI inference" (echoed in `src/tbip.rs:22-24` and the stub `_topica.pyi:4913-4915` — "validated against a PyTorch reference (scratchpad/tbip_vi.py)"). But `main()` (`tbip_parity.py:56-66`) never builds or fits any torch model — it samples from the generative model with numpy, fits `topica.TBIP`, and checks planted-position recovery `|r| > 0.85`. `import torch` (line 18) is the *only* torch use, yet it gates the whole script: without torch the parity **skips entirely** even though the computation needs only numpy+topica. And `scratchpad/tbip_vi.py` does not exist (scratchpad is transient/gitignored), so there is no committed PyTorch cross-check anywhere.

- [ ] Either commit the actual PyTorch reference and compare positions against it, or soften all three docstrings to "validated by planted-position recovery from the generative model" and drop the dead `import torch` gate so the numpy-only test runs in CI. (This is the "offline gold / honest parity" class from the CI-gold memory.)

### 2. Ideal-point sign non-identifiability is undocumented (LOW-MED, honesty — mirrors Wordfish #505 / IdealPointTM #498)

`ideal_points()` returns `mu_x` verbatim (`src/tbip.rs:134-136`, binding `src/python/tbip.rs:305-309`). TBIP is invariant under `x → -x, eta → -eta` (both N(0,1) priors and the `exp(x·eta)` likelihood are unchanged), so the axis sign is arbitrary and seed-determined; the scale is only soft-pinned by the N(0,1) priors. The tests and parity harness all take `abs(corrcoef)` or flip the sign (`test_tbip.py:79`, `tbip_parity.py:60-61`), quietly confirming the axis is unoriented. `ideal_points` for the same corpus can come out sign-flipped across seeds — users reading left/right onto the sign will be misled.

- [ ] Document in the `ideal_points` docstring/stub that the axis is identified only up to sign (and softly in scale), or add an optional `anchors=`/orient argument like Wordfish.

### 3. "Defaults mirror the reference" is an overclaim (LOW, honesty)

`src/tbip.rs:189,199-208`. Verified against `keyonvafa/tbip/tbip.py`: reference defaults are `learning_rate=0.01`, `batch_size=1024`, `max_steps=1000000`, `num_topics=50`. topica uses `learning_rate=0.05`, `batch_size=512`, `iters=7000` (`src/python/tbip.rs:101-102`), plus a 50%/80% LR-halving schedule (`tbip.rs:652-666`) and an empirical-word-frequency init for `mu_beta` (`:233-238`) where the reference initializes objective topics from a pre-fit (NMF/Poisson factorization). Only the Gamma(0.3,0.3) and N(0,1) priors truly match. These are legitimate engineering choices for a shorter default run, but not the reference's numbers.

- [ ] Reword to "priors match the reference; the optimization schedule (lr, batch, steps, LR decay, init) is tuned for topica's shorter default run."

### 4. θ/β prior-gradient ignores the clamp the forward ELBO honors (LOW, real but harmless)

`src/tbip.rs:500-511` (θ), `528-540` (β). The *likelihood* reparam gradient correctly zeros `grad_mu`/`grad_rs` when `theta_unclamped`/`beta_unclamped` is false (matching the forward's `exp(clamp(u))`), but the *prior-term* gradient `dt_dmu = a - b·z`, `dt_drs = (a-b·z)·eps·lg + lg/sig` is added **unconditionally** (`:508-509`, no clamp guard). Meanwhile the forward ELBO's prior term uses `logz = z.max(1e-300).ln() = clamp(u)`, which *is* frozen when saturated. So in the clamped tail the pair is inconsistent: for `u>6` the true `d(ELBO_prior)/dmu = 0` but the code applies `svi·(a - b·exp(6))`. Practically negligible — it only fires in the saturated tail, where the spurious `a - b·z` term is strongly negative (a benign restoring force pushing the mean back in-range) — and the FD test never reaches the clamp region (`mu_theta≈-1`, far from the `[-10,6]` bound). But strictly, the θ/β prior gradient is not the exact adjoint of the coded ELBO in the clamp corner, contradicting the docstring's "FD-checked hand-coded adjoint."

- [ ] Gate `dt_dmu`/`dt_drs` (and the entropy `lg/sig` term) behind the same `*_unclamped` flag as the likelihood gradient, and add an FD case that perturbs into the clamp region.

### 5. Minor / cosmetic

- **Poisson `-log y!` dropped from the reported ELBO** (`tbip.rs:425`): correct for gradients/optimization, but `fit_history` ELBO values are offset by a fixed constant vs the reference's `Poisson.log_prob`. Harmless for convergence monitoring; one-line comment.
- **FD coverage gap** (adversarial): the FD test runs at `batch_size = D = 6` so `svi = D/B = 1` — the minibatch D/B scaling factor is never exercised by FD and is verified only by inspection (it is correct: recon ×D/B, global β/η/x KL ×1). A `batch < D` FD case would close it.
- **Minibatch sampling** (PLAUSIBLE): topica draws B indices with replacement each step ("mirroring torch.randint"); the reference uses `tf.data` shuffle→repeat→batch (without-replacement epoching). Both are valid SVI subsampling; differ only in optimization noise.

### Checked and CORRECT (traced + hand-derived + FD-verified, source-confirmed)

Generative model `rate = RATE_FLOOR + Σ_k θ_dk·β_kv·exp(x_{a_d}·η_kv)`, Gamma(0.3,0.3) on θ/β, N(0,1) on η/x — matches paper Eq. 3 and `tbip.py`. Genuinely **full mean-field VI** (both `mu_*` and `rs_*=softplus→σ` learned for every factor; LogNormal `θ=exp(μ+σε)`, Normal `η=μ+σε`). **LogNormal-vs-Gamma term** `logp=(a-1)logz - b·z + a·ln b - lnΓ(a)`, `logq=-logz - lnσ - ½ln2π - ½ε²`, `E[-logq]` = exact LogNormal entropy `μ + ½ln(2πe σ²)` (m-dependent, correctly carried); reparam grads `dT/dμ=a-b·z`, `dT/drs=(a-b·z)·ε·lg + lg/σ` — single-sample MC used consistently forward and backward. **Analytic Gaussian −KL** `-(½(σ²+μ²-1) - lnσ)` with grads `-μ`, `-(σ-1/σ)·lg`. **Likelihood gradients** `g=counts/rate-1`; `dl_dβ=g·c/β`, `dl_dθ=Σ g·c/θ`, `dl_dη=g·c·x`, `dl_dx=Σ g·c·η`. **SVI scaling** recon and local-θ ×`D/B`, global β/η/x KL ×1 — unbiased full-data ELBO estimate, the DETM/InfoCTM bug absent. **FD harness** fixes eps, covers all 8 buffers over the full ELBO at rel-err 1e-3. **Adam** ascent sign correct. **Determinism** (single RNG, fixed 16-chunk in-order reduction, bit-reproducible under rayon; `test_determinism`). **Robustness** RATE_FLOOR 1e-7, σ floor, exp clamps prevent LogNormal overflow, `finite_pos` guards on `a_gamma`/`b_gamma`/`learning_rate` + `iters≥1`/`num_topics≥2` (#481 clean). **Outputs** `ideal_points=mu_x`, `position_se=σ_x` (honestly caveated as mean-field-understating), `topic_word`, `ideological_topics=mu_eta`, `doc_topic`. **Save/load** tag 35 round-trips all 8 buffers; `group=` default = per-document.

### Bottom line

The generative model, full-VI inference, LogNormal-Gamma KL, and SVI scaling are all correct and FD-verified. The actionable items are the honesty overclaims — the parity harness that never runs its advertised PyTorch reference (#1), the undocumented sign non-identifiability (#2), and the "defaults mirror the reference" claim (#3) — plus the clamp-region prior-gradient consistency fix (#4).

Relevant surfaces: `src/tbip.rs`, `src/python/tbip.rs`, `tests/test_tbip.py`, `parity/tbip_parity.py`, `python/topica/_topica.pyi:4900-4960`.

nealcaren:
Closes #506.

Three follow-ups from the TBIP faithfulness review. The core model, full-VI inference, LogNormal-Gamma KL, and SVI scaling were confirmed correct and FD-verified in the review; these are honesty/documentation fixes plus one clamp-corner gradient fix. No verified in-range gradient math is touched.

## (a) Parity harness honesty — implemented

`parity/tbip_parity.py` claimed to validate "against an independent PyTorch reference of the SAME model + the SAME mean-field VI inference," but `main()` never built or fit any torch model — it samples from the generative model with numpy, fits `topica.TBIP`, and checks planted-position recovery. The only torch use was an `import torch` that gated the whole script, so without torch the (numpy-only) check skipped entirely, and the advertised `scratchpad/tbip_vi.py` reference does not exist.

- Relabeled the harness docstring as a self-consistency (planted-position recovery) check, stating why the official TF1.14/TFP0.7 code is not run.
- Dropped the dead `import torch` gate so the check runs offline in CI. It now runs and passes: `Pearson |r| = 0.996`.
- Mirrored the wording in the `src/tbip.rs` module docstring and the `.pyi` class docstring (which said "a PyTorch reference"). Did not fabricate a reference run.

## (b) Sign identifiability — implemented (stub sync)

The `ideal_points` docstring documenting sign non-identifiability already landed in the Rust binding (`src/python/tbip.rs`) via #512, but was missing from the `.pyi` stub, where `ideal_points` was a bare `...`. Added the matching note (identified only up to sign; TBIP has no anchoring; compare by absolute correlation; scale softly pinned by N(0,1)) so the Python type surface matches the binding.

## (c) Clamp-region prior gradient — implemented (real bug)

In the log-space clamp tail the forward ELBO freezes `z` and `logz = clamp(u)`, so their gradients vanish and only the `ln(sig)` entropy term of `logq` stays live: `d/dmu = 0`, `d/drs = logistic(rs)/sig`. The theta/beta prior-term gradient applied `dt_dmu = a - b*z` and `dt_drs = (a - b*z)*eps*lg + lg/sig` **unconditionally**, so in the saturated tail it was not the adjoint of the coded ELBO (the likelihood term was already correctly zeroed there). Gated both behind the same `*_unclamped` flag as the likelihood gradient, keeping only `lg/sig` when clamped. In-range fits are unchanged (the `unclamped` branch is byte-identical to the old formula).

Added `fd_gradient_check_clamp_region`, which pushes theta and beta deep into their clamp regions (holding the other factor small to keep the Poisson rate O(10) and avoid FD cancellation). Verified the test **fails** on the old unconditional formula and **passes** on the fix.

## Verification

- `cargo test --lib tbip` — 5 passed (incl. new `fd_gradient_check_clamp_region` and the existing in-range `fd_gradient_check`).
- `pytest tests/test_tbip*.py -q` — 6 passed.
- `python parity/tbip_parity.py` — runs offline (no torch), `|r| = 0.996`, PASS.
- `./scripts/preflight.sh` — all checks passed (fmt, clippy, #481 guard, model tables, `.pyi` stub sync).

No version bump (fix PR). No `--no-verify`.

### DIFF
--- a/parity/tbip_parity.py
+++ b/parity/tbip_parity.py
@@ -1,24 +1,20 @@
-"""Parity check for TBIP (Text-Based Ideal Points, Vafa-Naidu-Blei 2020) against an
-independent PyTorch reference of the SAME model + the SAME mean-field VI inference.
+"""Self-consistency check for TBIP (Text-Based Ideal Points, Vafa-Naidu-Blei 2020).
 
-The official implementation is TensorFlow 1.14 / TFP 0.7 (graph-mode `tf.Session`), which
-has no Apple-Silicon build and does not install on modern Python, so we validate against a
-faithful PyTorch reimplementation of the paper's model (Eq. 3) and inference (sec. 4):
-lognormal q(theta,beta), normal q(eta,x), reparameterized single-sample SVI, Adam.
+This is NOT a cross-implementation parity run. The official implementation is
+TensorFlow 1.14 / TFP 0.7 (graph-mode `tf.Session`), which has no Apple-Silicon build
+and does not install on modern Python, so we do not run it here as a reference. Instead
+we validate the port end-to-end by *planted-position recovery*: sample a corpus from the
+TBIP generative model (Eq. 3) with known author ideal points, fit `topica.TBIP`, and
+confirm the recovered positions correlate with the planted ones (Pearson |r| > 0.85).
+The ideal-point axis is identified only up to sign (the model is invariant under
+x -> -x, eta -> -eta), so we compare by absolute correlation.
 
-Self-contained: data is sampled from the generative model with planted author ideal points
-(no external corpus). Skips cleanly when torch is unavailable (e.g. CI without torch).
+Self-contained: no external corpus and no reference toolchain, so it runs offline in CI.
 
 Run: python parity/tbip_parity.py
 """
 import sys
 import numpy as np
-
-try:
-    import torch
-except Exception:
-    print("SKIP: torch not available (reference unavailable)")
-    sys.exit(0)
 
 import topica
 
--- a/python/topica/_topica.pyi
+++ b/python/topica/_topica.pyi
@@ -5000,8 +5000,9 @@ class TBIP:
     exp(x_s * eta_kv); documents mix topics with positive per-doc intensities
     theta_dk. Fit by the paper's mean-field variational inference (reparameterized
     single-sample SVI, Adam, document minibatching). Recovers ideological scales
-    from unlabeled text. Validated by planted-position recovery and a PyTorch
-    reference (parity/tbip_parity.py)."""
+    from unlabeled text. Validated by synthetic planted-position recovery
+    (parity/tbip_parity.py); the reference TF1.14/TFP0.7 code is not run as a
+    cross-implementation check."""
     @property
     def settings(self) -> dict:
         """The constructor configuration as a JSON-serialisable dict,
@@ -5043,7 +5044,15 @@ class TBIP:
     @property
     def num_authors(self) -> int: ...
     @property
-    def ideal_points(self) -> numpy.typing.NDArray[numpy.float64]: ...
+    def ideal_points(self) -> numpy.typing.NDArray[numpy.float64]:
+        """Author ideal points (num_authors,), the posterior-mean positions mu_x.
+
+        Identifiability: identified only up to sign -- the model is invariant under
+        x -> -x, eta -> -eta, so the direction is arbitrary and determined by the
+        seed (TBIP has no anchoring). Compare runs by absolute correlation, or flip
+        to a chosen reference author. The scale is only softly pinned by the N(0, 1)
+        prior."""
+        ...
     @property
     def position_se(self) -> numpy.typing.NDArray[numpy.float64]:
         """Standard error of each author ideal point (num_authors,): the standard
--- a/src/tbip.rs
+++ b/src/tbip.rs
@@ -18,10 +18,11 @@
 //! sample. The gradients are hand-coded (reverse-mode by hand, like `prodlda.rs`)
 //! and FD-checked in the unit tests below.
 //!
-//! This is a faithful reimplementation of the published model and inference; the
-//! official code is TF1.14/TFP0.7 graph-mode (no Apple-Silicon build), so the
-//! numerics are validated against a PyTorch reference (scratchpad/tbip_vi.py),
-//! synthetic planted-position recovery, and FD gradient checks.
+//! This is a faithful reimplementation of the published model and inference. The
+//! official code is TF1.14/TFP0.7 graph-mode (no Apple-Silicon build), so we do
+//! not run it as a cross-implementation reference; instead the port is validated
+//! by synthetic planted-position recovery (`parity/tbip_parity.py`) and by the FD
+//! gradient checks in the unit tests below.
 
 use rand::Rng;
 use rayon::prelude::*;
@@ -479,6 +480,8 @@ fn elbo_and_grads(
     // theta/beta reparam: z=exp(u); dz/dmu=z; dz/drs = z*eps*logistic(rs).
     // theta/beta MC prior term: T = a*u - b*z + log(sig) + const, u=log z.
     //   dT/dmu = a - b*z ;  dT/drs = (a - b*z)*eps*logistic(rs) + logistic(rs)/sig.
+    //   In the clamp tail (u saturated) z and u are frozen, so dT/dmu = 0 and
+    //   dT/drs collapses to the ln(sig) entropy term logistic(rs)/sig (issue #506).
     // eta/x reparam: z=u; dz/dmu=1; dz/drs=eps*logistic(rs).
     // eta/x -KL: d/dmu=-mu ; d/dsig=-(sig-1/sig) ; d/drs = that*logistic(rs).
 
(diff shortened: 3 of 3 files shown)


nealcaren:
Dual review: Claude (faithfulness) + **Gemini `3.1-pro-high`** (adversarial, re-run against the actual PR worktree). Both confirm the clamp-region prior-gradient fix is the **correct adjoint of the coded ELBO**: in the saturated tail the forward freezes `logz=clamp(u)`, so the true gradient through the prior term is 0 for `dt_dmu` and only the entropy term `lg/sig` for `dt_drs`; the fix gates both behind the same `*_unclamped` flag as the (already-correct) likelihood term. In-range fits are byte-identical. The new `fd_gradient_check_clamp_region` is discriminating — Claude reverted the gate and reproduced the failure (analytic −120.73 vs FD 0). Parity harness honestly relabeled (dead torch gate removed, runs offline, no fabricated reference); .pyi sign-id note matches the binding. CI green.

PR #544, opened 2026-07-25
fix(bertopic): reconcile bm25, nr_topics, approximate_distribution with upstream (#488)

### LINKED ISSUE #488: BERTopic review: bm25/reduce_frequent, nr_topics, and approximate_distribution diverge from upstream (all outside gold coverage)
## Review result

topica's BERTopic reproduces the reference `bertopic` package (MaartenGr/BERTopic, Grootendorst 2022) on its **default path** — plain c-TF-IDF (`bm25=False`, `reduce_frequent=False`), no `nr_topics` — and both implementations recover the planted truth at ARI 1.0 on the committed gold. This review (Claude + Codex, adjudicated against upstream `_ctfidf.py` / `_bertopic.py`) found that the two documented c-TF-IDF knobs, topic reduction, and the approximate distribution each diverge from upstream, and **none of them is covered by the gold**. Nothing here is NaN/panic-critical except item 7.

Reference behavior verified against `bertopic/vectorizers/_ctfidf.py`: `fit()` computes `df` and `avg_nr_samples` (A) from the **raw** count matrix; `transform()` order is **L1-normalize rows → sqrt (if reduce_frequent) → multiply by idf**, no clamp on either idf formula.

### 1. `nr_topics` reduction is not BERTopic's algorithm, and is off-by-one against the `-1` topic

`src/bertopic.rs:211-222,297`. topica greedily selects the single most cosine-similar c-TF-IDF pair, merges it, recomputes c-TF-IDF, and repeats. BERTopic's `_reduce_to_n_topics` fits one `AgglomerativeClustering(n_clusters = nr_topics - self._outliers, metric="euclidean", linkage="ward")`, and **at fit time defaults to the topic *embeddings*, not c-TF-IDF**, as the distance source. Different merge trees, different distance space.

Off-by-one: BERTopic reads `nr_topics=N` as N **total** topics including the `-1` outlier topic (so it clusters real topics to `N - _outliers`); topica's `topic_count` ignores `-1`, so `nr_topics=N` yields N **real** topics — one more real topic than BERTopic for the same argument whenever noise exists.

- [ ] Align the target to `nr_topics - outliers`, or document the target-count convention and the greedy-vs-ward divergence explicitly (the module docstring's "agglomeratively merge the most c-TF-IDF-similar topics" overstates the correspondence).

### 2. BM25 c-TF-IDF clamps to 0 where BERTopic goes negative

`src/represent.rs:211-215`. topica computes `(1.0 + (A - f_t + 0.5)/(f_t + 0.5)).max(1.0).ln()`. BERTopic's `np.log(1 + (A - df + 0.5)/(df + 0.5))` has **no clamp**, and the argument drops below 1 (→ negative idf) whenever `df > A + 1`. Because `df` is the corpus-wide token count and `A` the average class size, `df > A` is the common case, not a corner case. Verified: `df=30, A=10` → upstream idf −1.02 vs topica 0.0. BERTopic actively ranks ubiquitous terms *below* everything; topica ties them with absent terms.

- [ ] Decide whether `bm25=True` should match upstream (drop the clamp) or document the deviation; add a `df > A` parity fixture.

### 3. `approximate_distribution` diverges from `BERTopic.approximate_distribution`

`src/bertopic.rs:351-401`, window idf from `idf_weights` at `bertopic.rs:270-295`.

- **No `min_similarity` gate.** BERTopic zeroes any window↔topic cosine below `min_similarity` (default 0.1) before summing; topica only clamps negatives, so low-similarity windows still contribute mass.
- **Window weighting ignores the knobs.** The window bag always uses the plain `ln(1 + A/f_t)` idf and never applies `bm25` or `reduce_frequent`, but the topic rows it is compared against (`ctfidf_raw`) were built with those knobs on. BERTopic transforms windows and topics through the *same* fitted `ctfidf_model`, so the weightings always match. Mismatched per-dimension weighting points the window in a different direction than the reference.
- **Uniform vs zero on no evidence.** An all-OOV / zero-similarity document becomes a **uniform** row in topica; BERTopic's thresholded similarities sum to zero and L1 normalization leaves a **zero** row. Different meaning of "no evidence."

- [ ] Thread `bm25`/`reduce_frequent` into the window idf (reuse the weighting used for `ctfidf_raw`); add an optional `min_similarity` threshold; decide zero-row vs uniform-row semantics.

### 4. `-1` (noise) documents are excluded from c-TF-IDF; BERTopic keeps them as an outlier class

`src/represent.rs:178`. topica skips negative labels entirely. BERTopic groups documents by topic **including `-1`**, so the outlier group's tokens enter `df` and `A`, shifting every real-topic idf whenever noise exists. This is also the root of the reduction-count distinction in item 1.

- [ ] Decide whether the `-1` pseudo-class should participate in `df`/`A` for reference parity; document either way.

### 5. Save/load silently drops four constructor fields

`src/python/embedding_cluster.rs:742-763,1304-1311`. `BertopicState` omits `resolution`, `knn_neighbors`, `umap_params`, and `diagnostics`; `load()` hardcodes them to defaults (`1.0`, `15`, `UmapParams::default()`, `true`). A model fit with e.g. `clusterer="louvain", resolution=2.0` reports the wrong `.settings` after a save→load round-trip. Functionally safe (a loaded model never re-clusters, and `merge_topics`/`reduce_outliers` rebuild only from persisted knobs), but `.settings` is inaccurate.

- [ ] Persist the four fields (with `#[serde(default)]` for old-save compatibility) so `.settings` is truthful after load, or document that fit-time knobs reset on load.

### 6. `avg_nr_samples`: float mean vs BERTopic's integer truncation

`src/represent.rs:201`, `src/bertopic.rs:287-291`. BERTopic uses `int(X.sum(axis=1).mean())`; topica uses the untruncated float mean. Shifts every idf by a hair with no ranking change on the default corpus. Trivial (`.floor()`) if exact-value parity is wanted.

### 7. Ragged `doc_embeddings` panic instead of `ValueError`

`src/python/embedding_cluster.rs:969`, `src/cluster.rs:32`. The binding checks finiteness and row count but not rectangularity; a jagged embedding matrix reaches the clusterer, where an identical-row-dimension assertion escapes as a Rust panic rather than a clean `PyValueError`. (All other robustness paths checked out: empty corpus, all-noise labels, `num_topics==0`, OOV tokens, zero-norm cosine, and zero-sum normalize are guarded, with tests.)

- [ ] Reject ragged embedding matrices at the binding before calling the reducer/clusterer.

### 8. Default pipeline is not the package's default (documentation honesty)

`src/python/embedding_cluster.rs:859`. topica defaults to `reducer="pca"`, `min_cluster_size=15`; the `bertopic` package defaults to UMAP (`n_components=5, n_neighbors=15, min_dist=0`, cosine) + HDBSCAN with `min_topic_size=10`. This is a deliberate determinism/no-heavy-deps choice, documented well in the gold harness, but neither the constructor docstring nor `docs/guides/embedding.md` notes that the package itself defaults to UMAP / `min_cluster_size=10`.

- [ ] One sentence in the docstring/guide naming the upstream defaults so PCA-default isn't read as parity.

### Explicitly refuted (checked, not a bug)

`reduce_frequent_words` does **not** change the idf. Upstream computes `df`/`A` from raw counts and applies `sqrt` only in `transform`; the `transform`-time L1 row-normalization is a per-class scalar that cancels under both topica's topic-row normalization and cosine similarity, so `sqrt(tf)·idf` yields identical within-topic top words. No action.

### Highest-value follow-up

Extend `parity/bertopic_gold.py` beyond the default path to a `bm25=True`, a `reduce_frequent=True`, and an `nr_topics`-with-noise case — that coverage would have caught items 2, 3, and 4.

Relevant surfaces: `src/bertopic.rs`, `src/represent.rs`, `src/python/embedding_cluster.rs`, `parity/bertopic_gold.py`, `tests/test_bertopic_gold.py`.

nealcaren:
Closes #488.

The #488 review flagged three BERTopic knobs as diverging from the upstream `bertopic` package (MaartenGr/BERTopic) and outside the gold coverage. The default c-TF-IDF path was already faithful (ARI 1.0 on the committed gold); these are the opt-in knobs. Each item is reconciled with upstream, or the divergence is documented precisely where a faithful port is a larger, separate change.

### (a) `bm25` / `reduce_frequent` c-TF-IDF variants — FIXED (bm25); already-faithful (reduce_frequent)

- **`bm25`**: dropped the `.max(1.0)` clamp on the class-based BM25 idf in `ctfidf_weighted` so it matches upstream `ClassTfidfTransformer(bm25_weighting=True)` exactly, `ln(1 + (A - f_t + 0.5)/(f_t + 0.5))` with no floor. When `f_t > A + 1` (the common case, since `f_t` is the corpus-wide count and `A` the average class size) the idf goes **negative**, which is how upstream ranks a term ubiquitous across classes *below* one that is merely absent. Previously topica clamped it to 0, tying ubiquitous terms with absent ones.
  - To keep topica's cross-model `topic_word` surface a valid non-negative distribution, the normalized `topic_word` floors the negative entries to zero *before* normalizing (a new `topic_word_from_ctfidf` helper). This only touches terms upstream already ranks at the bottom, so the top-word ordering matches. The signed raw c-TF-IDF is retained for the cosine similarities in `approximate_distribution` and topic reduction — the places upstream uses the signed values.
- **`reduce_frequent`**: already faithful. As the #488 review's "explicitly refuted" note establishes, upstream applies `sqrt(tf)` only in transform and its L1 row-normalization is a per-class scalar that cancels under both topica's topic-row normalization and cosine, so `sqrt(tf)·idf` gives the same within-topic ranking. No change; the docstring no longer overstates the bm25 case as matching while it clamped.

### (b) `nr_topics` reduction — DOCUMENTED (not changed)

topica reduces by a greedy merge of the most c-TF-IDF-similar pair; upstream fits one `AgglomerativeClustering(linkage="ward", metric="euclidean")` over the topic **embeddings** and cuts at `nr_topics - _outliers`. Different distance space and different merge tree, and topica's representation is embedding-free at this stage. A faithful ward-on-embeddings port is a larger, separate change, so this PR states the divergence precisely rather than overstating equivalence:

- topica reads `nr_topics` as the number of **real** topics (the `-1` noise topic is never counted); upstream counts `-1` toward the total.
- The greedy-c-TF-IDF vs ward-on-embeddings difference is now called out in the module docs, `fit_bertopic`, the constructor docstring, the `.pyi` stub, and the embedding guide. The old "agglomeratively merge the most c-TF-IDF-similar topics" wording that overstated the correspondence is corrected.

### (c) `approximate_distribution` — FIXED

- **Window weighting now tracks the knobs.** Each sliding window is built with the **same** `bm25`/`reduce_frequent` idf the topics were built with (`idf_weights` gained a `bm25` argument; the stored `idf` is now the fitted, weighting-aware vector, and `reduce_frequent` damps window counts by `sqrt` before the idf). Previously the window always used the plain idf while the topic rows used the knobs, pointing the window in a different per-dimension space than the reference.
- **`min_similarity` gate added.** Any window-to-topic cosine below `min_similarity` (and any negative cosine) is dropped before averaging, matching upstream's gate. Exposed as a constructor parameter and as a per-call override on `approximate_distribution(...)`. Default `0.0` preserves current output; upstream uses `0.1`, so pass `min_similarity=0.1` to match it (documented).
- **Zero-row vs uniform-row: documented decision.** A document with no surviving evidence stays a **uniform** row so `doc_topic` remains a valid per-document distribution (topica's cross-model contract and conformance requirement); upstream leaves it a zero row. Documented as an intentional difference rather than changed.

Also added a one-line honesty note (constructor docstring + guide) that topica's `reducer="pca"` / `min_cluster_size=15` defaults are not the package's parity defaults (UMAP + `min_topic_size=10`).

Out of scope for this PR: the review's items 5 (save/load dropping four fields), 6 (`avg_nr_samples` float vs int truncation), and 7 (ragged-embedding panic) — none of the three flagged items.

## Verification

- `cargo test --lib --features embeddings` — all pass, including 4 new unit tests: unclamped BM25 idf goes negative when `df > A` (`represent.rs`), negative-weight flooring in `topic_word`, bm25-aware window idf equals the ctfidf idf, and the `min_similarity` gate + uniform fallback (`bertopic.rs`).
- `pytest tests/test_bertopic_upstream.py -q` — 9 passed (new): bm25 topic_word is a valid distribution, bm25 drops the ubiquitous term from top words, bm25+reduce_frequent compose, `nr_topics` reduces to the requested real-topic count, `approximate_distribution` is a valid distribution, the `min_similarity` gate sparsifies then falls back to uniform, and `min_similarity` round-trips through `settings` and save/load.
- `pytest tests/test_bertopic_gold.py` — passes; the default gold path is unchanged.
- `./scripts/preflight.sh` — all checks passed (fmt, clippy `-D warnings`, `#481` guard scan, model tables, `.pyi` stub sync).

The pre-existing 59 failures in the full `pytest tests/` run are all `ModuleNotFoundError: scipy` / torch in unrelated gold tests, from the minimal offline agent venv (reference toolchains intentionally not installed); none touch BERTopic.

### DIFF
--- a/docs/guides/embedding.md
+++ b/docs/guides/embedding.md
@@ -75,7 +75,7 @@ model.doc_topic                        # (num_docs, num_topics) soft membership
 model.labels                           # hard cluster per doc; -1 is noise
 ```
 
-Two BERTopic features carry over. `nr_topics` merges the most similar topics down
+Two BERTopic features carry over. `nr_topics` reduces the discovered topics down
 to a target count:
 
 ```python
@@ -90,6 +90,37 @@ topic. It is the default `doc_topic`, and you can also run it on new documents:
 ```python
 dist = model.approximate_distribution(new_docs, window=4, stride=1)  # (n, num_topics)
 ```
+
+`min_similarity` (default `0.0`) drops any window-to-topic cosine below its value
+before averaging; a document with no surviving evidence becomes a *uniform* row
+so `doc_topic` stays a valid distribution.
+
+### Where topica differs from the `bertopic` package
+
+topica reproduces the reference `bertopic` package on its default c-TF-IDF path,
+but a few knobs diverge on purpose (issue #488), so name them when you report a
+BERTopic run:
+
+- **Defaults are not parity settings.** topica defaults to `reducer="pca"` and
+  `min_cluster_size=15` for a deterministic, dependency-free pipeline. The package
+  itself defaults to UMAP (`n_components=5`, `n_neighbors=15`, `min_dist=0`, cosine)
+  + HDBSCAN with `min_topic_size=10`. Pass `reducer="umap"` and matching sizes to
+  get close to the package's behavior.
+- **`nr_topics` reduction.** topica greedily folds the most c-TF-IDF-similar pair
+  of topics; the package fits one ward `AgglomerativeClustering` over the topic
+  *embeddings*. Different distance space and merge tree, so the two can pick
+  different merges. topica also reads `nr_topics` as the number of *real* topics
+  (the `-1` noise topic is never counted), whereas the package counts `-1` toward
+  the total.
+- **`bm25=True`** matches the package's class-based BM25 idf exactly, including the
+  unclamped log that goes *negative* for a term common to every class (which ranks
+  such terms last). The normalized `topic_word` surface floors those negatives to
+  zero so it stays a valid distribution; ranking is unaffected.
+- **`approximate_distribution`** weights each window with the same
+  `bm25`/`reduce_frequent` idf the topics used. The `min_similarity` gate defaults
+  to `0.0`; the package uses `0.1` for its own approximate distribution, so pass
+  `min_similarity=0.1` to match it. topica returns a uniform row for a document
+  with no surviving evidence where the package returns a zero row.
 
 ## Top2Vec
 
--- a/python/topica/_topica.pyi
+++ b/python/topica/_topica.pyi
@@ -3292,9 +3292,14 @@ class Top2Vec:
 class BERTopic:
     """BERTopic (Grootendorst 2022): the same reduce/cluster pipeline as Top2Vec,
     but topics are defined by class-based TF-IDF over their documents' words, so
-    no word embeddings are needed. `nr_topics` merges the most similar topics down
-    to a target; `doc_topic` is the approximate distribution. You bring the
+    no word embeddings are needed. `nr_topics` reduces the discovered real topics
+    down to a target (topica's greedy c-TF-IDF merge, not the upstream package's
+    ward agglomeration over topic embeddings, and it counts real topics only, not
+    the -1 noise topic); `doc_topic` is the approximate distribution. You bring the
     document embeddings; the topic count is discovered (before any reduction).
+    topica defaults to ``reducer="pca"`` / ``min_cluster_size=15`` for
+    determinism; the upstream package defaults to UMAP + HDBSCAN
+    (``min_topic_size=10``), so the defaults are not a parity setting (issue #488).
     No embedder of your own? ``topica.llm_embed(texts, model=...)`` builds it."""
     @property
     def settings(self) -> dict:
@@ -3321,6 +3326,7 @@ class BERTopic:
         n_neighbors: int = 15,
         bm25: bool = False,
         reduce_frequent: bool = False,
+        min_similarity: float = 0.0,
         clusterer: str = "hdbscan",
         num_clusters: int | None = None,
         resolution: float = 1.0,
@@ -3365,6 +3371,7 @@ class BERTopic:
         *,
         window: int | None = None,
         stride: int | None = None,
+        min_similarity: float | None = None,
     ) -> numpy.typing.NDArray[numpy.float64]: ...
     def transform(
         self,
--- a/src/bertopic.rs
+++ b/src/bertopic.rs
@@ -6,12 +6,26 @@
 //! BERTopic needs no word embeddings. Two features are characteristic and we port
 //! them here:
 //!
-//! - `nr_topics`: agglomeratively merge the most c-TF-IDF-similar topics down to a
-//!   target count, BERTopic's topic reduction.
+//! - `nr_topics`: reduce the discovered topics down to a target count. topica's
+//!   reduction is a **greedy** merge — it repeatedly folds the single most
+//!   cosine-similar pair of topic c-TF-IDF rows and recomputes — which differs
+//!   from upstream BERTopic. Upstream fits one `AgglomerativeClustering`
+//!   (`linkage="ward"`, Euclidean) over the topic *embeddings* and cuts it at
+//!   `nr_topics - _outliers` clusters. Different distance space (c-TF-IDF vs
+//!   embeddings) and different merge tree (greedy vs ward), so the two need not
+//!   pick the same merges. topica also reads `nr_topics` as the number of **real**
+//!   topics (the `-1` noise topic is not counted), whereas upstream counts the
+//!   `-1` topic toward the total. See `fit_bertopic` (issue #488).
 //! - `approximate_distribution`: a soft per-document topic distribution. We slide
-//!   a window over each document, build the window's c-TF-IDF vector, measure its
-//!   cosine to every topic, and average across windows. This is the document-topic
-//!   distribution BERTopic reports without re-running the clustering.
+//!   a window over each document, build the window's c-TF-IDF vector *with the same
+//!   `bm25`/`reduce_frequent` weighting the topics were built with* (issue #488),
+//!   measure its cosine to every topic, drop any window↔topic similarity below
+//!   `min_similarity`, and average across windows. This is the document-topic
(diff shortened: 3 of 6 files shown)


nealcaren:
Dual review: Claude (faithfulness) + **Gemini `3.1-pro-high`** (adversarial, re-run against the actual PR worktree). Both confirmed the three knobs are faithful/honest: bm25 idf matches upstream's unclamped `ln(1 + (A-f+0.5)/(f+0.5))` (goes negative for ubiquitous terms), the floor is applied only to the exposed `topic_word` while the signed c-TF-IDF still feeds the cosine/reduction, `approximate_distribution` windows now share the topics' weighting + `min_similarity` gate, and the **default path is byte-identical** (flooring is a no-op on non-negative idf; `min_similarity=0.0` drops nothing new). Both independently flagged one real edge — an all-negative topic row → all-zero (breaks rows-sum-to-1) — with the same uniform-fallback fix. **Amended**: added the uniform fallback + a regression test, and softened the bm25 docstring (upstream truncates the average class size to an int; topica doesn't, so it matches the *formula*, not byte-for-byte).

PR #545, opened 2026-07-25
fix(combinedtm): faithful adapt_bert encoder + raw BoW (#503)

### LINKED ISSUE #503: CombinedTM/ZeroShotTM review: shared ProdLDA VAE core faithful + FD-verified, ZeroShotTM exact; CombinedTM encoder diverges from Bianchi et al. (no adapt_bert projection, normalized vs raw BoW)
## Review result

Two independent Claude reviewers — a faithfulness reviewer (encoder-input modes, Laplace prior, KL, batchnorm vs MilaNLProc/contextualized-topic-models + AVITM) and an adversarial VAE-gradient reviewer (hand-deriving the batchnorm/reparam/KL/decoder backward to *break* it) — plus a maintainer trace **verified against the upstream `inference_network.py` source**. (Codex ghosted and Gemini's headless sandbox blocked the file reads, so the second reviewer is a second Claude subagent with an adversarial brief.) CombinedTM and ZeroShotTM are the **same** ProdLDA VAE core (`src/prodlda.rs`) selected by `InputMode::BowEmb` vs `InputMode::EmbOnly`; this issue covers both.

**Bottom line: the VAE core is a faithful, FD-verified ProdLDA/AVITM implementation, and ZeroShotTM matches the reference exactly — but CombinedTM's encoder architecture diverges from Bianchi et al.** No hand-coded-gradient bug: the adversarial reviewer confirmed the three-term batchnorm backward, the softmax+reparam (cross-term + 0.5/eps chain), the logistic-normal KL, and the softmax-CE decoder are all correct and FD-verified over a **batch of 3 with all three batchnorms active** (the exact config that would expose a missing batch-coupling term).

### 1. CombinedTM encoder does not match `CombinedInferenceNetwork` — missing `adapt_bert` projection + normalized vs raw BoW (MEDIUM, CombinedTM only)

Verified against `contextualized_topic_models/networks/inference_network.py` (upstream master):

- **Upstream `CombinedInferenceNetwork`:** `adapt_bert = nn.Linear(bert_size, input_size)` projects the SBERT embedding to **vocab size**, then `x = cat([x, adapt_bert(x_bert)])` (raw BoW ⊕ adapted embedding) → `input_layer = nn.Linear(input_size + input_size, hidden)` = `Linear(2·V, hidden)`. The BoW is fed as **raw CountVectorizer counts**, no normalization.
- **topica CombinedTM (`InputMode::BowEmb`):** encoder input is `[normalized_bow (V), raw_emb (E)]` → `w1` is `hidden × (V + E)` (`src/prodlda.rs:244-274,281-336`). There is **no `adapt_bert` layer**, and the BoW is length-normalized (`normalized_bow`, `prodlda.rs:1622-1629`).

Two source-confirmed divergences from the published CombinedTM: **(a)** the learned `adapt_bert` embedding→vocab projection is absent (topica concatenates the raw embedding directly, so the first-layer input is `V+E` not `2V`, and the embedding is not adapted into the BoW space before mixing); **(b)** the encoder BoW is normalized rather than raw, changing the BoW-block-to-embedding-block scale ratio the encoder sees. This changes CombinedTM's encoder capacity and input geometry relative to the reference. The committed `combinedtm_gold.py` cannot catch this — it mirrors topica's simplified architecture (also normalizes the BoW, no `adapt_bert`), so the parity claim is against a matched in-repo reimplementation, not the published `contextualized-topic-models` package. (ZeroShotTM is unaffected — no BoW in its encoder.)

- [ ] Decide: either implement CombinedTM faithfully (add an `adapt_bert` `Linear(E, V)` projection, concatenate with **raw** BoW, first layer `2V`), or document prominently that topica's CombinedTM is a *simplified* variant (raw-embedding concat, normalized BoW, no adapt projection) that differs from Bianchi et al.'s architecture — and regenerate/annotate the gold accordingly so "CombinedTM (Bianchi, Terragni & Hovy 2021)" isn't read as architecture-faithful.

### 2. ZeroShotTM — faithful (confirmed against source)

`ContextualInferenceNetwork` upstream: `input_layer = nn.Linear(bert_size, hidden)`, BoW **ignored** (`x = x_bert`). topica `InputMode::EmbOnly`: encoder input is `[emb (E)]`, `w1` first layer `hidden × E`, BoW columns never read or updated. **Exact match**, including the cross-lingual `transform` path (embedding-only encoder → OOV tokens in a new language are harmless).

### 3. Minor / family-wide

- **Batch/epoch defaults follow ProdLDA-original, not the CTM package** (LOW): `batch_size=200`, `iters/epochs=200` (`neural.rs:2696`) follow Srivastava-Sutton; the upstream CTM class defaults are `num_epochs=100`, `batch_size=64`. Defensible (these *are* ProdLDA) and the docstring says "reference 200/0.002", but a user expecting CTM-package defaults gets different ones. Cosmetic.
- **`lr`, `hidden_size`, `batch_size` escape the #481 guards** (LOW, both + plain ProdLDA): `new` (`neural.rs:2613-2621`) guards `num_topics≥2`, `finite_pos(alpha)`, `dropout∈[0,1)`, but not `lr`/`hidden_size`/`batch_size` — a NaN/inf/negative `lr` or `hidden_size=0` yields NaN/garbage topics silently. Family-wide omission (plain `ProdLDA` at `neural.rs:2045-2048` shares it). Add `finite_pos(lr)`, `hidden_size≥1`, `batch_size≥1`.
- **ZeroShotTM stores a dead `hidden×V` weight block** (LOW, efficiency): for `EmbOnly`, `w1` is allocated `hidden×(V+E)` but the V-column block is never read or updated (gradient 0, `wd=0`) — wasted memory + Adam state at large V. Could allocate `hidden×E`.
- **No explicit logvar clamp** (robustness note, not a bug): unlike DETM, ProdLDA has no `logvar` clamp, but the logvar head is affine-free batch-normed so it stays bounded — consistent with the reference, which also has none.

### Checked and CORRECT (traced against `parity/refs/avitm.py` + upstream + FD-verified)

**Laplace prior** (`prodlda.rs:407-418`): `mu1 = ln a − mean(ln a)`, `var1 = (1/a)(1−2/K) + (1/K²)Σ(1/a)` — byte-identical to `refs/avitm.py`, reduces to `(0, 1−1/K)` at `alpha=1` (the upstream CTM default). **KL** (`prodlda.rs:1073-1081`): `0.5·Σ[var_q/var_p + (mu_p−mu_q)²/var_p − 1 + ln var_p − ln var_q]` — matches term-for-term including **both** log-determinant terms; gradients `dmu += (mu−pm)/pv`, `dlv += 0.5(exp(lv)/pv − 1)` correct. **Triple affine-free batchnorm** on the mu head, logvar head, and decoder logits (`eps=1e-5`, `momentum=0.1`), biased (÷n) variance in the forward with Bessel (÷(n−1)) running stats — matches `nn.BatchNorm1d(affine=False)`; three-term backward FD-verified. **Reparam** `z=mu+exp(0.5·lv)·eps`, `theta=softmax(z)` with softmax cross-term. **Reconstruction** `softmax(BN(theta_do·beta))`, loss `−Σ count·log(recon+1e-10)`, decoder gradient `total·recon − count`, unnormalized `beta` (product-of-experts), dropout on `theta` before the matmul — all match. **Defaults** `alpha=1`, `hidden=100`, `dropout=0.2`, `lr=0.002`, `beta1=0.99`, `wd=0`. **Determinism** (`ChaCha8Rng` + Fisher-Yates), save/load (persists `w_e`, `mode`, logvar-BN stats), stub in sync. The `prior="dirichlet"/"stick_breaking"` and `contrastive` flags are topica extensions beyond Bianchi et al. (laplace-only), correctly gated off by default and separately FD-checked.

### Bottom line

ZeroShotTM is faithful. CombinedTM's math and gradients are correct, but its encoder architecture is a simplified variant of the published CombinedTM (no `adapt_bert`, normalized rather than raw BoW) — the one substantive item to fix or document. Everything else is minor guard/efficiency/default polish shared across the ProdLDA family.

Relevant surfaces: `src/prodlda.rs`, `src/python/neural.rs:2424-3100`, `parity/combinedtm_gold.py`, `parity/zeroshot_gold.py`, `parity/refs/avitm.py`, `python/topica/_topica.pyi:3883-3898`.

nealcaren:
Closes #503.

CombinedTM's encoder now matches Bianchi, Terragni & Hovy (2021)'s
`CombinedInferenceNetwork`. ZeroShotTM, plain ProdLDA, and SCHOLAR are unchanged.

## Background

The review on #503 confirmed the shared ProdLDA/AVITM VAE core is faithful and
FD-verified and that ZeroShotTM is exact, but flagged two source-confirmed
divergences in CombinedTM's encoder (verified against upstream
`contextualized_topic_models/networks/inference_network.py` and `models/ctm.py`):

- **(a)** the reference projects the contextual (SBERT) embedding through a learned
  `adapt_bert = nn.Linear(bert_size, input_size)` into vocabulary space *before*
  concatenation, so the first encoder layer is `Linear(2V, hidden)`. topica omitted
  this and concatenated the raw embedding (first layer `V + E`).
- **(b)** the reference feeds the encoder the **raw** `CountVectorizer` bag of words;
  topica fed the length-normalized BoW.

## What changed

Both items are implemented for CombinedTM. Because SCHOLAR (`src/scholar.rs`) reuses
the existing `InputMode::BowEmb` for its own covariate-concatenation encoder
(normalized BoW + a raw auxiliary channel, no `adapt_bert`, and validated that way),
changing `BowEmb` globally would have broken SCHOLAR. So the faithful encoder is a
new, isolated variant:

- **New `InputMode::BowEmbAdapt`** (`src/prodlda.rs`): raw BoW concatenated with
  `adapt_bert(emb)`, first layer `Linear(2V, hidden)`. CombinedTM's binding now selects
  this mode; `BowEmb` (SCHOLAR), `EmbOnly` (ZeroShotTM), and `BowOnly` (ProdLDA) are
  untouched.
- **New `adapt_bert` layer** (`w_adapt` `V×E`, `b_adapt` `V`) added to `Weights`,
  wired through the encoder forward, backward, `Grad`, `Optim` (Adam), and save/load.
  It is drawn only for `BowEmbAdapt`, so every other mode's RNG stream — and therefore
  its fitted weights — is byte-identical to before.
- **Raw BoW at the encoder** for `BowEmbAdapt` in both `batch_forward`/`batch_backward`
  and `transform_with_emb` (`batch.counts` / `raw_bow` instead of `normalized_bow`).
  The reconstruction target was already the raw counts, so it is unchanged.
- **Save/load**: `CtmEmbState` gains `w_w_adapt`/`w_b_adapt` (`#[serde(default)]`) and a
  new mode byte `3`. A pre-#503 CombinedTM save carries mode `1` and loads back through
  the earlier simple-concat encoder, matching the weights it was trained with.

Per-item: **(a) implemented**, **(b) implemented**. Nothing skipped.

## Gold / parity honesty

The committed gold's PyTorch reference (`parity/combinedtm_gold.py`, and the live
`parity/combinedtm_compare.py`) previously mirrored the simplified architecture, so it
could not have caught this divergence. Both reference reimplementations are now updated
to the faithful architecture (`adapt_bert = Linear(E, V)`, `Linear(2V, hidden)`, raw
BoW) and `parity/combinedtm_gold.{npz,json}` were regenerated with torch. On the clean
planted-block design topica-vs-reference aligned topic-word cosine is **0.9846**
(reference seed-to-seed self cosine 0.9797, bar 0.8797).

## Verification

- New Rust FD gradient check `batch_gradients_match_fd_bow_emb_adapt` (perturbs the new
  `w_adapt`/`b_adapt` blocks end-to-end); new recovery test
  `combinedtm_adapt_recovers_planted_blocks` (also asserts `w_adapt` is live/trained);
  `embedding_modes_are_deterministic` extended to `BowEmbAdapt`.
- `cargo test --lib`: 240 passed.
- New Python tests: `test_combinedtm_adapt_bert_projection_is_live` (embedding channel
  measurably flows through `adapt_bert`) and **`test_zeroshottm_output_unchanged_regression`**
  (frozen ZeroShotTM `doc_topic` values — the requested ZeroShotTM-unchanged guard).
- `pytest tests/test_combinedtm*.py tests/test_prodlda*.py tests/test_scholar.py
  tests/test_stub_sync.py`: 165 passed. Save/load, InfoCTM, ETM suites: green.
- `./scripts/preflight.sh`: all checks passed (fmt, clippy, #481 guard, model tables,
  stub sync).
- ZeroShotTM stays exact: the `EmbOnly` code path is not touched by any edit (the new
  logic is gated behind `BowEmbAdapt`), and the regression guard freezes its output.

## Note for reviewer

This adds a new learned layer (`adapt_bert`) to a paper-referenced model, which is an
architecture change and a save-format addition (new `Weights` fields + `CtmEmbState`
fields + mode byte `3`). New saves are self-consistent; old CombinedTM saves route
through the prior simple-concat encoder via the mode byte (positional bincode means a
pre-#428-era save still needs a refit, unchanged from before). No version bump (fix PR).


nealcaren:
Dual review: Claude (faithfulness) + **Gemini `3.1-pro-high`** (adversarial, re-run against the actual PR worktree). Both confirmed the core is correct: the **byte-identical-other-modes claim holds** — on main `cols=v+e` fixed with per-mode `fan_in`, and #545 preserves that for BowOnly/BowEmb/EmbOnly while drawing `adapt_bert` conditionally only for BowEmbAdapt, so ProdLDA/ZeroShotTM/SCHOLAR/InfoCTM RNG streams (and fitted weights) are unchanged (I verified the `cols/fan_in` match against `git show main`). The adapt_bert forward is `Linear(2V,hidden)` on `[raw_bow ; adapt_bert(emb)]`, the V×E gradient orientation is correct, Adam updates both blocks, and it's FD-verified (`batch_gradients_match_fd_bow_emb_adapt`); `transform_with_emb` uses `raw_bow` for BowEmbAdapt; no panic for non-adapt modes. One real MEDIUM both flagged: the save-compat comment overclaimed — under positional bincode old CombinedTM/ZeroShotTM saves fail to load (not graceful fallback). **Amended** the comments to the honest #428 wording. Note: the gold (disjoint 5-block) is a weak discriminator (any K=5 model passes); the real validation is the FD-verified gradient + the faithful-reference cosine (0.9846) — a tighter non-disjoint gold is a reasonable follow-up.

PR #547, opened 2026-07-25
fix(spectral): configurable projection threshold, default 10000 to match R stm (#542)

### LINKED ISSUE #542: Spectral: random-projection threshold (V>3000) diverges from R stm (~10000); approximate init for mid-size vocab
## Fidelity gap (engine)
`topica-core/src/spectral.rs:27-32` sets `PROJ_THRESHOLD = 3000` (with `PROJ_DIM = 1024`, fixed `PROJ_SEED`). Vocabularies between ~3000 and ~10000 types therefore get a **random-projected** spectral init, where R `stm`'s spectral (Arora anchor-word) still uses the exact co-occurrence Gram matrix up to ~10000. Result: a different (approximate) initialization → a different fit for those corpora.

## Why here (not in a binding)
This is a shared-engine constant. It affects **every binding**: faSTM (R), Python topica, and fastm (Stata) all inherit it. It cannot be fixed downstream.

## Options
- Raise `PROJ_THRESHOLD` to match stm (~10000), or
- Make it configurable (with the current value as default), and/or
- Document the divergence and the determinism guarantee (fixed `PROJ_SEED`).

## Refs
- `topica-core/src/spectral.rs:27-32`
- Downstream symptom tracked at nealcaren/faSTM-stata#6 (found via a Stata-side fidelity review).

nealcaren:
Closes #542.

## Problem

The spectral (anchor-word) init in the shared `topica-core` crate hardcoded a random-projection switch-over at a vocabulary of `PROJ_THRESHOLD = 3000` (`topica-core/src/spectral.rs`). Vocabularies between ~3000 and ~10000 types therefore received a **random-projected** (approximate) init, while R `stm`'s spectral init still builds the **exact** co-occurrence Gram matrix up to ~10000. The result was a different, less faithful init (and hence fit) for mid-size corpora. Because this lives in the shared engine, every binding inherited it.

## Change

Make the switch-over point configurable and **raise the default to 10000 to match R `stm`**.

- **`topica-core/src/spectral.rs`**: rename the hardcoded const to `pub const DEFAULT_PROJ_THRESHOLD = 10000`; add `spectral_init_with_threshold(docs, k, v, proj_threshold)` and keep `spectral_init` as a wrapper that passes the default. `PROJ_DIM` and `PROJ_SEED` are unchanged. Fixed the stale comment claiming the 3000 threshold matched `stm`, and documented the fixed-`PROJ_SEED` determinism guarantee (holds on both the exact and projected paths).
- **`topica-core/src/ctm.rs`**: thread a `spectral_proj_threshold` parameter through `fit_ctm` and `fit_ctm_svi` into the spectral-init call.
- **`src/python/mod.rs`**: expose it as an optional keyword on **`STM.fit`** and **`CTM.fit`**:
  - parameter name: **`spectral_projection_threshold`**, default **`10000`** (`spectral::DEFAULT_PROJ_THRESHOLD`).
- Updated the `python/topica/_topica.pyi` stub and `src/conformance.rs`.

The exact and projected code paths themselves are unchanged; only the switch-over point and its default move.

## Behavioral change

Raising the default from 3000 to 10000 changes the spectral init (hence the fit) for corpora with a vocabulary in the ~3000-10000 range: they now get the **exact** init, matching `stm`. Small-vocab (<3000) behavior is unchanged. **faSTM (R) and fastm (Stata) inherit the new default** when they update their vendored `topica-core` (and will need to pass the new `fit_ctm`/`fit_ctm_svi` argument).

The committed STM/CTM gold fixtures use a frozen poliblog subsample with a vocabulary of **2632 types** (< 3000), so they take the exact path both before and after this change — no gold shift, and both `test_stm_gold.py` and `test_ctm_gold.py` pass unchanged.

## Verification

- `cargo test --lib -p topica-core` — 44 passed (includes the new `threshold_selects_exact_vs_projected_path_and_stays_deterministic` plus the two existing projected-path tests, updated to force the projected path via an explicit low threshold).
- `cargo test --lib` — passed.
- `pytest tests/test_stm_gold.py tests/test_stm.py tests/test_stm_model.py tests/test_ctm_gold.py tests/test_ctm.py` — 190 passed (includes 3 new `TestSpectralProjectionThreshold` tests: default == high-explicit threshold, low threshold changes the result, projected path deterministic). Gold tests require `scipy`; they skip/error cleanly when the reference stack is absent.
- `./scripts/preflight.sh` — all checks passed (fmt, clippy, `#481` guard scan, model tables, `.pyi` stub sync).

### DIFF
--- a/python/topica/_topica.pyi
+++ b/python/topica/_topica.pyi
@@ -473,6 +473,7 @@ class CTM:
         em_tol: Optional[float] = None,
         keep_eta_cov: bool = True,
         num_threads: Optional[int] = None,
+        spectral_projection_threshold: int = 10000,
     ) -> "CTM":
         """EM stops once the relative change in the variational bound falls below
         convergence_tol or after iters iterations, whichever comes first. Pass
@@ -650,6 +651,7 @@ class STM:
         covariates: Optional[numpy.typing.NDArray[numpy.float64]] = None,
         keep_eta_cov: bool = True,
         num_threads: Optional[int] = None,
+        spectral_projection_threshold: int = 10000,
     ) -> "STM":
         """Fit. prevalence (or covariates, a symmetric alias) is (num_docs, F)
         covariates driving topic proportions (mu_d = X_d gamma; intercept
--- a/src/conformance.rs
+++ b/src/conformance.rs
@@ -369,6 +369,7 @@ mod ctm_conformance_tests {
             GammaPrior::Pooled,
             true,
             false,
+            crate::spectral::DEFAULT_PROJ_THRESHOLD,
             &mut rng,
         );
 
--- a/src/python/mod.rs
+++ b/src/python/mod.rs
@@ -6427,9 +6427,15 @@ impl CTM {
     /// `em_tol` is the relative-bound tolerance for EM early stopping — the run
     /// stops when the relative change in the variational evidence bound falls below
     /// it (the criterion R `stm` uses).
+    /// `spectral_projection_threshold` is the vocabulary size above which the
+    /// spectral (anchor-word) init switches from the exact V*V co-occurrence to a
+    /// deterministic random projection. The default (10000) matches R `stm`, which
+    /// stays exact up to that size; lower it to force the cheaper approximate path
+    /// on smaller vocabularies. It only affects `init="spectral"` runs.
     #[pyo3(signature = (data, *, iters=500, convergence_tol=1e-5, inference="batch",
                         batch_size=256, tau=64.0, kappa=0.7, beta_init=None, em_tol=None,
-                        keep_eta_cov=true, num_threads=None))]
+                        keep_eta_cov=true, num_threads=None,
+                        spectral_projection_threshold=spectral::DEFAULT_PROJ_THRESHOLD))]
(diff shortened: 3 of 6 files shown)


nealcaren:
Dual review: Claude (faithfulness) + **Gemini `3.1-pro-high`** (adversarial, against the PR worktree). Both confirm the change is faithful and correctly plumbed: branch direction is `V ≤ threshold → exact Gram`, `V > threshold → random projection`; the default is consistent at 10000 across the rust const, both bindings, conformance, and stub; the exact/projected paths, `PROJ_DIM`, and `PROJ_SEED` are untouched (determinism preserved); and there's no gold shift (every spectral-touching gold is < 3000 vocab, so exact either way). Raising 3000→10000 removes topica's spurious divergence from stm in the mid-vocab band. Non-blocking notes: (1) the `spectral_init` wrapper default also reaches LDA/STS/DTM `init="spectral"` (not just STM/CTM) — same direction (more faithful), no gold shift, those models simply don't expose the kwarg; (2) V∈(3000,10000] now builds the dense O(V²) Gram (matches stm, more RAM — opt out with `spectral_projection_threshold=3000`). CI green.

Issue #555, opened 2026-07-26
reducer="umap" (now the BERTopic/Top2Vec default) collapses to ~2 topics on real embeddings: UMAP layout stage, not HDBSCAN/init/epochs


nealcaren:
## Summary

Since #549/#550 made `reducer="umap"` the **default** for `BERTopic` and `Top2Vec`
(0.50.0), the default embedding-clustering path is driven by topica's in-house Rust
UMAP (`src/umap.rs`). On real sentence embeddings that reducer is **unstable and
collapse-prone**: for moderate corpora it produces a layout that HDBSCAN merges into
**~2 topics on the large majority of seeds**, where the reference `umap-learn` reliably
recovers dozens — under *either* spectral or random init. The cluster structure survives
topica's reduction (KMeans and trustworthiness match `umap-learn`); what is missing is the
**density separation HDBSCAN needs**. The defect is in topica's UMAP **layout
optimization**, not in HDBSCAN, c-TF-IDF, initialization, or the epoch budget. I have
isolated the stage but **not** the precise numerical cause.

This is a correctness/robustness issue for the default surface, not a tuning nit.

> **Correction note (supersedes an earlier draft of this issue): the cause is NOT random
> initialization.** `umap-learn` with `init="random"` separates fine here (14–33 topics
> every seed), and raising topica's `n_epochs` to 2000 does not reliably help (it improves
> one seed, *worsens* another). Both hypotheses are falsified below. The prior #343
> ablation ("init irrelevant") holds; the divergence is elsewhere in the layout SGD.

## Reproduction

20 Newsgroups (train, headers/footers/quotes stripped), `all-MiniLM-L6-v2` embeddings,
matched config throughout: UMAP `n_neighbors=15, n_components=5, min_dist=0.0,
metric="cosine"`; HDBSCAN `min_cluster_size=10, euclidean, eom`. The *same* `hdbscan`
library clusters every reduction, so any difference is attributable **only** to the
reducer.

### 1. End-to-end BERTopic parity (N=800), stage-decomposed

| stage | topics (topica/ref) | noise | ARI | top-word Jaccard |
|---|---|---|--:|--:|
| **A** end-to-end: `topica.BERTopic(reducer="umap")` vs `bertopic` pkg | **2 / 13** | 0% / 12% | **0.069** | 0.53 |
| **B** HDBSCAN-only: topica clusters `umap-learn`'s reduction directly | 13 / 13 | 15% / 12% | **0.939** | 0.70 |

[...]

PR #558, opened 2026-07-26
perf(bertopic): sparse + parallel approximate_distribution, bit-identical (#557)

### LINKED ISSUE #557: perf: BERTopic.fit approximate_distribution is O(docs·windows·topics·V), single-threaded — dominates fit on medium corpora
## Summary

`topica::bertopic::approximate_distribution` (src/bertopic.rs:433) is the wall-clock
bottleneck of `BERTopic.fit` on medium/large corpora. Native-stack sampling of a
`topica.BERTopic(reducer="umap").fit(...)` on **5000 congressional bills** (MiniLM,
mcs=10) showed **99% of one core sitting in `approximate_distribution` for 70+ minutes** —
the UMAP reduce, HDBSCAN, and c-TF-IDF were all comparatively trivial. It is called
**unconditionally** in `fit_bertopic` (src/bertopic.rs:264), so every `.fit()` pays it.

Discovered while validating the #555 UMAP fix on a second dataset (the correctness fix is
unrelated and already merged in #556).

## Why it is slow

Per document, per sliding window (default `window=4, stride=1` ⇒ ≈ one window per token,
i.e. hundreds of windows for a long bill), the inner loop:

1. allocates **two dense vocab-length (`V`) vectors** (`counts`, `vec`) even though a
   window has ≤ `w` nonzero tokens (src/bertopic.rs:458-471);
2. calls `cosine(&vec, row)` against **every one of `num_topics` topic rows**, each an
   **O(V)** dense pass (src/bertopic.rs:472-477);
3. `cosine` (src/bertopic.rs:498) **recomputes the topic row's norm `nb` on every call**,
   though it is constant across all windows and documents.

Total cost ≈ **O(n_docs · windows_per_doc · num_topics · V)**, entirely single-threaded.
On the congress run that is 5000 · (hundreds) · ~111 · V.

## Suggested fixes (all determinism-preserving)

- **Sparse windows.** A window has ≤ `w` nonzero tokens; compute each cosine's dot and the
  window-vector norm from just those nonzeros (ascending token order ⇒ the accumulation
  sequence of nonzero terms is identical to the dense pass, so results stay **bit-identical**).
  Cuts the per-window-per-topic cost from O(V) to O(w).
- **Precompute topic-row norms once** instead of inside `cosine` per call.
- **Parallelize the outer `docs.iter()` with rayon** (`par_iter`) — each document is
  independent; collecting in order preserves determinism.
- Optionally gate the whole step behind a `calculate_distribution=`-style flag (upstream
  BERTopic computes `approximate_distribution` only on explicit request), but the default
  should be fast enough not to need opting out.

Expected: congress 5000-doc fit from ~70 min to seconds, with a bit-identical `doc_topic`.

## Repro

`topica.BERTopic(reducer="umap", min_cluster_size=10).fit(docs, emb)` on ~5000 medium-length
documents; sample the process (`sample <pid>`) and observe the stack pinned in
`approximate_distribution`.

nealcaren:
## Summary

`approximate_distribution` (`src/bertopic.rs`) dominated `BERTopic.fit` wall-clock on medium corpora. Native-stack sampling of a **5000-doc congress fit** showed **99% of one core in this function for 70+ minutes**. It is called unconditionally by `fit_bertopic`, so every `.fit()` paid it. (Found while validating the #555 UMAP fix on a second dataset — unrelated to that correctness fix.)

## Why it was slow

Per document, per sliding window (default `window=4, stride=1` ⇒ ≈ one window per token), the inner loop:
1. built two **dense vocabulary-length (`V`)** vectors for a window with ≤4 nonzero tokens;
2. ran a full **O(V)** `cosine` against **every** topic row;
3. recomputed each topic row's L2 norm **on every one of those calls**.

Total: **O(docs · windows · topics · V)**, single-threaded.

## Rewrite

- Keep only the window's nonzero `(token, weight)` pairs → per-window cosine drops from O(topics·V) to **O(topics·w)**.
- Precompute each topic row's norm **once**.
- Parallelize the outer document loop with **rayon** (each doc independent; ordered collect preserves determinism).

## Bit-identical

Iterating the window's nonzero tokens in **ascending index order** reproduces the exact floating-point accumulation of the dense `t=0..V` loop (skipped entries add exactly `0.0`); the topic-row norm is summed in the same order; zero-norm windows/rows contribute `0.0` (skipped) exactly as `cosine` returned `0.0` and failed the `sim > 0` gate before. So `doc_topic` is unchanged to the last bit.

New regression test `approximate_distribution_bit_identical_to_dense` asserts the new impl equals the **verbatim pre-#557 dense reference** across OOV tokens (≥V), repeated tokens, `reduce_frequent` on/off, a nonzero `min_similarity` gate, and varied window/stride. The two existing `approximate_distribution` tests pass unchanged.

## Result

| | before | after |
|---|--:|--:|
| congress 5000-doc `BERTopic.fit` (umap, seed 0) | ~72 min | **5.5 s** |
| `doc_topic` | — | bit-identical |

## Process

Diagnosis + plan independently reviewed by a cross-family (Gemini pro) pass, which confirmed the bit-identity argument (IEEE-754 `x + 0.0 == x`, ascending-order accumulation, rayon order-preservation) and **rejected two alternative algorithms** (incremental sliding window; per-token precomputed contributions) as non-bit-exact and no faster.

## Verification
- Full gate green: `./scripts/preflight.sh` + `3876 passed` pytest + `cargo test --lib`.
- Kept the eager-at-fit contract (no `calculate_distribution` flag): the sparse+parallel version is cheap enough that the default needs no opt-out, and `doc_topic` is a required cross-model `Estimator` surface.

Closes #557

### DIFF
--- a/src/bertopic.rs
+++ b/src/bertopic.rs
@@ -31,6 +31,7 @@
 //! does not embed the text.
 
 use crate::{cluster, reduce, represent};
+use rayon::prelude::*;
 
 /// A fitted BERTopic model. Exposes the branch's shared surface: `topic_word`
 /// (K x V, normalized c-TF-IDF), `doc_topic` (D x K, the approximate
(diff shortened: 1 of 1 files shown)


nealcaren:
## Adversarial review (Codex, gpt-5.6-sol, high effort)

Ran an adversarial pass tasked with *breaking* the bit-identity claim. Verdict: **"Within the intended finite K×V invariant, I found no actual differing bit."** It surfaced one real regression and some hardening asks, all addressed in the amended commit:

**Fixed — real regression:** `Vec::with_capacity(window)` could panic with capacity-overflow for a huge public `window` (constructor / `.approximate_distribution()` / Python binding); the dense path allocated `V`-sized buffers and never had this. Now bounded by `w.min(doc.len())`. Verified a `2^63` window returns cleanly via the Python binding.

**Fixed — test gaps:** the bit-identity regression test now includes empty docs, an all-OOV doc, a zero-norm topic row, an oversized window (10000), the exact `min_similarity` boundary, and `num_topics == 0`.

**Fixed — claim wording:** the "adds exactly 0.0" comment now scopes the bit-identity guarantee to a fitted model (finite rows, `num_topics` rows each of length `V = idf.len()`); a `debug_assert` documents that invariant. The only way to violate it is ragged/non-finite hand-constructed input, unreachable from a fitted model.

Full gate re-run green (preflight + 3876 pytest + rust lib); the 3 `approximate_distribution` Rust tests pass.

Issue #560, opened 2026-07-26
STS default fit is not a faithful port of R STS.R init (prevalence η / Σ⁻¹ approximated, not STM-spectral); validate against the faithful profile


nealcaren:
## Summary

The STS cross-implementation validation leg (`parity/sts_r_compare.py`) deterministically
**fails**: topica-STS aligns to the authors' R reference (poliblog RDS) at **0.849**, tripping
the relative gate `sts_vs_ref > stm_vs_ref − 0.1` (0.849 vs 0.859). The docstring's expectation
is "mid-0.90s," and the leg was green through 0.34.0. This surfaced during the 0.52.0 paper
regeneration; it is **not** caused by 0.52.0.

## Bisect (each point = rebuild + full R+STS+STM run)

| Point | topica-STS ↔ R-STS | topica-STM ↔ R-STM |
|---|--:|--:|
| v0.34.0 | **0.930** ✅ | 0.976 |
| `cd59840` (parent of #474) | **0.930** ✅ | 0.976 |
| current (0.52.0) | **0.849** ❌ | 0.959 |

**Root cause: `c4272f4` — "feat(sts): reference-fidelity fitting profiles (paper + CRAN) (#454) (#474)".**
It is the only substantive change to `src/sts.rs` in the window (the other, #413, is determinism
plumbing). Ironically, the commit intended to *improve* reference fidelity dropped R-alignment by
~0.08. STM alignment did **not** rise (0.976→0.959), so this is an STS-side regression, not the
relative gate tightening.

The parity fit call is byte-identical across the window
(`STS(num_topics=K, init="spectral").fit(..., iters=50, kappa_estimation="lasso")`), so the
regression is a **default-behavior change inside `.fit()`**. #474 added a `reference: str = "none"`
profile parameter and, on the default `"none"` + `"lasso"` path the harness uses, introduced
phi-mass-weighted sentiment aggregation and half-step κ damping.

## Mechanism (measured)

The misalignment concentrates in the topics where topic signal is "parked in the sentiment
direction": on poliblog K=5, topics 1 & 3 have sentiment latent mean ≈0.20 with std ≈0.01, and
for those `cos(β@mean, β@neutral) = 0.011` — the at-mean representative is **hypersensitive**, so
topica's and R's slightly different κ_t/κ_s splits land in visibly different topic-words, dragging
the 5-topic average from ~0.95 to 0.849. The extraction is correct (`topic_word_at(level)` =
[...]

PR #561, opened 2026-07-26
fix(sts): faithful STS.R initialization; default reproduces R (#560)

nealcaren:
## Summary

topica's `STS` default fit diverged from the Chen & Mankad (2024) reference, aligning to the authors' poliblog fit at a topic-word cosine of **0.849** where the reference expects the mid-0.90s — the validation leg (`parity/sts_r_compare.py`) had begun failing. This makes the default fit a **faithful port** of R's `STS.R`, so `topica.STS(...).fit(...)` reproduces the reference out of the box.

## Root cause (verified against the R source)

The reference-init path was not faithful. It *approximated* the prevalence latents from a phi log-ratio and built Σ from the init-eta variance. I dumped what R's `STS.R` actually seeds from — `stm(..., max.em.its = 0)` on the poliblog corpus — and it returns **`eta = 0`** and **`invsigma = diag(1/20)`** exactly. So the faithful init is simply: prevalence latents at **0**, Σ = **diag(20)** across all `2K−1` dims, sentiment latents = centered seed, then the κ warm-start. The old default also used Σ=identity (variance 1, not 20) and the topica-native **ridge** κ estimator — R uses **lasso**.

## Changes
- **`src/sts.rs`** — `reference_init` rewritten to R's exact seed (η=0, Σ=diag(20)); phi-approximation removed.
- **`src/python/mod.rs` / `.pyi`** — the default fit now uses the reference init, and the default κ estimator is **`lasso`** (R `STS.R`'s default) instead of ridge. Ridge stays an opt-in (`kappa_estimation="ridge"`) for large-K regimes. Docstrings corrected.
- **`parity/sts_r_compare.py`** — validates against the faithful profile.

## Result (real R, end-to-end)

| | before | after |
|---|--:|--:|
| default `topica.STS(...)` ↔ R-STS | 0.849 | **0.942** |
| topica-STM ↔ R stm (baseline) | 0.959 | 0.959 |
| leg status | FAIL | **OK** |

The alignment is a *consequence* of replicating R's init, not tuning to the threshold. On this corpus lasso is also ~3× faster than ridge (144s vs 464s/fit), so the faithful default is faster here too.

## Verification
- 302 Rust tests + 29 Python STS tests pass; `./scripts/preflight.sh` green.
- Diagnosis is grounded in the reference `Code/STS.R`, `opt.kappa.R`, and a direct dump of `stm(max.em.its=0)$eta`/`$invsigma`.

## Notes / follow-ups
- Corrects the record on #560: an earlier bisect wrongly blamed #474 (which is innocent — tested at 0.930). The exact commit that regressed 0.930→0.849 is a later non-STS change and is now moot, since the default is faithful by construction.
- The paper's `parity`-generated validation files need one regen to pick up the fixed STS + the 0.52.0 stamp (separate).

Refs #560

### DIFF
--- a/parity/sts_r_compare.py
+++ b/parity/sts_r_compare.py
@@ -202,9 +202,14 @@ def run(verbose: bool = True) -> dict:
     K = r_sts.shape[0]
     X = rating.reshape(-1, 1)
     sts = STS(num_topics=K, init="spectral")
+    # reference="paper" replicates Chen & Mankad's STS.R init faithfully: prevalence
+    # latents at 0 and Σ = diag(20) (their stm(max.em.its=0) returns eta=0,
+    # invsigma=diag(1/20)), the "lasso" (plain group-mean) κ estimator, and the
+    # centered-seed sentiment init. This is the profile the reference RDS was produced
+    # with, so it is what we validate against.
     sts.fit(docs, sentiment_seed=rating.tolist(), prevalence=X,
             prevalence_names=["rating"],
-            iters=50, kappa_estimation="lasso")  # match the reference's kappa estimator
+            iters=50, kappa_estimation="lasso", reference="paper")
     stm = STM(num_topics=K, init="spectral")
     stm.fit(docs, X, prevalence_names=["rating"], iters=80)
 
--- a/python/topica/_topica.pyi
+++ b/python/topica/_topica.pyi
@@ -916,17 +916,20 @@ class STS:
 
         EM stops once the relative change in the variational bound falls below
         convergence_tol or after iters iterations. kappa_estimation chooses the
-        topic-word estimator: None (default) uses the topica-native "ridge" unless a
-        reference profile overrides it; "ridge" (fast, topica-native; kappa_ridge
-        sets the ridge), "lasso" (an L1 Poisson path with AIC-selected penalty),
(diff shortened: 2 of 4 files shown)


nealcaren:
## Adversarial review (Gemini 3.1 pro, high effort)

Tasked with breaking the faithfulness claim. Verdict: **the core fix is faithful for all corpora, not just poliblog.** It independently confirmed the key reasoning — at `stm(max.em.its=0)` zero EM iterations run, so `eta` is all-zero and no M-step computes a sample covariance; `invsigma` is returned as the hardcoded prior `diag(1/20)` regardless of the corpus. So Σ=diag(20) is universal. It also confirmed the `unwrap_or("lasso")` default interacts correctly with the `user_set_kappa` conflict checks (no spurious errors), and found no determinism/NaN/dimension risk (diag(20) is better-conditioned than the old log-ratio approximation).

One finding — `reference_init=false` (Σ=identity) is unreachable from Python since every profile sets it true. It is **not dead code** (the Rust `sts_conforms` test exercises the `false` branch), and disabling it from Python is **intentional**: Σ=identity is a strictly less faithful init with no principled use for an STS port. Added a comment at the profile match documenting this, per the reviewer's "explicitly documented" option. No behavioral change.

Merge recommendation from the review: approve after documenting the above (done).

Issue #569, opened 2026-07-27
Tracking: multithreading coverage across the model roster (only LDA/STM/CTM/keyATM thread today)


nealcaren:
## Summary

Only **4** models expose a threading knob today: `LDA`, `STM`, `CTM`, `keyATM`
(all via `num_threads`/`fit`). The other ~40 run single-threaded. This tracks
extending coverage, grouped by the mechanism that applies to each family. Child
issues hold the concrete work.

## By mechanism

**1. Collapsed Gibbs — AD-LDA partition-and-merge (reuse LDA's path)** → #566
`DMR`, `GDMR`, `LabeledLDA`, `PA`, `SeededLDA`, `GSDMM`, `BTM`.
Also Gibbs but harder / lower priority: `PT`, `PLTM`, `PolylingualLDA`, `DiscLDA`,
`AnchorLDA`, `HDP` (dynamic K), `HLDA` (nested CRP), `TensorLDA`.

**2. Variational EM — per-document E-step (STM/CTM-style knob)** → #567
`SupervisedLDA` (E-step already `par_iter`, needs a knob; Gibbs backend serial).
Also: `RTM`, `STS`, and the DMR/GDMR λ data term.

**3. Matmul / BLAS-bound** → #568
`NMF`, `LSA`. Also `SAGE` (optimization matmuls).

**4. Hand-coded Rust VAE / neural (batched matmul, decoder GEMM)** → #378
`ProdLDA`, `ETM`, `DETM`, `CombinedTM`, `ZeroShotTM`, `Scholar`, `InfoCTM`, `TBIP`,
`IdealPointTM`, `IdealPointSentenceTM`, `Wordfish`, `PartyEmbeddings`,
`EmbeddingLDA`, `NarrativeTM`.

**5. Embedding-cluster / LLM — threading lives in the external stack, not topica core**
`BERTopic`, `Top2Vec` (UMAP/HDBSCAN thread counts), `FASTopic` (Sinkhorn matmul),
`TopicGPT` (LLM calls, already IO-bound). Lower priority for topica-side threading.

## Notes

- Context: found while adding a threads axis to the full-model benchmark. Without a
  knob, reference implementations (tomotopy `workers`, sklearn/torch via BLAS) thread
  while topica does not, so "single-threaded" comparisons were really
  topica-1-core vs reference-multicore until the benchmark started pinning threads.
- The determinism contract from LDA should hold everywhere: `num_threads=1` stays
[...]

Issue #568, opened 2026-07-27
Perf: parallelize NMF/LSA matmuls (single-threaded vs sklearn BLAS)


nealcaren:
## Summary

`NMF` (multiplicative updates) and `LSA` (truncated SVD) run their dominant matmuls
single-threaded (the Rust `matrixmultiply` path), while the `scikit-learn`
references thread through BLAS (OpenBLAS/MKL). In the full-model benchmark this is a
topica-1-core vs sklearn-multicore comparison for these two models.

Options:
- Parallelize the dominant GEMMs (rayon-blocked matmul, or a threaded GEMM backend).
- And/or expose `num_threads` so the parallelism is controllable and pinnable.

Related: the VAE decoder GEMM idea in #378.

## Why

sklearn NMF/`TruncatedSVD` have no `n_jobs`; their only parallelism is BLAS threads.
topica has neither a BLAS-backed matmul nor a threads knob here, so it cannot match
even single-vs-single once BLAS is uncapped. A parallel matmul would close the gap.

## Acceptance

NMF/LSA get a multithreaded matmul path with a measurable speedup at larger K/V,
deterministic results, and (ideally) a `num_threads` knob.

nealcaren:
**Recon note (from the #569 roadmap pass):** the parallelize-the-matmul half is largely already done — the NMF multiplicative-update matmuls are rayon-parallelized over independent output rows (`src/nmf.rs:92`), and LSA's randomized SVD reuses that same rayon+sparse path (`src/lsa.rs:242`, which already has a 1-thread-vs-many-thread pool test). So the dominant GEMMs already thread on the global pool.

The remaining work is the **knob**: expose `num_threads` on NMF/LSA via `run_with_threads` (`src/python/mod.rs:113`) so the parallelism is bounded and pinnable — which is what makes the benchmark honest (topica vs sklearn/BLAS at matched thread counts) rather than topica-uncapped vs reference-capped. Small.

Issue #567, opened 2026-07-27
Perf: expose num_threads for SupervisedLDA (variational E-step already parallel; Gibbs backend serial)


nealcaren:
## Summary

`SupervisedLDA`'s variational EM (`fit_slda`) already runs the per-document E-step
with `rayon` `par_iter` on the global thread pool, but there is **no `num_threads`
knob** to bound or configure it the way `STM`/`CTM` expose one. Add `num_threads`
to `SupervisedLDA` so the E-step parallelism is controllable (and so a benchmark can
pin it to 1 for a fair single-thread comparison).

Separately, the new `inference="gibbs"` backend (added in #565) samples **serially**.
Give it AD-LDA-style parallel sampling as well (see the Gibbs-count-model issue).

## Why

Without a knob, the variational E-step silently uses all cores, which makes
single-threaded benchmarking dishonest and leaves no way to tune it. The Gibbs
backend has no parallelism at all.

## Acceptance

`SupervisedLDA(num_threads=n)` controls the variational E-step; `num_threads=1` is
deterministic and matches today's serial result numerically; the Gibbs backend
gains a parallel path that stays deterministic for fixed `num_threads`+`seed`.

nealcaren:
**Recon note (from the #569 roadmap pass):** the variational E-step is already parallel — `fit_slda` runs the per-document E-step with `par_iter` on the global rayon pool (`src/slda.rs:304`). So this is a *knob* task, not a parallelization task: wrap the fit in `run_with_threads(num_threads, ...)` (`src/python/mod.rs:113`, the same helper STM/CTM use) and add a `num_threads` constructor/fit param to bound and pin it. Small, mechanical.

The `inference="gibbs"` backend (#565, `fit_slda_gibbs`) is still serial — that half is AD-LDA partition-and-merge and folds into the #566 count-model pattern (shared `reconcile_worker_outs` from #570).

Issue #566, opened 2026-07-27
Perf: multithread the collapsed-Gibbs count models (DMR, GDMR, LabeledLDA, PA, SeededLDA, GSDMM, BTM)


nealcaren:
## Summary

`LDA` supports MALLET-style approximate parallel Gibbs (AD-LDA) through
`num_threads`: `fit()` partitions documents across workers, samples each partition
against a per-worker copy of the count tables, and merges, staying deterministic
for a fixed `num_threads`+`seed`. The sibling collapsed-Gibbs count models reuse
the same `sample_doc`/count-table machinery but sample serially, with no
`num_threads` knob.

Extend the AD-LDA partition-and-merge to these models:

- **DMR / GDMR** (highest priority; both are benchmarked, and GDMR is a pure-Python
  wrapper over DMR so it inherits the fix). The per-sweep token sampling parallelizes
  directly; the periodic L-BFGS λ optimization can keep its data term parallel too.
- **LabeledLDA** (labeled sparse sweep)
- **PA** (Pachinko, per-document)
- **SeededLDA** (seeded sparse sweep)
- **GSDMM** (short-text collapsed Gibbs)
- **BTM** (biterm)

Out of scope here (harder, dynamic structure): HDP (changing K), HLDA (nested CRP).

## Why

Today these run single-threaded while their references (tomotopy `workers=`, MALLET
threads) scale across cores, so the benchmark compares topica-1-core to
reference-multicore. LDA's threading already gives ~3.5x at 8 threads on poliblog;
the same win should carry to these models with a shared implementation.

## Acceptance

`num_threads>1` gives a wall-clock speedup, remains deterministic for fixed
`num_threads`+`seed`, and `num_threads=1` stays byte-identical to today.

nealcaren:
**PR #570** lands the first slice: **DMR + GDMR** threading (the two highest-priority, benchmarked targets).

The reusable groundwork is done — LDA's exact merge is now a shared `reconcile_worker_outs`, and `parallel_sweep_dmr_batched` shows the pattern for a per-document-prior sweep. The remaining count models (**LabeledLDA, PA, SeededLDA, GSDMM, BTM**) each reuse that same helper; the only per-model work is wiring their own worker sampler + the `num_threads` knob.

Determinism contract holds (num_threads=1 byte-identical; fixed num_threads>1 deterministic); ~3.65x at 8 threads on DMR, matching LDA. Details in #570.

nealcaren:
[...]

PR #570, opened 2026-07-27
feat(dmr): AD-LDA multithreading via num_threads (DMR + GDMR) (#566)

nealcaren:
First slice of the multithreading roster (#569): extends LDA's MALLET-style approximate-parallel Gibbs (AD-LDA partition-and-merge) to **DMR** and, for free, **GDMR** (a pure-Python DMR wrapper). These were the highest-priority targets in #566 — both are benchmarked, and DMR reuses LDA's exact `sample_doc`/count-table machinery, so the threading is a direct reuse of a proven, determinism-safe path.

## What changed

- **Shared reconciliation.** Extracted the exact merge from LDA's `parallel_sweep_batched` into `reconcile_worker_outs` (`final = Σ_workers − (W−1)·original` for the topic-word + tokens-per-topic counts, plus the doc-topic writeback). The DMR-family and plain-LDA sweeps now differ *only* in how workers sample, not in how results reconcile — this is the shared helper the remaining #566 count models (LabeledLDA/PA/SeededLDA/GSDMM/BTM) will reuse.
- **`parallel_sweep_dmr_batched`.** Workers sample their document slice with `dmr::run_sweep_dmr` against private count tables, passing the matching `doc_alpha` slice (the only DMR difference: α varies by document), then reconcile once per sweep.
- **`num_threads` on DMR** — constructor + per-call `fit(num_threads=)` override + `settings`. Both the training loop and the posterior-sampling loop take the parallel path when `>1`, with per-sweep seeds derived from `seed_base` (non-overlapping across phases) so a fixed `num_threads`+`seed` is deterministic. `num_threads=1` is the exact serial path, byte-identical to before. Ignored by the warp/cvb0 backends.
- **GDMR** forwards `num_threads` through construct/fit and save/load; `.pyi` stub updated.

## Determinism contract (identical to LDA)

- `num_threads=1` reproduces **byte-for-byte** (serial path unchanged; new code is guarded behind `>1`).
- a fixed `num_threads>1` is **deterministic** across runs (same threads+seed → identical `topic_word`/`feature_effects`).
- `num_threads` across *different* thread counts intentionally differs — that is the AD-LDA approximation, same as LDA.

## Performance

Measured on 6k docs, V=500, K=20, 150 iters:

| threads | 1 | 2 | 4 | 8 |
|---|--|--|--|--|
| time | 5.36s | 3.29s | 2.20s | 1.47s |

~3.65x at 8 threads, in line with LDA's ~3.5x.

## Tests

- New `tests/test_dmr_parallel.py`: serial reproducibility, fixed-thread determinism at 2/4/8, zero-thread clamp, `fit()` override, doc_topic simplex + shape invariants, covariate-structure recovery, GDMR passthrough.
- 248 Rust + 281 DMR/GDMR Python + 175 LDA-threading regression tests pass; `./scripts/preflight.sh` green (fmt, clippy, stub sync, model tables).

Part of #569. Follow-ups: the same shared helper extends to the other count models in #566, then #567 (SupervisedLDA), #568 (NMF/LSA), #378 (VAE).

### DIFF
--- a/python/topica/_topica.pyi
+++ b/python/topica/_topica.pyi
@@ -299,6 +299,7 @@ class DMR:
         alpha_epsilon: float = 1e-10,
         lbfgs_iters: int = 20,
         sampler: str = "sparse",
+        num_threads: int = 1,
     ) -> None:
         """Create an unfitted DMR model. prior_variance is the Gaussian prior
         variance on the feature weights; lbfgs_iters caps L-BFGS steps per round.
@@ -312,7 +313,13 @@ class DMR:
         compute. As with plain LDA, warp is flat in K and overtakes sparse on
         speed around K~=50 (dominating at large K), while sparse keeps a small-K
         coherence edge and a convergence trace; the "warp"/"cvb0" paths record no
-        convergence trace, so convergence_tol has no effect there."""
+        convergence trace, so convergence_tol has no effect there.
+
+        num_threads > 1 runs MALLET-style approximate-parallel Gibbs on the
+        default sparse backend (partition documents, sample against per-worker
+        count copies, merge; deterministic for a fixed num_threads+seed); 1 is the
+        exact serial path. It is ignored by the warp/cvb0 backends and can be
+        overridden per call via fit(num_threads=)."""
         ...
 
     def fit(
@@ -332,6 +339,7 @@ class DMR:
         check_every: int = 10,
         covariates: Optional[numpy.typing.NDArray[numpy.float64]] = None,
         offset: Optional[numpy.typing.NDArray[numpy.float64]] = None,
+        num_threads: int | None = None,
     ) -> "DMR":
         """Fit by collapsed Gibbs with the per-document Dirichlet prior
         alpha_{d,t} = exp(lambda_t . x_d + offset[d, t]). `features` (or
--- a/python/topica/gdmr.py
+++ b/python/topica/gdmr.py
@@ -269,6 +269,11 @@ class GDMR:
     sampler:
         Gibbs sampler variant: ``"sparse"`` (default), ``"warp"``, or
         ``"cvb0"``.  See ``topica.DMR`` for details.
+    num_threads:
+        Worker count for the inner ``DMR`` sparse Gibbs fit.  ``> 1`` enables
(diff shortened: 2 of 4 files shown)


nealcaren:
## Dual review — both APPROVE

**Accuracy review (Claude subagent, static): FAITHFUL** — no correctness defects.
Verified: `reconcile_worker_outs` extraction is behavior-preserving (LDA path semantically unchanged; the removed `let v` was reconcile-only and is recomputed identically); `parallel_sweep_dmr_batched` slices docs/doc_alpha/doc_topics on the same `(start,end)` and `run_sweep_dmr` indexes all three by the same local `doc_idx` (no misalignment); workers share no mutable state; `num_threads=1` never advances differently from the pre-PR serial `rng` (byte-identical); train seeds (`iter∈1..=iters`) and posterior seeds (`post_sweep∈iters+1..`) are disjoint (odd golden-ratio multiplier ⇒ bijection ⇒ no collision); merge is integer-only so threaded runs reproduce for fixed num_threads+seed; both sweep sites switched; λ/SE/θ bookkeeping reads a reconciled model; wiring complete incl. GDMR save/load `.get("num_threads",1)` back-compat; `partition_ranges` caps at `num_docs` (no empty-partition panic).

**Adversarial review (Google Gemini 3.1-pro, `agy`, outsider/cross-family): SHIP** — no defects.
Independently confirmed the same: determinism + disjoint seed sets, `sort_unstable_by` on packed entries is a total order (low bits = unique topic index), `i64` accumulation with non-positive entries filtered before the `u32` cast (no underflow), over-threading/empty-corpus safe, LDA extraction verbatim, downstream consumers see a fully reconciled state at `batch=1`.

**Non-blocking nits (both reviewers, cosmetic):** the `batch` param is always `1` for DMR (dead flexibility mirroring LDA's turbo path); on early convergence `post_sweep` reserves seed space against requested `iters` (harmless — no collision); the `parallel_sweep_dmr_batched` docstring says "DMR-family count models" though DMR is the only current caller (forward-looking to the rest of #566). None warrant a change.

Merging once CI (test matrix incl. windows-latest) is green.

PR #571, opened 2026-07-27
feat(labeledlda): AD-LDA multithreading via num_threads (#566)

nealcaren:
Second count model in the #569/#566 threading roster: extends the AD-LDA partition-and-merge (shared `reconcile_worker_outs` from #570) to **LabeledLDA**.

## The one wrinkle vs DMR

DMR reused a raw-slice sweep (`run_sweep_dmr` over `&mut [Vec<u32>]`). LabeledLDA's restricted sweep (`labeled::run_sweep_labeled`) instead operates on a whole `TopicModel` via its packed-table methods and indexes `model.doc_topics[d]`. So each worker gets a **lightweight `TopicModel`** carrying a clone of the shared count tables plus only its *slice* of `doc_topics` — indices then align with `docs[start..end]` / `allowed[start..end]`, and `run_sweep_labeled` is reused **unchanged**. Merge is the shared `reconcile_worker_outs`.

## Correctness under the label constraint

The additive merge (`final = Σ_workers − (W−1)·original`) stays exact: a worker touches only its own documents' tokens relative to the snapshot, and a topic disallowed for every doc in a worker's slice contributes a **zero net delta** — so `tokens_per_topic` can never underflow. Plan reviewed by **Gemini before implementation** (verdict PLAN SOUND, with the merge-correctness and no-underflow proofs).

## Wiring

`num_threads` on the constructor (default 1), `fit()` override, `settings`, and `load` (→1). Both the training and posterior-sampling sweep sites dispatch serial-vs-parallel with disjoint per-sweep seeds (`iter` / `post_sweep`). `num_threads=1` is byte-identical serial; **ignored by the cvb0 backend**.

## Determinism + performance

- `num_threads=1` reproduces byte-for-byte; a fixed `num_threads>1` is deterministic (phi and theta).
- ~**3.7x at 8 threads** (20k docs, K=12, incl. unconstrained docs): 4.22s → 1.14s.

## Tests

New `tests/test_labeled_parallel.py`: serial reproducibility, fixed-thread determinism at 2/4/8, zero-clamp, `fit()` override, doc_topic simplex, labeled+unconstrained mix, and — the key invariant — **the label constraint holds under threading** (zero out-of-label theta mass at every thread count).

248 Rust + full Python suite (3908 passed, 32 skipped) green; `./scripts/preflight.sh` clean.

Part of #569. Next: SeededLDA → BTM → GSDMM, then the #567/#568 knob PRs.

### DIFF
--- a/python/topica/_topica.pyi
+++ b/python/topica/_topica.pyi
@@ -1839,7 +1839,7 @@ class LabeledLDA:
 
     def __init__(
         self, *, alpha: float = 0.1, beta: float = 0.01, seed: int = 42,
-        sampler: str = "sparse",
+        sampler: str = "sparse", num_threads: int = 1,
     ) -> None:
         """Create an unfitted model. alpha is the symmetric per-topic prior.
 
@@ -1849,7 +1849,13 @@ class LabeledLDA:
         off the allowed topics). CVB0 enforces the same supervised constraint
         deterministically and tends to higher coherence; it produces no MCMC
         theta_draws. (WarpLDA is not offered here: masked proposals mix poorly,
-        whereas masking is free in CVB0.)"""
+        whereas masking is free in CVB0.)
+
+        num_threads > 1 runs the sparse backend as MALLET-style approximate-parallel
+        restricted Gibbs (partition documents, sample against per-worker count
+        copies, merge; deterministic for a fixed num_threads+seed); 1 is the exact
+        serial path. It is ignored by the cvb0 backend and can be overridden per
+        call via fit(num_threads=)."""
         ...
 
     def fit(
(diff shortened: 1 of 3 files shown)


nealcaren:
## Dual review — both APPROVE

**Accuracy review (Claude subagent, static): FAITHFUL** — no correctness defects.
Verified: the per-worker `TopicModel` struct literal populates all 11 fields with no `topic_mask`/`topic_bits` transposition and both tables cloned from the pre-sweep snapshot; slice alignment is exact (`docs`/`allowed`/`doc_topics` all `[start..end]`, local index `d`→global `start+d`, writeback `[start..end)`); the additive merge is model-agnostic so label constraints only restrict destination topics (zero net delta for a topic disallowed across a partition); no `u32` underflow (a token is decremented only for a topic it currently occupies, so `tokens_per_topic[old]≥1`); `num_threads=1` byte-identical serial with the outer RNG untouched when threaded; train (`iter∈1..=iters`) and posterior (`iters+1..`) seeds disjoint (odd multiplier ⇒ bijection); wiring complete incl. cvb0 returning before any parallel sweep; edge cases (threads>docs, unconstrained/empty docs) safe.

**Adversarial review (Google Gemini 3.1-pro, `agy`, outsider/cross-family): SHIP** — "clean, robust, and ready to ship." Independently confirmed struct-literal correctness, slice alignment, merge exactness + underflow protection, determinism/seed disjointness, cvb0 isolation, and that `reconcile_worker_outs` (shared with LDA/DMR) is untouched.

**Non-blocking nits (both, cosmetic):** `batch` is hardcoded to 1 (turbo path unexercised, mirrors DMR); `seed_base`/`post_sweep` unused in serial mode; the docstring drop of "warp" fixes a pre-existing inaccuracy (the constructor never accepted it). None warrant a change.

Merging once CI (test matrix incl. windows-latest) is green.

PR #586, opened 2026-07-27
perf(vae): GEMM the ProdLDA-family decoder matmuls (#378)

### LINKED ISSUE #378: Perf: ProdLDA-family VAE fit is single-threaded scalar Rust (4-7x slower than PyTorch at scale)
## Summary

The hand-coded VAE backbone shared by `ProdLDA`, `CombinedTM`, `ZeroShotTM`, `ETM(inference="vae")`, `InfoCTM`, and now `Scholar` (#376/#377) is single-threaded scalar Rust. On realistic-density inputs it is within ~2-3x of a CPU PyTorch reference at small sizes but degrades to **4.7-6.8x** as `V·K` grows, because the dense decoder matmul (`logit = theta·beta`, `K×V`) and the encoder GEMMs dominate and PyTorch runs them on multithreaded BLAS.

This surfaced in the SCHOLAR PR1 (#377) benchmark. The gap is **inherited wholesale from the ProdLDA backbone — Scholar itself adds only ~1%** over plain `ProdLDA` — so it is a family-wide perf item, not model-specific.

## Measurements (fit-only, from the #377 benchmark, CPU)

| size | docs / V / K / iters | topica | dallascard/scholar (CPU PyTorch) | ratio |
|---|---|---|---|---|
| small  | 500 / 500 / 10 / 200   | 4.1s   | 2.6s  | 1.60x |
| medium | 2000 / 2000 / 20 / 200 | 49.0s  | 10.5s | 4.69x |
| large  | 4000 / 3000 / 30 / 200 | 157.3s | 23.0s | 6.84x |

`Scholar` vs plain `ProdLDA` at the medium size: 48.6s vs 47.9s (~1.01x) — confirms the cost is the backbone.

## Where the time goes

- Decoder: `logit_raw[i][j] += theta_do[i][t] * beta[t*V+j]` over the full `K×V` beta for every doc (`prodlda.rs::batch_forward`, and its transpose in `batch_backward`). Dense `N×K×V`.
- Encoder layer-1/2 GEMMs and the head projections, all scalar loops.

## Options to explore (not committed to any)

- Multi-thread the per-document batch work with rayon, combining partial sums in a **fixed document order** to preserve the bit-for-bit determinism guarantee (the hard constraint — see `docs/contributing/conventions.md`).
- Block/tile the decoder matmul, or route the `N×K · K×V` products through a small internal GEMM (`matrixmultiply`/`ndarray`) while keeping deterministic reduction order.
- SIMD-friendly inner loops (contiguous `beta` rows already help; the backward accumulation into `g.beta` is the awkward one).

## Constraints

- **Determinism is non-negotiable:** fits must stay bit-for-bit identical regardless of thread count (fixed-order reductions). Any parallelization must be proven with `np.array_equal` at two thread counts.
- No PyTorch / autodiff dependency — the whole point of the backbone is a dependency-light hand-coded VAE.
- Must not regress the FD gradient checks or the parity of any VAE-family model.

## Priority

Correctness/faithfulness is fine today (all VAE-family models validate); this is a throughput improvement. Filing so the known profile is tracked rather than rediscovered per-model.

nealcaren:
Closes #378 (the decoder half; see **Scope & follow-ups**).

## What

The hand-coded VAE backbone shared by `ProdLDA`, `CombinedTM`, `ZeroShotTM`, `ETM(inference="vae")`, `InfoCTM`, and `Scholar` spent most of a fit in three dense `O(N·K·V)` decoder terms written as scalar triple-loops, running **4.7–6.8× slower than a CPU PyTorch reference** at scale. This routes them through a pure-Rust, BLAS-free GEMM (`matrixmultiply`):

```
forward : logit_raw = theta_do · beta          (matmul_nn)
backward: dtheta_do = dlogit_raw · betaᵀ        (matmul_nt, betaᵀ via stride-swap, no copy)
backward: g.beta   += theta_doᵀ · dlogit_raw   (matmul_tn — the cross-document reduction)
```

New `src/gemm.rs` wraps `matrixmultiply::dgemm` with row-major `nn`/`nt`/`tn` helpers (transposed operands expressed by swapping strides — nothing is copied or transposed in memory) plus scalar-parity unit tests. The decoder sections of `batch_forward`/`batch_backward` flatten the `Vec<Vec>` operands, GEMM, and scatter back; SCHOLAR's content-deviation terms stay scalar and bit-identical (accumulation order preserved). `matrixmultiply` was already a transitive dependency via `ndarray`, so **no PyTorch, no BLAS, no new external weight** — the single-wheel promise is unchanged.

## Determinism (the hard constraint)

A GEMM's reduction order is fixed, so fits are **bit-for-bit identical regardless of thread count**. `matrixmultiply` parallelizes over *output blocks* while each element's `k`-reduction stays single-threaded, so bit-identity holds even with its `threading` feature on — this is exactly the property a rayon-over-batch reduction of `g.beta` could not offer. Verified end-to-end (not just microbenchmark): a real `ProdLDA` fit gives identical `topic_word`/`doc_topic` hashes across `MATMUL_NUM_THREADS = 1, 2, 4`, on both a small config and a large one where threading actually engages. Output differs from the old scalar path only at float-rounding level (**~1e-15 relative**), so reference-parity (gold) and finite-difference-gradient tests are unchanged.

## Expected speedup

**Decoder terms in isolation** (single-thread, `benchmarks/vae_gemm_probe`) — scales with K:

| K | 10 | 20 | 30 | 50 | 80 |
|---|----|----|----|----|----|
| decoder speedup | 2.7× | 4.2× | 4.9× | 5.4× | 6.8× |

**End-to-end fit** (single-thread, decoder is ~half the iteration at typical K, so Amdahl-bounded):

| workload | main | this PR | speedup |
|---|---|---|---|
| D2000 / V2000 / K20 / 40 iters | 27.40s | 18.63s | **1.47×** |

At typical K (10–30) expect **~1.4–1.6× end-to-end**, rising toward the decoder's own 6–7× as K grows (at K=80 the decoder is ~80% of the iteration).

**More cores:** the decoder GEMM scales bit-identically, but this PR only touches the decoder — the per-document encoder and the elementwise `O(N·V)` work are still single-threaded, so end-to-end multi-core scaling is capped (and negligible at small `V·K`, where the decoder is below `matrixmultiply`'s internal threading threshold). Meaningful multi-core scaling shows only at large `V·K`; broad scaling needs the follow-ups below.

## Scope & follow-ups

This is **step 1** of the direction explored in `benchmarks/notes_issue_378_vae_gemm.md` (the decoder GEMM — the dominant cost, and the safest deterministic change). Deliberately deferred so this determinism-critical shared hot path lands in a reviewable increment:

- **Encoder GEMMs** (layer-2 + head projections; keep layer-1 sparse).
- **rayon over the batch** for the per-document-independent elementwise `O(N·V)` work (batchnorm apply, recon softmax, reparam), reductions kept in fixed doc order.

Together those unlock the multi-core scaling this PR leaves on the table.

## Testing

- `cargo test --lib` — `gemm` 4/4, `prodlda` 26/26, `etm_vae` 9/9, `infoctm` 4/4, `scholar` 9/9, `etm::` 12/12 (all the finite-difference gradient + planted-recovery checks).
- `cargo clippy --workspace --all-targets --all-features -D warnings` — clean; `cargo fmt --check` — clean.
- Python: 170 VAE-family + determinism tests pass; the gold/parity tests (`prodlda`/`combinedtm`/`etm`/`zeroshottm`/`infoctm`) are unchanged.

### Independent verification

An independent agent (which did **not** write the change) re-read the diff, verified the three GEMMs and the SCHOLAR restructure by hand, re-ran the full FD suite, confirmed bit-identical hashes across `MATMUL_NUM_THREADS=1/2/4` on two configs, ran 124 Python tests (0 failures), and produced the true before/after above (main 27.40s → 18.63s). Verdict: *"a faithful, deterministic, and faster implementation … safe to PR."* Its two non-blocking notes: `matmul_*` don't special-case `k==0` (relies on `dgemm` zeroing C via `beta=0`; not a real VAE scenario), and the determinism guarantee rests on `matrixmultiply`'s output-block partitioning strategy (`Cargo.toml` pins `0.3`, and `src/gemm.rs` documents the dependency).

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,8 +31,8 @@ follow [Semantic Versioning](https://semver.org/spec/v2.0.0.html) once released.
 ### Changed
 
 - **ProdLDA-family VAE decoder now uses a BLAS-free GEMM (2.7–6.8× faster on the
-  decoder)** (#378). The hand-coded VAE backbone shared by `ProdLDA`, `CombinedTM`,
-  `ZeroShotTM`, `ETM(inference="vae")`, `InfoCTM`, and `Scholar` spent most of a fit
+  decoder)** (#378). The hand-coded dense-decoder VAE backbone shared by `ProdLDA`,
+  `CombinedTM`, `ZeroShotTM`, `InfoCTM`, and `Scholar` spent most of a fit
   in three dense `O(N·K·V)` decoder terms written as scalar triple-loops, which ran
   4.7–6.8× slower than a CPU PyTorch reference at scale. These now route through
   `matrixmultiply` (pure-Rust, cache-blocked + SIMD, BLAS-free — no PyTorch, no new
@@ -44,8 +44,10 @@ follow [Semantic Versioning](https://semver.org/spec/v2.0.0.html) once released.
   decoder scales across cores while keeping that guarantee. Output differs from the
   previous scalar path only at floating-point-rounding level (~1e-15 relative), so
   reference-parity and finite-difference-gradient tests are unchanged. SCHOLAR's
-  content-deviation terms stay scalar and bit-identical. Encoder GEMMs and the
-  elementwise `O(N·V)` work are untouched (candidate follow-ups).
+  content-deviation terms stay scalar and bit-identical. `ETM(inference="vae")` is
+  not affected — it has its own sparse per-document decoder, not the dense
+  `theta·beta`. Encoder GEMMs and the elementwise `O(N·V)` work are untouched
+  (candidate follow-ups).
 
 ## [0.53.0] - 2026-07-26
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,25 @@ follow [Semantic Versioning](https://semver.org/spec/v2.0.0.html) once released.
   covariate fit shifts negligibly (R-gold keyword cosine 0.912 → 0.906, still far
   above bar, sign agreement 1.00).
 
+### Changed
+
+- **ProdLDA-family VAE decoder now uses a BLAS-free GEMM (2.7–6.8× faster on the
+  decoder)** (#378). The hand-coded VAE backbone shared by `ProdLDA`, `CombinedTM`,
+  `ZeroShotTM`, `ETM(inference="vae")`, `InfoCTM`, and `Scholar` spent most of a fit
+  in three dense `O(N·K·V)` decoder terms written as scalar triple-loops, which ran
+  4.7–6.8× slower than a CPU PyTorch reference at scale. These now route through
+  `matrixmultiply` (pure-Rust, cache-blocked + SIMD, BLAS-free — no PyTorch, no new
+  external/BLAS dependency, so the single-wheel promise is unchanged): the forward
+  `logit = theta·beta`, the backward `dtheta = dlogit·betaᵀ`, and the
+  cross-document `g.beta += theta_doᵀ·dlogit` reduction. A GEMM's reduction order is
+  fixed, so fits stay **bit-for-bit identical regardless of thread count** (verified
+  across `MATMUL_NUM_THREADS = 1, 2, 4`), and with the `threading` feature the
+  decoder scales across cores while keeping that guarantee. Output differs from the
+  previous scalar path only at floating-point-rounding level (~1e-15 relative), so
+  reference-parity and finite-difference-gradient tests are unchanged. SCHOLAR's
+  content-deviation terms stay scalar and bit-identical. Encoder GEMMs and the
+  elementwise `O(N·V)` work are untouched (candidate follow-ups).
+
 ## [0.53.0] - 2026-07-26
 
 ### Fixed
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -190,6 +190,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "2304e00983f87ffb38b55b444b5e3b60a884b5d30c0fca7d82fe33449bbe55ea"
 
 [[package]]
+name = "hermit-abi"
+version = "0.5.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "fc0fef456e4baa96da950455cd02c081ca953b141298e41db3fc7e36b1da849c"
+
+[[package]]
 name = "indoc"
 version = "2.0.7"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -232,7 +238,10 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "a06de3016e9fae57a36fd14dba131fccf49f74b40b7fbdb472f96e361ec71a08"
 dependencies = [
  "autocfg",
+ "num_cpus",
+ "once_cell",
  "rawpointer",
+ "thread-tree",
 ]
 
 [[package]]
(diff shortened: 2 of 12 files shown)


nealcaren:
## Double review (faithfulness + adversarial) + local verification

Reviewed by a correctness/faithfulness pass, an independent adversarial pass (Gemini pro-tier), and end-to-end local verification. **The numerics are correct and the determinism design is sound; one documentation overclaim needs fixing before merge.**

### Correctness & determinism — verified ✅
- **GEMM wrappers** (`gemm.rs`): the `nn`/`nt`/`tn` stride arguments and transpose-via-stride-swap are correct, cross-checked against the in-file scalar references; the call-site `(m,k,n)` mappings in `prodlda.rs` match the operand shapes (θ N×K, β K×V, dlogit N×V) at all three sites.
- **`g.beta` reduction**: no clobber. The `tn` GEMM writes a fresh zeroed scratch buffer (`gbeta`), then `g.beta += gbeta` — preserving the old accumulate semantics and any coupled caller's pre-seeded gradient. This was the main hazard and it's correctly avoided (β=0 into a private buffer).
- **Flatten/scatter**: contiguous row-major, no off-by-one or row/col swap; `logit_raw` is zeroed and overwritten, matching the old path (SCHOLAR content added after, as before).
- Ran locally: `gemm` scalar-parity tests, **all 26 `prodlda` FD-gradient tests** (bow / emb / bow+emb / contrastive / dirichlet / stick-breaking), **58 VAE-family Python gold tests** (ProdLDA/CombinedTM/ZeroShotTM/Scholar/InfoCTM) — all pass. Cross-thread determinism confirmed: identical `topic_word` MD5 across `MATMUL_NUM_THREADS = 1, 2, 4` on an 800-doc ProdLDA fit.

### Change needed — ETM(inference="vae") overclaim 🔧 (blocker: doc accuracy)
The CHANGELOG, the `gemm.rs` module docstring, and `benchmarks/notes_issue_378_vae_gemm.md` all list **`ETM(inference="vae")`** among the models whose decoder now routes through the GEMM. It does not: `etm_vae.rs` imports only the reparam/KL/InfoNCE helpers from `prodlda` (not `batch_forward`/`batch_backward`) and has its **own sparse per-document decoder** (`doc_backward`, `m += theta[t]*beta[t][w]`), which this PR leaves untouched — so ETM(vae) gets no speedup. The real beneficiaries are the 5 dense-decoder callers: **ProdLDA, CombinedTM, ZeroShotTM** (in `prodlda.rs`) plus **InfoCTM** and **Scholar** (the only external callers of `batch_forward`/`batch_backward`). Please drop ETM from those three lists, or note explicitly that ETM(vae)'s separate sparse decoder is out of scope (a candidate follow-up). topica's anti-overclaim norm makes this worth getting exactly right.

### Optional (nit) — `k==0` contract
`matmul_tn` on an **empty batch** (N=0) is called with the contraction dim `k=0` while `m=K≠0, n=V≠0`, so the `m==0||n==0` guard doesn't fire and `dgemm` runs with `k=0`. It's safe today only because every output buffer is freshly `vec![0.0]` (matrixmultiply may skip the β-scale when k=0, leaving C untouched). The wrapper docstring says "Overwrites `c`", which is slightly stronger than the impl guarantees for `k==0`. Consider a one-line `if k == 0 { c.fill(0.0); return; }` (or a doc note) so the contract holds unconditionally.

Fix the ETM wording and this is good to merge — the core refactor is faithful, fast, and deterministic.

nealcaren:
Pushed `c8fccf6` addressing the review:

- **ETM(inference="vae") overclaim fixed** — corrected the model list in the `gemm.rs` docstring, CHANGELOG, and `benchmarks/notes_issue_378_vae_gemm.md` to the five real dense-decoder beneficiaries (ProdLDA, CombinedTM, ZeroShotTM, InfoCTM, Scholar), with an explicit note that ETM(vae)'s separate sparse decoder is out of scope.
- **`k==0` guard added** — `matmul_*` now zero-fills `C` when the contraction dim is 0 (hit by the `tn` reduction on an empty batch), so the documented "overwrites `c`" contract holds unconditionally rather than relying on pre-zeroed buffers. Added a `zero_contraction_zeroes_c` unit test.

Correctness of the core refactor was already verified locally (gemm parity tests, all 26 prodlda FD-gradient tests, 58 VAE-family gold tests, and bit-identical `topic_word` across `MATMUL_NUM_THREADS = 1/2/4`). fmt + all-features clippy green.

PR #588, opened 2026-07-27
feat(compare): topica.compare(fit_a, fit_b) — statistical topic-drift comparison (#415)

### LINKED ISSUE #415: topica.compare(fit_a, fit_b): statistical topic-drift comparison and regression test
The analysis-manifest V2 roadmap (#394) lists "manifest compare" as remaining, and today comparing two fits is a manual affair. Make it a first-class, *statistical* operation rather than a diff — one that answers the question social scientists actually ask ("did the topics change, or did the sampler just wander?") and doubles as a regression test for the library itself.

### Proposal
```python
cmp = topica.compare(fit_a, fit_b)     # two fitted models, or two AnalysisManifests

cmp.aligned                            # matched topic pairs (Hungarian on top-word / word-dist overlap)
cmp.unmatched_a, cmp.unmatched_b       # topics with no honest counterpart (appeared / vanished / split)
cmp.drift                              # per-matched-pair distance + whether it exceeds a reseed baseline
cmp.prevalence_shift                   # change in topic prevalence, with uncertainty
cmp.render()                           # HTML/markdown card, same house as the manifest render()
```

### Three use cases, one tool
1. **Over time / two corpora** (same model, different slice): which topics are stable, which drift, which are new or gone.
2. **Two seeds / two versions of the same corpus**: how much of the apparent change is just Monte Carlo wander — the *baseline* against which real drift is judged.
3. **Library regression test**: refit a canonical corpus across topica versions and flag when a "harmless" refactor silently moves the topics. This is a self-test we do not currently have.

### Design points
- **Alignment**: reuse the topic-alignment already living in the `topica.ensemble` code (Hungarian assignment on top-word overlap / Hellinger on word distributions). Do not reinvent it.
- **Honest "unmatched" bucket**: alignment across genuinely different topics — and especially different K — is ambiguous. Force nothing; a topic with no good match is reported as unmatched, not paired to its least-bad neighbor. A split (one topic in A → two in B) is a named outcome.
- **Drift needs a null**: raw distance between two topics is uninterpretable on its own. Provide a reseed/subsample baseline so "this pair moved more than reseeding alone would" is a claim with a threshold, not a vibe.
- **Manifest-native**: `compare` should accept two `AnalysisManifest`s (fingerprints + retained top words) as well as two live models, so a comparison can run from provenance records without re-fitting.

### Non-goals
Not a merge/ensemble step (that is `topica.ensemble`). This describes and tests difference; it does not build a consensus model.

nealcaren:
Closes #415 (the live-model path; see **Scope & follow-up**).

## What

Comparing two fits was a manual affair. This makes it a first-class, *statistical* operation that answers the question social scientists actually ask — *"did the topics change, or did the sampler just wander?"* — and doubles as a library regression test.

```python
cmp = topica.compare(fit_a, fit_b)

cmp.aligned            # matched topic pairs (Hungarian on topic-word distance)
cmp.unmatched_a        # topics that vanished (only in A) — never force-paired
cmp.unmatched_b        # topics that appeared (only in B)
cmp.splits, cmp.merges # one-to-many outcomes, named honestly (esp. across different K)
cmp.drift              # per-pair distance + whether it beats the reseed null
cmp.prevalence_shift   # change in topic prevalence, with a standard error
print(cmp.render())    # HTML card (manifest house style); to_markdown() too
```

## Design commitments (from the issue)

- **Alignment is reused, not reinvented.** Matching runs through the existing `align_topics` (Hungarian assignment), which already yields matches, splits, merges, and unaligned topics.
- **The unmatched bucket is honest.** A topic with no good counterpart is reported as *appeared* / *vanished*, never paired to its least-bad neighbor; a split (one topic in A → two in B) is a named outcome. Matters most across different K. Every A and B topic is surfaced (matched, a split/merge participant, or unmatched) — no silent drops.
- **Drift needs a null.** A raw distance is uninterpretable on its own. Pass a reseed baseline and each matched pair is flagged as drifting *beyond reseed noise* — a claim with a threshold, not a vibe:
  ```python
  cmp = topica.compare(a, b, refit=lambda s: topica.LDA(num_topics=20, seed=s).fit(corpus))
  cmp = topica.compare(a, b, reseed_fits=[a2, a3, a4])   # refits you already have
  cmp = topica.compare(a, b, baseline=0.70)              # a family self-agreement ceiling
  ```
  The floor for each A-topic is the *worst* self-agreement it keeps across the reseeds; an A↔B match below that floor moved more than reseeding A alone did (`drifted=True`). Without a null, distances are still reported but `drifted` is `None` (honest "unknown"), never a false verdict.
- **Prevalence shift with uncertainty.** `b - a` of `doc_topic.mean(0)` per matched pair, with a standard error when a posterior over theta is available (via `composition_theta` — no corpus needed for logistic-normal models or Gibbs models with retained draws).
- **Render in the house style.** `render(path=None, *, title=None) -> str` (HTML) and `to_markdown() -> str`, mirroring `AnalysisManifest.render` — plain strings, not a Plotly/`_repr_html_` object.
- **Not an ensemble.** Describes and tests *difference*; does not build a consensus model.

## Files

- New `python/topica/compare.py` — `compare()` + `CompareResult` / `MatchedPair` / `UnmatchedTopic`, exported from `__init__.py` and `__init__.pyi`.
- `tests/test_compare.py` — permuted-topic matching; the honest unmatched/split coverage invariant at different K (both directions); the drift-vs-reseed-null verdict (only the genuinely-moved topic flags; seed-wander does not); the `baseline=`, `reseed_fits=`, and `refit=` null paths; prevalence shift; render/markdown/dict; rejects >1 null source; a real-LDA integration test.
- Docs: a `compare` section in the model-comparison guide (built on the self-agreement ceiling it already describes) + CHANGELOG.

## Testing

- `tests/test_compare.py` — 12/12 pass (incl. the slow real-LDA integration test).
- `test_stub_sync` + `test_align_topics` + `test_manifest` + `test_ensemble` — pass; the new export breaks nothing. Dependent suites (`standard_errors`, `coherence`, `analysis`) green. `mkdocs build --strict` clean.

### Independent verification

An independent agent (which did **not** write this) read the module and issue, then tried to break it with its own adversarial inputs: identical fits + a reseed null flag nothing (floors ~0.9997); a topic re-worded beyond reseed noise flags while its neighbors don't; a topic rewritten to duplicate another is surfaced as *vanished* + a named split, **not** force-matched; K5→3 and K3→5 both union-cover every topic; `prevalence_shift` matches `b-a` of the doc-topic means to 1e-12 with an SE for real LDA and `None` for raw arrays; `to_dict()` is `json.dumps`-able; render writes a file. Verdict: *"a faithful, honest, correct implementation … reuses `align_topics`, never force-pairs, judges drift against a real reseed null, reports prevalence with uncertainty, and renders in the manifest house style; safe to PR."* Its notes are non-blocking: the reseed floor is a conservative `min` over a small sample (a documented tripwire choice, not a statistical CI), and — addressed here — it asked for a fast unit test of the `refit=` callable path (added).

## Scope & follow-up

The issue also lists a **manifest-native path** (compare two `AnalysisManifest`s without refitting). Today's manifests retain only fingerprints, **not top words**, so that path needs `record_fit` extended to retain top words + prevalence (a provenance-schema change) plus a set-overlap aligner (manifests carry no beta for `align_topics`). Deferred as a tracked follow-up to keep this PR focused on the statistical core; called out honestly rather than half-built.

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,13 +10,15 @@ follow [Semantic Versioning](https://semver.org/spec/v2.0.0.html) once released.
 
 - **`topica.compare(fit_a, fit_b)` — statistical two-fit topic comparison** (#415).
   Comparing two fits is now a first-class operation rather than a manual diff. It
-  aligns topics one-to-one (reusing `align_topics`' Hungarian assignment), reports
-  topics with no honest counterpart as *appeared* / *vanished* (never force-paired)
-  and one-to-many relationships as named *splits* / *merges*, and — given a reseed
-  baseline (`refit=` a callable, `reseed_fits=` a list, or `baseline=` a similarity
-  floor) — flags each matched pair as drifting *beyond reseed noise* rather than
-  reporting an uninterpretable raw distance. It also reports each topic's
-  `prevalence_shift` (with a standard error when a posterior over theta is
+  aligns topics one-to-one (reusing `align_topics`' mutual-best above-threshold
+  match), reports topics with no honest counterpart as *appeared* / *vanished*
+  (never force-paired) and one-to-many relationships as named *splits* / *merges*,
+  and — given a reseed baseline (`refit=` a callable, `reseed_fits=` a list, or
+  `baseline=` a similarity floor) — flags each matched pair when it moves *beyond
+  the range of self-agreement A shows across the reseeds* (a heuristic band, not a
+  calibrated test) rather than reporting an uninterpretable raw distance. It also
+  reports each topic's `prevalence_shift` (with a posterior-spread uncertainty,
+  combined across the two fits in quadrature, when a posterior over theta is
   available) and renders an HTML/Markdown card in the same house as the analysis
   manifest (`cmp.render()` / `cmp.to_markdown()`). Three uses, one tool: track topic
   stability across two corpora/time slices, separate real change from seed wander,
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,25 @@ follow [Semantic Versioning](https://semver.org/spec/v2.0.0.html) once released.
 
 ## [Unreleased]
 
+### Added
+
+- **`topica.compare(fit_a, fit_b)` — statistical two-fit topic comparison** (#415).
+  Comparing two fits is now a first-class operation rather than a manual diff. It
+  aligns topics one-to-one (reusing `align_topics`' Hungarian assignment), reports
+  topics with no honest counterpart as *appeared* / *vanished* (never force-paired)
+  and one-to-many relationships as named *splits* / *merges*, and — given a reseed
+  baseline (`refit=` a callable, `reseed_fits=` a list, or `baseline=` a similarity
+  floor) — flags each matched pair as drifting *beyond reseed noise* rather than
+  reporting an uninterpretable raw distance. It also reports each topic's
+  `prevalence_shift` (with a standard error when a posterior over theta is
+  available) and renders an HTML/Markdown card in the same house as the analysis
+  manifest (`cmp.render()` / `cmp.to_markdown()`). Three uses, one tool: track topic
+  stability across two corpora/time slices, separate real change from seed wander,
+  and use it as a library regression test across topica versions. Not an ensemble —
+  it describes and tests difference, it does not build a consensus model.
+  (Manifest-native comparison from provenance records is the tracked follow-up; it
+  needs `record_fit` to retain top words, which it does not yet.)
+
 ### Fixed
 
 - **`DMR` / `GDMR` covariate prior now faithfully reproduces `tomotopy`** (#563).
--- a/docs/guides/model-comparison.md
+++ b/docs/guides/model-comparison.md
@@ -279,3 +279,51 @@ corpus, with K, and with how far each fit is run toward convergence (a
 lightly-iterated Gibbs run self-agrees less than a long one). The ordering across
 families, though — deterministic and batch-variational at the top, online VB at
 the bottom — is the stable, transferable lesson.
+
+## Comparing two fits statistically: `topica.compare`
+
+The ceiling above is exactly what makes a raw distance between two topics
+uninterpretable on its own: is a drop from cosine 1.00 to 0.80 real movement, or
+just the Monte-Carlo wander any refit shows? `topica.compare(fit_a, fit_b)` answers
+that question as a first-class operation rather than a manual diff.
+
+```python
+cmp = topica.compare(fit_a, fit_b)
+
+cmp.aligned            # matched topic pairs (Hungarian on topic-word distance)
+cmp.unmatched_a        # topics that vanished (only in A) — never force-paired
+cmp.unmatched_b        # topics that appeared (only in B)
+cmp.splits, cmp.merges # one-to-many outcomes, named honestly (esp. across different K)
+cmp.drift              # per-pair distance + whether it beats the reseed null
+cmp.prevalence_shift   # change in topic prevalence, with a standard error
+print(cmp.render())    # HTML card (same house as the manifest render()); to_markdown() too
+```
+
+Alignment reuses `align_topics` (Hungarian assignment), so a topic with no
+honest counterpart is reported as *appeared* / *vanished*, not paired to its
+least-bad neighbor — and a split (one topic in A → two in B) is a named outcome, not
+a silent mismatch. This matters most when the two fits have different K.
+
+**Drift needs a null.** Give `compare` a reseed baseline and each matched pair is
+flagged as drifting *beyond reseed noise* — a claim with a threshold, not a vibe:
+
+```python
+# refit A on the same corpus under fresh seeds; compare builds the per-topic floor
+cmp = topica.compare(fit_a, fit_b, refit=lambda s: topica.LDA(num_topics=20, seed=s).fit(corpus))
+
+# or pass refits you already have, or a flat similarity floor (a family ceiling)
+cmp = topica.compare(fit_a, fit_b, reseed_fits=[a2, a3, a4])
+cmp = topica.compare(fit_a, fit_b, baseline=0.70)   # e.g. an online-VB ceiling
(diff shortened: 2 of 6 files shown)


nealcaren:
## Full review (faithfulness/statistical + adversarial)

Reviewed by a statistical-validity pass and an independent adversarial pass, plus a read of `align_topics`/`_prevalence`. **The engineering is clean and the matching machinery is correct; the two headline *statistical* claims are framed more strongly than they're grounded. CHANGES-NEEDED — mostly wording/calibration, no crashes.**

First, to clear the deck: an adversarial pass flagged a "critical inversion" in `_reseed_null` (treating `1.0 - dist` as if `dist` were a similarity). **That's a false alarm** — `AlignmentResult` subclasses `list`, and the *list* base holds `(a, b, distance)` (`legacy_pairs`) while `.matches` holds `(a, b, similarity)`. `_reseed_null` iterates the list (distance, correctly named `dist`) and `compare()` iterates `.matches` (similarity); both are dimensionally consistent, and the null is metric-consistent across cosine/js/rbo/emd. No inversion bug.

### Blocker 1 — the reseed null is a `min`-of-N heuristic, not a calibrated test
`_reseed_null` sets each topic's floor to `min(self-similarity over n_reseed refits)` and flags drift as `sim < floor`. For a genuinely-null B (just another reseed), P(below the min of `n_reseed`) ≈ `1/(n_reseed+1)` — **~20% false positive at the default `n_reseed=4`** — and because the floor is a `min`, *adding* reseeds **lowers** it, making drift *harder* to flag (backwards from "more null data ⇒ tighter test"). The docstring/guide framing ("a claim with a threshold, not a vibe") oversells this. Fix either way: (a) use a lower-quantile floor at a stated significance level (e.g. 5th percentile), or (b) reframe the prose to "moved beyond the observed reseed range" and drop the calibrated-test language. Also note it's asymmetric (only A is reseeded).

### Blocker 2 — the prevalence-shift SE assumes the two fits are independent
`shift_se = sqrt(se_a² + se_b²)` is only valid when A and B are fit on **different** corpora. For the advertised "two seeds / two versions of the same corpus" use, `prev_a` and `prev_b` are correlated (shared documents), so the difference's SE omits the covariance term and generally overstates uncertainty. `compare()` can't tell which regime it's in and doesn't warn. At minimum document the independence assumption; ideally detect/require same-corpus and note the SE is only valid across independent fits.

### Minors
- **"Standard error" is a posterior SD.** `_prevalence` returns the between-draw spread of the mean over `composition_theta` draws — a legitimate posterior uncertainty, but not a frequentist SE over documents. Relabel in the docstring to avoid conflation.
- **"Hungarian" is mislabeled.** `.aligned` comes from `align_topics`' *mutual-uniqueness threshold* rule, not the Hungarian assignment; the `compare` docstring and `docs/guides/model-comparison.md` both say "Hungarian on topic-word distance." (The reseed null *does* use the Hungarian `legacy_pairs` — so the two paths use different matching rules, worth stating.) The never-force-pair behavior is good; just describe it accurately.
- **`threshold=0.3` + `cosine` is permissive on real corpora** — shared common words push cross-topic cosine above 0.3, over-reporting splits/merges. Every fast test uses disjoint one-hot blocks, so this regime is never exercised; the realistic LDA test is `@slow`. Consider a higher default or a documented caveat.
- **Nits:** empty `reseed_fits=[]` reports `kind:"reseed", n_reseed:0` but yields no floor (all `drifted=None`) — should be `kind:"none"` or raise; no friendly validation that `a`/`b` are compatible fitted models (disjoint vocab raises an undocumented `ValueError` from `align_topics`); NaN in `doc_topic` propagates silently; K=1 untested.

### What's good
Reuses `align_topics` rather than reinventing matching; **honest about missing information** (returns `None`/`unmatched` rather than fabricating pairs or floors — the right instinct); the directional tests genuinely exercise the claims (identical fits → all-stable, a reseed null flagging exactly the moved topic, appeared-topic-at-higher-K, the baseline and refit-callable paths). The gaps are the false-positive regime, the realistic cosine-on-shared-vocab regime, and K=1/NaN/same-corpus-SE.

Net: the machinery is sound and the instincts are right; the required changes are to make the *prose* match the *statistics* (blockers 1–2 + the "SE"/"Hungarian" relabels). Once the null is either calibrated or reframed and the SE assumption is stated, this is a genuinely useful addition.

nealcaren:
Pushed `cf362c1` addressing the review — all fixes are to make the **prose match the mechanics** (the machinery was already correct):

- **Reseed null reframed** (Blocker 1): dropped the "a claim with a threshold, not a vibe" / "beyond reseed noise" framing. It's now described as what it is — a **heuristic band** ("beyond the range of self-agreement A shows across the reseeds"), explicitly **not a calibrated significance test**, with the `~1/(n_reseed+1)` rough false-positive character and the A-only asymmetry stated. (module + `compare()` docstrings, docs guide, CHANGELOG, and the `render()` label.)
- **Prevalence "standard error" relabeled** (Blocker 2): now called what it is — each fit's posterior spread (between-draw SD of mean prevalence) combined in quadrature — with the independence assumption documented (correct for distinct fits, conservative as they approach identical). On a second look this isn't a formula bug: quadrature is right for two genuinely distinct fits; the fix is precision of wording, not the math.
- **"Hungarian" corrected**: `.aligned` is `align_topics`' mutual-best above-`threshold` match, not the Hungarian assignment; fixed everywhere, and noted the reseed null *does* use the Hungarian self-assignment.
- **`threshold=0.3` permissiveness** for cosine on real corpora — documented.
- **Nit**: an empty `reseed_fits=[]` now reports baseline `kind="none"` instead of a size-0 reseed run.

No behavior change beyond that last label; the 12 fast `test_compare` tests pass. Left as optional follow-ups for the author: tests for the false-positive regime, the realistic shared-vocab cosine regime, and K=1/NaN (all flagged in the review). The false-alarm "inversion bug" from the adversarial pass was confirmed a non-issue (the list base yields distance by design), so nothing to do there.

Issue #400, opened 2026-07-22
Uniform model settings introspection (`model.settings` / `get_params()`)


nealcaren:
Surfaced while building the analysis manifest (#394): there is no uniform way to read a fitted model's constructor configuration back off the object.

### Problem
Models expose their settings unevenly — `LDA` has `alpha`/`beta`, `STM` exposes `num_topics` but not much before fitting, and most models expose nothing structured. To record settings, the manifest maintains a hand-written per-model allowlist (`_SETTINGS_ALLOWLIST`, currently LDA + STM only) and falls back to "generic" for everything else. That does not scale to ~40 models and is a second inventory to keep in sync.

### Proposal
Add a uniform, sklearn-style introspection surface returning the model's configuration as a JSON-serialisable dict:

```python
model.settings          # or model.get_params()
# -> {"num_topics": 20, "alpha": [...], "beta": 0.01, "seed": 42, ...}
```

- Cover every model with its own hyperparameters (the constructor arguments), keyword-named to match the constructor.
- Include `seed` (see #399), so a single call captures the full construction config.

### Why
- **Generally useful**: anyone introspecting or logging a model wants this; it is a standard estimator affordance.
- **Retires a manifest V2 backlog item**: with `model.settings` in place, the manifest just calls it and the per-model settings adapters (the deferred V2 "registry adapters for every model family") are no longer needed — the fix belongs on the model, not in the manifest.

Relates to #394 (and would replace the manifest's `_SETTINGS_ALLOWLIST`).

nealcaren:
Independent second opinion (OpenAI Codex, focused review) — endorses the uniform surface and sharpens the implementation:

**Verdict: yes, add it.** Use sklearn's actual contract — `get_params()` returning constructor parameters as a dict keyed by `__init__` argument names (models can ignore `deep`), with an optional read-only `settings` alias. Familiar for provenance, cloning, and inspection.

Key refinements:
- **Do not derive settings from Rust internal fields** — internal representation often differs from user-facing configuration. Instead declare each model's constructor config **once**, via a declarative Rust macro/trait that generates `#[new]`, `get_params`/`settings`, and (optionally) the property getters from one spec. Each pyclass supplies values to that spec rather than maintaining a second allowlist — that is what actually kills the drift.
[...]

PR #597, opened 2026-07-28
feat(llm): drop the llm CLI dep; dispatch llm_backend/llm_embed to lightweight provider SDKs

nealcaren:
Replaces the single heavyweight `llm` (Simon Willison CLI) dependency with per-provider **lightweight SDKs**, dispatched by the model name. The public API (`llm_backend` / `llm_embed`) is unchanged.

## Why
`uv pip install "topica[llm]"` silently backtracks to an ancient topica, because `llm` core requires a **pre-release** `sqlite-migrate` that uv blocks by default. `llm` also pulls **~42 packages** for the handful of functions topica used.

## Design (unified dispatcher)
```python
topica.llm_backend("gpt-4o-mini")               # -> OpenAI     topica[openai]
topica.llm_backend("claude-3-5-haiku-latest")   # -> Anthropic  topica[anthropic]
topica.llm_backend("gemini-2.5-flash")          # -> Gemini     topica[gemini]
topica.llm_backend("llama3", base_url="http://localhost:11434/v1")  # any OpenAI-compatible endpoint
topica.llm_embed(texts, model="text-embedding-3-small")            # OpenAI / Gemini embeddings
```
- `provider=` overrides auto-detection; `base_url=` forces the OpenAI-compatible client (ollama, openrouter, vLLM, LM Studio, groq).
- `key=` passes a key directly when no env var is set; local servers get a placeholder automatically.
- `topica[llm]` now means **all three** SDKs; `all` follows suit. `llm` + `llm-ollama` dropped.

## SDK footprint (all wheel-only, no pre-releases, uv-clean)
| extra | SDK | ~real deps |
|---|---|---|
| `topica[openai]` | openai | ~8 |
| `topica[anthropic]` | anthropic | ~18 |
| `topica[gemini]` | google-genai | ~25 |
| old `topica[llm]` | llm CLI | ~42 + pre-release `sqlite-migrate` |

## Embeddings / local
`llm_embed` dispatches OpenAI vs Gemini. Anthropic has no embedding API. For fully offline, in-process embeddings without a server, use `sentence-transformers` directly and pass the array (docs/examples updated).

## Live verification (real keys)
- OpenAI `gpt-4o-mini`, Gemini `gemini-2.5-flash` (native), and Claude via OpenRouter `base_url` all labeled a planted 3-topic corpus correctly; embeddings return proper matrices.
- Explicit `key=` works with no env var; no-key-anywhere raises the SDK's clear "pass an api_key / set OPENAI_API_KEY" error.

## Changes
- `labeling.py`: dispatcher + `_openai`/`_anthropic`/`_gemini` chat helpers.
- `embedding.py`: `llm_embed` OpenAI/Gemini dispatch.
- `topicgpt.py`: `TopicGPT(base_url=, key=)` forwarded to `llm_backend`.
- `pyproject.toml`: `openai` / `anthropic` / `gemini` extras; `llm` = all three.
- `tests/test_llm_backend.py`: dispatch + assembly with fake SDK modules.
- Docs + examples migrated off the old `llm`-CLI strings/plugins.

Checks: 4086 tests collect with no import errors; targeted LLM/labeling/eval/topicgpt suites pass; `mkdocs build --strict` clean; pre-push preflight (fmt/clippy/stub) green. No Rust touched.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -62,7 +62,11 @@ follow [Semantic Versioning](https://semver.org/spec/v2.0.0.html) once released.
   unchanged **except**: `llm_embed` drops the old `batch=` keyword (the SDKs batch
   natively), it no longer reaches in-process `sentence-transformers` models (compute
   those embeddings directly and pass the array, as every embedding model already
-  accepts), and Anthropic has no embedding API (`llm_embed` raises for it).
+  accepts), and Anthropic has no embedding API (`llm_embed` raises for it). Provider
+  routing is now by model family, not by the old `llm`-CLI plugin slug: a prefixed
+  name like `"ollama/qwen3"` or `"openrouter/…"` resolves by its last path segment
+  (so it defaults to OpenAI) — point at a local/alternate endpoint with `base_url=`
+  instead.
 - **ProdLDA-family VAE decoder now uses a BLAS-free GEMM (2.7–6.8× faster on the
   decoder)** (#378). The hand-coded dense-decoder VAE backbone shared by `ProdLDA`,
   `CombinedTM`, `ZeroShotTM`, `InfoCTM`, and `Scholar` spent most of a fit
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -51,6 +51,18 @@ follow [Semantic Versioning](https://semver.org/spec/v2.0.0.html) once released.
 
 ### Changed
 
+- **LLM helpers now use lightweight per-provider SDKs instead of the `llm` CLI**
+  (#597). `llm_backend` / `llm_embed` dispatch by model name to the `openai`,
+  `anthropic`, or `google-genai` SDKs (`gpt*`→OpenAI, `claude*`→Anthropic,
+  `gemini*`→Gemini; `provider=` overrides; `base_url=` forces an OpenAI-compatible
+  endpoint for ollama / openrouter / vLLM / LM Studio / groq). This drops the single
+  heavyweight `llm` dependency (which pulled ~42 packages and a pre-release
+  `sqlite-migrate` that `uv` blocks). Extras are now `topica[openai]` /
+  `[anthropic]` / `[gemini]`; `topica[llm]` installs all three. The public API is
+  unchanged **except**: `llm_embed` drops the old `batch=` keyword (the SDKs batch
+  natively), it no longer reaches in-process `sentence-transformers` models (compute
+  those embeddings directly and pass the array, as every embedding model already
+  accepts), and Anthropic has no embedding API (`llm_embed` raises for it).
 - **ProdLDA-family VAE decoder now uses a BLAS-free GEMM (2.7–6.8× faster on the
   decoder)** (#378). The hand-coded dense-decoder VAE backbone shared by `ProdLDA`,
   `CombinedTM`, `ZeroShotTM`, `InfoCTM`, and `Scholar` spent most of a fit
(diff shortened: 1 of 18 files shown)


nealcaren:
## Double review (faithfulness + adversarial) + local reproduction

The design is sound (lightweight per-provider SDKs, clean lazy imports, `base_url` correctly forces the OpenAI-compatible client for ollama/openrouter/vLLM), and the **open-model *defaults* are preserved** — `gpt-4o-mini` is the pre-existing default on `main`, not introduced here, and `llm_coherence`/`llm_intrusion` still require an explicit model. But there's a **CI-failing regression** plus a convention slip and a few minors. **CHANGES-NEEDED.**

### Blocker — CI is failing: `test_settings_keys_match_constructor[TopicGPT]`
`TopicGPT.__init__` gained `base_url` and `key`, but the `settings` dict wasn't updated to include them, so the #400 settings-contract test fails (`TopicGPT.settings keys mismatch`) on macOS + ubuntu. Reproduced locally (full suite, SDKs blocked to match CI): `1 failed, 2479 passed`. **Fix:** add `base_url` and `key` to `TopicGPT.settings`.

### Should-fix — open-model showcase regression
The `TopicGPT` docstring example flipped from `model="ollama/qwen3"` (main) to `model="gpt-4o-mini"` while the surrounding prose still says *"showcases open-source models."* Per topica's standing convention (default to / showcase open models, validate against openrouter qwen), keep the example open — revert that one to `ollama/qwen3` (or a local `base_url=` open model).

### Minors
- **`llm_embed` dropped `batch=`** (main had `batch=True`). No internal caller passes it and the new SDK batches natively, so behavior is fine — but it's an unannounced public-API break; add a CHANGELOG line (or keep it as an accepted no-op).
- **`llm_embed` with `provider="anthropic"` (or any non-gemini value) silently routes to OpenAI** rather than raising "Anthropic has no embedding API." Same for an unrecognized `provider=` in `llm_backend` → silent OpenAI. Validate `prov in {...}` and raise.
- **`_gemini_embed` single-text path:** please verify `resp.embeddings` is populated (not `None`) for a 1-item `contents` in the pinned `google-genai` version — the adversarial pass suspects a `TypeError` on `llm_embed(["one"], model="gemini-...")`. (Unverified — needs a live key; worth a quick check or a guarded `resp.embeddings or [resp.embedding]`.)

### Good
Dispatcher routing correct (base_url wins first, then `claude*`/`gemini*`, else OpenAI; `provider=` honored; namespaced `anthropic/claude-…` and `models/gemini-…` resolve via `split("/")[-1]`); no top-level SDK imports (core imports clean without extras); `pyproject.toml` extras well-formed and wheel-only; `test_llm_backend.py` genuinely exercises the dispatch matrix (all three providers, base_url + localhost placeholder key, env/explicit key, `provider=` override, both embed providers, missing-SDK ImportError messages); no secrets hardcoded; the dropped in-process `sentence-transformers` embed path is documented (BYO-array is already the norm).

Fix the settings-keys blocker and the showcase example, and this is good to go; the rest are minors.

nealcaren:
Pushed `c9aedf9` addressing the review:

- **Blocker fixed (CI):** `TopicGPT.settings` now reports `base_url`; `key` is deliberately kept out (it's a secret — must never land in a provenance record or save) and excluded from the #400 contract in the test, alongside the `backend` callable. `test_settings_keys_match_constructor[TopicGPT]` passes (verified locally with SDKs present and blocked).
- **Open-model showcase restored:** the `TopicGPT.with_prompt` docstring example is back to `model="ollama/qwen3"`.
- **Dispatch robustness:** `llm_backend` / `llm_embed` now raise a clear `ValueError` on an unknown `provider=` instead of silently routing to OpenAI; `llm_embed` raises for an Anthropic model (no embedding API); `_gemini_embed` guards the single-text case (`resp.embeddings or [resp.embedding]`).
- **CHANGELOG:** added the missing entry for the provider-SDK swap, documenting the `llm`-dep drop, the new extras, and the `llm_embed` `batch=` / sentence-transformers / Anthropic changes.

Verified locally: the settings blocker + `test_llm_backend`/`test_llm_eval`/`test_labeling` pass both with the SDKs installed and with them blocked (CI sim). No Rust touched.

PR #598, opened 2026-07-28
perf(vae): GEMM the ProdLDA-family encoder (~1.8–2× single-thread) (#378)

nealcaren:
Sequel to #586 (which GEMM'd the ProdLDA-family **decoder**). This GEMMs the **encoder** — the other half of the fit — for another single-thread speedup. Part of #378.

## What

The encoder's dense **layer-2** and **head projections** were per-document scalar loops. They now batch into the same deterministic, BLAS-free `matrixmultiply` helpers (`src/gemm.rs`) that #586 introduced:

```
forward : PRE2 (N×H) = H1·W2ᵀ ;  MU_RAW/LV_RAW (N×K) = HD·W_mu/lsᵀ
backward: g.w2 += dpre2ᵀ·H1 ;  g.w_mu/ls += dmu/dlvᵀ·HD   (cross-document reductions)
          dH1 = dpre2·W2 ;  dHD = dmu·W_mu + dlv·W_ls
```

- **Layer 1 stays sparse.** It's a sparse, mode-dependent matvec (bag-of-words / embedding / CombinedTM `adapt_bert`), so densifying would lose the sparsity win. Its backward accumulation into `g.w1` stays a per-document sum **in fixed document order** (a sparse cross-document reduction kept serial for bit-identity); its forward matvec is now a **rayon-parallel, order-preserving** map.
- `encode_raw` is refactored to share a new `encode_layer1` helper and is kept intact for single-document inference (`transform`) and tests.
- `g.w2`/`g.w_mu`/`g.w_ls` are GEMM'd into a scratch buffer then `+=` into the gradient — never clobbered, so a coupled/pre-seeded caller (SCHOLAR/InfoCTM) is safe, same pattern as `g.beta` in #586.

## Determinism (the hard constraint)

There are now **two** thread knobs — `RAYON_NUM_THREADS` (the layer-1 map) and `MATMUL_NUM_THREADS` (the GEMMs) — and a full `ProdLDA` fit is **bit-for-bit identical across the whole cross product 1/2/4 × 1/2/4**. The GEMMs have a fixed reduction order; the rayon map is an indexed, order-preserving collect; every remaining reduction (biases, sparse layer-1, loss, batch-norm stats) stays in fixed document order. Output differs from the old scalar path only at ~1e-15 (float rounding), so FD-gradient and reference-parity tests are unchanged.

## Result

| | before (decoder GEMM only) | after (+ encoder GEMM) | speedup |
|---|---|---|---|
| single-thread, D3000/V2000/K20/H200/40 iters | ~40–48s | ~20–27s | **~1.8–2×** |

(My in-repo run: 47.7s → 27.1s = 1.76×. The independent reviewer's clean baseline-vs-HEAD before/after against the exact merge-base measured **~2.0×**.) Combined with #586, the VAE family is ~2.3× faster single-threaded than before either change.

**Multi-core is only modest (~1.10× on 4 cores), by design/discovery:** once the encoder *and* decoder are both GEMM'd, the remaining serial `O(N·V)` reconstruction-softmax / batch-norm and the sparse layer-1 dominate. The prize here turned out to be single-thread, not multi-core — reframed honestly. Rayon-ing the `O(N·V)` elementwise remainder is a candidate follow-up.

## Scope

`ETM(inference="vae")` is unaffected (its own sparse per-document decoder). The dense-decoder family — ProdLDA, CombinedTM, ZeroShotTM, InfoCTM, Scholar — all benefit.

## Testing

- `cargo test --lib` — prodlda 26, etm_vae 9, infoctm 4, scholar 9, `etm::` 12, gemm 5 (all the finite-difference gradient + planted-recovery checks across every input mode and prior).
- `cargo clippy --workspace --all-targets --all-features -D warnings` — clean; `cargo fmt --check` — clean.
- Python: the VAE-family gold + model + determinism suites pass unchanged.

### Independent verification

An independent agent (which did **not** write this) re-derived every `matmul_*` call against the old scalar loops, ran the full FD suite (65/65), confirmed **bit-identical hashes across `RAYON`×`MATMUL` ∈ {1,4}²** (4/4 identical + run-to-run stable), ran the gold + model suites (**124 passed, 0 failed**), checked the numeric delta vs the exact baseline (`topic_word` max-rel **1.6e-15**), and produced the ~2.0× before/after against the correct merge-base. Verdict: *"a faithful, deterministic, and faster implementation — safe to PR."* Its one note (the CHANGELOG under-claimed at 1.76× vs the measured ~2.0×) is reconciled in the latest commit.

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,6 +48,22 @@ follow [Semantic Versioning](https://semver.org/spec/v2.0.0.html) once released.
   not affected — it has its own sparse per-document decoder, not the dense
   `theta·beta`. Encoder GEMMs and the elementwise `O(N·V)` work are untouched
   (candidate follow-ups).
+- **ProdLDA-family VAE encoder now uses a BLAS-free GEMM (1.76× faster
+  single-threaded)** (#378, sequel to the decoder GEMM above). The encoder's dense
+  layer-2 and head projections were per-document scalar loops; they now batch into
+  the same deterministic `matrixmultiply` GEMMs — forward `PRE2 = H1·W2ᵀ`,
+  `MU_RAW/LV_RAW = HD·W_mu/lsᵀ`; backward the `g.w2`/`g.w_mu`/`g.w_ls` cross-document
+  reductions (`dpre2ᵀ·H1`, `dmu/dlvᵀ·HD`) and `dH1`/`dHD`. Layer 1 is sparse and
+  mode-dependent (bag-of-words / embedding / CombinedTM `adapt_bert`), so it stays a
+  per-document accumulation in fixed document order (its forward matvec is now a
+  rayon-parallel, order-preserving map). A full ProdLDA fit is **bit-for-bit
+  identical across `RAYON_NUM_THREADS` and `MATMUL_NUM_THREADS` = 1/2/4**; output
+  differs from the old scalar path only at ~1e-15, so FD-gradient and
+  reference-parity tests are unchanged. Measured 47.7s → 27.1s single-threaded at
+  D3000/V2000/K20/H200 (on top of the decoder GEMM). Multi-core gain is modest
+  (~1.10× on 4 cores): with the encoder and decoder both GEMM'd, the remaining serial
+  `O(N·V)` reconstruction-softmax/batch-norm and the sparse layer-1 now dominate
+  (rayon-ing those is a candidate follow-up).
 
 ## [0.53.0] - 2026-07-26
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,7 +48,7 @@ follow [Semantic Versioning](https://semver.org/spec/v2.0.0.html) once released.
   not affected — it has its own sparse per-document decoder, not the dense
   `theta·beta`. Encoder GEMMs and the elementwise `O(N·V)` work are untouched
   (candidate follow-ups).
-- **ProdLDA-family VAE encoder now uses a BLAS-free GEMM (1.76× faster
+- **ProdLDA-family VAE encoder now uses a BLAS-free GEMM (~1.8–2× faster
   single-threaded)** (#378, sequel to the decoder GEMM above). The encoder's dense
   layer-2 and head projections were per-document scalar loops; they now batch into
   the same deterministic `matrixmultiply` GEMMs — forward `PRE2 = H1·W2ᵀ`,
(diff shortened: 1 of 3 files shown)


nealcaren:
## Double review (faithfulness + adversarial) + local verification — clean ✅

Reviewed by a correctness pass, an independent adversarial pass (Gemini pro-tier), and end-to-end local verification. **Both APPROVE, no blockers.** A faithful sequel to #586.

**Correctness verified:**
- Every encoder forward/backward GEMM call-site `(m,k,n)` matches the operand shapes and the `nn`/`nt`/`tn` contracts (PRE2 = H1·W2ᵀ, MU/LV_RAW = HD·W_μ/lsᵀ; backward g.w2/g.w_mu/g.w_ls tn-reductions, dH1/dHD nn).
- **Clobber-free (the key hazard):** all three encoder weight gradients (`g.w2`, `g.w_mu`, `g.w_ls`) GEMM into a fresh scratch buffer then `+=` into `g.*` — never a β=0 overwrite — so a coupled/pre-seeded caller (SCHOLAR/InfoCTM) is preserved, same pattern as #586's `g.beta`. `dHD = dmu·W_mu + dlv·W_ls` correctly sums both terms (second GEMM into `tmp`, then added).
- **Layer-1 bit-identity holds:** the parallel forward is an order-preserving pure per-doc map (no cross-doc reduction), and the `g.w1` backward accumulation stays a serial fixed-order sum. `encode_layer1` is a pure extraction — single-doc `transform` and all three input modes (bow/emb/adapt) unchanged.

**Ran locally:** all 26 `prodlda` FD-gradient tests, 53 VAE-family gold tests (ProdLDA/CombinedTM/ZeroShotTM/Scholar) — pass. Two-knob determinism confirmed: identical `topic_word` MD5 across the full `RAYON_NUM_THREADS × MATMUL_NUM_THREADS ∈ {1,2,4}²` cross-product (9/9 bit-identical).

Two non-blocking nits from the accuracy pass: a `pre1`/`h1` clone in the forward-assembly that could be a move (pure allocation, no correctness impact), and a suggested one-line comment noting the stored `pre2` is post-bias (so the backward `sigmoid(pre2)` argument reads correctly). Neither blocks.

nealcaren:
Thanks for the thorough double review. Pushed `eeb1aaa` addressing both non-blocking nits:

- **Move, not clone** — the `DocCache` assembly now consumes `l1` by value (`into_iter().enumerate()`), so `pre1`/`h1`/`adapted` move into the cache instead of being cloned (they're not read again). No correctness change.
- **Post-bias `pre2` note** — added a comment where `pre2_flat` is overwritten with the post-`b2` pre-activation, so it's explicit that the backward reads `softplus'(pre2) = sigmoid(pre2)` from that cached value (matching the old per-doc `sigmoid(dc.pre2)`).

`prodlda` FD-gradient suite still 26/26, fmt clean. Clean fast-forward on top of the reviewed `24006ee`, so the two-knob determinism / gold-parity results you verified are unchanged.

---
_Generated by [Claude Code](https://claude.ai/code)_

Issue #611, opened 2026-07-29
HDP/HLDA on large corpora: cosine agreement is meaningless for discovered-K, and HLDA tree-Gibbs doesn't scale to 18k docs


nealcaren:
Surfaced while building the `benchmarks/full_model_run.py` core fidelity+speed table (20NG, K=50 for the plain models). The two nonparametric/hierarchical models both misbehave at scale, in two distinct ways.

## 1. HDP — topic-word cosine agreement is not meaningful for discovered-K

On the full 18k-doc 20NG corpus, HDP produced:

```
HDP | tomotopy | K=auto | agreement=0.01 | topica 2.5s vs tomotopy 203.9s (82.5x)
```

The **0.01 agreement is not a faithfulness defect** — it is an artifact of the metric. topica's HDP and tomotopy's HDPModel each *discover* their own number of topics, and they land on different counts / different topic-word distributions, so an aligned topic-word cosine is near-zero by construction (same underlying phenomenon as the ETM/GSDMM discussion in #610, but more extreme because K itself differs).

**The speed number is real and striking** (topica 2.5s vs tomotopy 204s), but the agreement column is misleading.

**Options:**
- Score HDP with a discovered-K-appropriate metric (e.g. best-matching-subset cosine over min(K_a, K_b), or a document-clustering NMI like the embedding models use), OR
- Mark HDP/HLDA agreement as "discovered-K — topic-word cosine not comparable" in the table rather than printing a number.

## 2. HLDA — tree-structured Gibbs does not scale to 18k docs

topica HLDA (depth-3 tree Gibbs) on the full 20NG corpus ran for **3.2 hours at 100% CPU without completing** and had to be killed. The reference (tomotopy HLDAModel) cached fine during `--build-refs`; it is the **topica** side that does not scale here.

This is plausibly just an inherent-cost issue (nested CRP Gibbs over 18k docs × tree operations), but the ~3h-and-counting wall time is worth a look:
- Profile topica HLDA at, say, 2k / 5k / 18k docs to see whether it scales linearly or worse.
- If it's inherent, document a practical corpus-size ceiling for HLDA and have the benchmark run HLDA on a capped subset (like the covariate family runs on poliblog).
- If there's a superlinear hot spot (e.g. re-walking the full tree per token, reallocation per sweep), that's a real optimization target.

## Immediate benchmark workaround
[...]

PR #619, opened 2026-07-29
perf(hlda): tabulate log_gamma in the marginal precompute (~15x) (#611)

nealcaren:
## What

Profiling the nested-CRP path move showed the per-document marginal precompute
(`level_log_marginal` over all nodes) is **~95% of HLDA runtime**, dominated by
`log_gamma` calls. The previously-suspected `swap_remove`/`delete_empty_subtrees`
O(D²) hot spot is **0.1%** — a non-issue.

Every `level_log_marginal` argument is `integer_count + η` (word terms) or
`integer_count + V·η` (denominator), and no count exceeds the total token budget,
so precompute `log_gamma(k+η)` and `log_gamma(k+V·η)` once per fit and index them
instead of calling `log_gamma` per token per node. The tables are transient
(rebuilt each fit, not serialized).

## Result

| N (20NG) | before | after | speedup |
|---|--:|--:|--:|
| 1000 | 32.3s | 2.1s | 15.4× |
| 2000 | 94.6s | 6.8s | 13.9× |

Composes with the #616 threading (~4.7×). The 18k-doc corpus that ran **3.2h** in
#611 is now **~13min** — resolves the HLDA scaling half of #611.

## Faithfulness / safety

Same Dirichlet-multinomial marginal. Output is **bit-identical on every tested
corpus** — 800/1000/2000-doc subsets *and* the exact 2000-doc benchmark config
(same node count, assignment SHA, topic-word SHA), verified old-vs-new by checksum
— so the merged #618 cross-NMI row (0.213) is unchanged.

It is **not provably bit-identical for all inputs**: the table argument
`(k+m) as f64 + η` differs from the old `(k as f64 + η) + m as f64` by a few ULP
under float non-associativity (see the `lgamma_table_matches_direct_computation`
test, which asserts numerical agreement, not bit-equality). That never flipped a
sampling decision in any tested fit. Threading and fixed-seed determinism are
preserved. Fit-time memory now scales ~16 bytes/token (documented).

## Gate B (dual review)

- **Reviewer A (faithful — Codex):** approve, no blocker; marginal math correct,
  table bounds provably safe. Nits (test-comment wording, memory note) applied.
- **Reviewer B (adversarial — Claude opus; Gemini headless-blocked):** no blocker;
  independently confirmed no OOB path (`cap = total_tokens + max_doc + 1` is a
  rigorous bound since the doc is removed before the marginal), determinism holds,
  ULP framing honest. Memory-scaling note applied.

## Tests
12 Rust HLDA tests (incl. multithreaded bit-identity + new table-correctness test)
and 17 Python HLDA golden/level-prior tests pass; fmt/clippy/preflight clean.

### DIFF
--- a/src/hlda.rs
+++ b/src/hlda.rs
@@ -112,6 +112,13 @@ pub struct HldaModel {
     paths: Vec<Vec<usize>>,
     /// Per document, per token: the level 0..L it is assigned to.
     levels: Vec<Vec<usize>>,
+    /// Fit-time `log_gamma` lookup tables (see [`HldaModel::build_lgamma_tables`]).
+    /// `lg_eta[k] = log_gamma(k as f64 + eta)`, `lg_veta[k] = log_gamma(k as f64 +
+    /// V*eta)`. Every `level_log_marginal` argument is `integer + eta` or
+    /// `integer + V*eta`, so these replace the dominant `log_gamma` cost with a
+    /// table lookup. Transient (rebuilt each fit, not serialized).
+    lg_eta: Vec<f64>,
+    lg_veta: Vec<f64>,
 }
 
 impl HldaModel {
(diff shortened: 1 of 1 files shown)


nealcaren:
## Phase 5.5 — post-fix reference sanity check

Ran the shipped (optimized + Gate-B-fixed) build head-to-head against the live tomotopy reference on the real 2000-doc 20NG subset (the benchmark config):

| | speed | tree |
|---|---|---|
| topica (optimized) | 26.4s | 684 nodes |
| tomotopy (reference) | 0.8s | live_k=8 |

- **Accuracy: cross-NMI 0.213** vs tomotopy — identical to the merged #618 benchmark row, so the post-review fixes did not regress fidelity.
- **Speed vs reference improved ~13×**: pre-optimization topica was ~445× slower than tomotopy at iters=150; it is now 34.6×. The residual gap is the documented fine-tree (684 nodes at beta=0.01) vs tiny-tree (8 nodes) tradeoff, not a defect.

No regression; accuracy and speed claims reflect the shipped code.

Issue #10, opened 2026-06-06
Inconsistent first-arg conventions across toolkit (model vs matrix vs word-lists) + cryptic errors


nealcaren:
**Severity: Medium** · Found in external user-testing of v0.7.0.

Toolkit functions disagree on what their first argument is, and guessing wrong yields cryptic errors:

| Convention | Functions |
|---|---|
| Fitted **model** | `exclusivity`*, `representative_docs`, `document_intrusion`*, `word_intrusion`*, `plot_report`, `summary` |
| Raw **matrix** (`topic_word`/`doc_topic`) | `frex`, `relevance`, `topic_correlation`, `find_thoughts`, `label_topics` |
| **Top-word lists** | `coherence`, `topic_diversity` |

(* already accept the flexible `model_or_phi`/`model_or_theta` form — the right pattern.)

Example cryptic failure:
```python
topic_correlation(model)
# TypeError: float() argument must be a string or a real number, not 'topica.LDA'
find_thoughts(model, texts, topic=2)
# TypeError: float() argument ... not 'topica.LDA'
```

### Suggestions
- Extend the accept-either `model_or_phi`/`model_or_theta` pattern to `frex`, `relevance`, `topic_correlation`, `find_thoughts`, `coherence`, `label_topics`.
- At minimum, type-check the first arg and raise a helpful message: *"expected doc_topic matrix or fitted model; got LDA — did you mean `model.doc_topic`?"*

nealcaren:
Fixed in 5a74dda. frex, label_topics, relevance, topic_correlation, and find_thoughts now accept either a fitted model or the raw matrix as the first argument (vocabulary derived from the model when omitted), via the same coercion exclusivity/intrusion tests already used. coherence/topic_diversity already accepted a model. A bare matrix with no vocabulary now raises a clear 'vocabulary is required' message instead of the cryptic float() error. Matrix-first calls stay backward compatible.

PR #681, opened 2026-08-08
feat(diagnostics): embedding_coherence + inverted_rbo fit measures

nealcaren:
## What

Adds two intrinsic topic-model quality metrics that need **neither a reference corpus** (unlike the PMI coherences: `c_v`/`c_npmi`/`u_mass`/`c_uci`) **nor an LLM** (unlike the `topica.llm.*` suite). Together they close the sharpest gap against the OCTIS standard evaluation suite — the tier most relevant to topica's neural/embedding-model roster (ETM, ProdLDA, CTM, EmbeddingLDA, the embedding-cluster family).

### `topica.embedding_coherence(topics, word_embeddings, vocabulary=None, *, topn=10, method="pairwise")`
Per-topic coherence measured as top-word proximity in a word-embedding space (OCTIS `we_pairwise` / `we_centroid`; Ding, Nallapati & Xiang 2018). Returns a `(num_topics,)` array.
- `method="pairwise"` — mean pairwise cosine of the top-`topn` words.
- `method="centroid"` — mean cosine of each top word to the topic's word centroid.
- `word_embeddings` is a `{word: vector}` dict (matched by word) **or** a `(V, E)` matrix aligned to `vocabulary` — the same idioms `mmr()` and `align_topics()` already accept. Vectors are unit-normed internally.
- Unembedded words are dropped from a topic; a topic left with `< 2` embedded words scores `nan`.
- Docstring explicitly warns that scoring a model against **its own** learned embeddings (ETM/DETM/IdealPointTM `rho`) is circular — pass an external embedding for an honest number.

### `topica.inverted_rbo(topics, *, topn=10, p=0.9)`
Rank-weighted diversity: `1 - mean pairwise RBO` over topics' top-word **rankings** (OCTIS `InvertedRBO`; Bianchi, Terragni & Hovy 2021). Sharper than the set-based `topic_diversity` — two topics sharing their #1–#2 words are penalized more than two sharing their #9–#10. Returns a float (`nan` for `< 2` topics).

## Why

Inventory of topica's fit measures showed full corpus-coherence + a best-in-class LLM suite, but no embedding-space coherence and no rank-weighted diversity — the two OCTIS metrics that sit between "needs a corpus" and "needs an LLM". Both are cheap, intrinsic, and directly relevant to the neural models.

## Notes for reviewers

- The private `_rbo` kernel is **promoted** from `validation.py` into `coherence.py` so `inverted_rbo` and the existing `align_topics` RBO path share one implementation (`validation.py` now imports it). Behavior is byte-identical — the function body was moved, not rewritten.
- No Rust changes; pure Python/numpy.
- 16 new tests: hand-computed expected values (identical/orthogonal vectors, centroid = 1/√2, rank-sensitivity vs `topic_diversity`), matrix-vs-dict equivalence, normalization invariance, missing-word `nan`, model acceptance, and input-validation raises.

## Gate

- `./scripts/preflight.sh` — fmt/clippy/tables/stub all green
- `pytest tests/test_coherence.py tests/test_coherence_fast.py tests/test_validation.py tests/test_naming_conventions.py tests/test_input_validation.py` — pass
- `mkdocs build --strict` — exit 0


nealcaren:
## Three-way review + fixes (commit c309a9f)

Ran a faithfulness, adversarial, and first-time-user review. No crashes or math bugs in the core metrics; `inverted_rbo` and the `_rbo` relocation are faithful to OCTIS (`InvertedRBO`, `topk=10/weight=0.9`, `1 - mean rbo_ext`) and behavior-identical. Findings addressed:

**Accuracy**
- Citation fix: `we_pairwise` → Belford & Greene 2019 (was Ding et al.); centroid keeps Ding, Nallapati & Xiang 2018.
- `method="centroid"` intentionally reports cosine *similarity* (higher=better, unified with pairwise) rather than OCTIS's cosine *distance* — now documented as a deliberate convention, no longer claims OCTIS bit-parity.
- Documented the missing-word divisor divergence (topica averages over covered words; OCTIS divides by fixed `topn`).

**Robustness**
- Zero-norm / NaN / inf / tiny vectors are now genuinely *dropped* (treated as OOV) instead of being silently read as cosine ~0 or mis-normalized by the old `norm+1e-12` guard. "Words with no embedding are dropped" now actually holds.

**Usability**
- `nan` topics now emit a warning naming per-topic coverage; docstring/guide recommend `np.nanmean`.
- Added the missing "where do embeddings come from" link: `topica.llm_embed(model.vocabulary)` one-liner in the docstring + guide, and `embedding_coherence` added to `llm_embed`'s consumer list.
- Interpretation note (comparable only across models on the same embedding; not centered at 0) + a guide note on the `inverted_rbo` `topn=10` vs `topic_diversity` `topn=25` default mismatch; `inverted_rbo` added to the contrastive diversity prose.

+6 tests (zero/NaN/inf/tiny dropping, mutation safety, nanmean path). Gate green: preflight, coherence/validation suites, `mkdocs --strict`.

nealcaren:
Follow-up (eaf1185): made `method="centroid"` faithful to OCTIS `we_centroid`. The centroid is now built from the raw top-word vectors (as OCTIS does) rather than unit-averaged, and reported as similarity (`1 - distance`) to keep the house higher=better direction, so `embedding_coherence(method="centroid") == 1 - OCTIS_we_centroid` at full coverage. Both variants are now OCTIS-faithful; the only intentional divergence (documented) is partial-coverage averaging vs OCTIS's fixed-topn divisor. Added a faithfulness test on non-unit vectors to guard the raw-sum construction.

Issue #575, opened 2026-07-27
Add Mechanistic Topic Models (MTM): topics over sparse-autoencoder features (Zheng et al. 2025)


nealcaren:
## Model to add

**Mechanistic Topic Models (MTMs)** — "Model Directions, Not Words: Mechanistic Topic Models Using Sparse Autoencoders", Zheng, Beltran-Velez, Karlekar, Shi, Nazaret, Mallik, Feder & Blei (2025; arXiv:2507.23220, TACL forthcoming).
- Paper: https://arxiv.org/abs/2507.23220
- Reference code: https://github.com/blei-lab/mechanistic-topic-models (MIT license, Python/Poetry)

## What it is

Instead of modeling topics as distributions over **words** (bag-of-words, as in LDA/CTM/STM), MTM models topics as distributions over **interpretable features from a sparse autoencoder (SAE)** trained on an LLM's activations (the reference uses Gemma-2-9b SAE features). A document is represented by its SAE feature activations; topics are directions in that feature space. This yields:
- richer, more abstract topic descriptions (feature concepts, not just word lists), and
- **controllable generation via topic steering vectors** — the distinctive capability over classical/neural topic models.

Reference (MIT): consumes **pretrained** SAE features (it does not train the SAE); `scripts/train.py` / `scripts/evaluate.py` over config files; HuggingFace/OpenAI/DeepSeek APIs for feature extraction and LLM-based eval; evaluated on 8 datasets (20 Newsgroups, AG News, Yelp, Bills, Wiki, GoEmotions, PoemSum, WritingPrompts).

## How it fits topica

This is a natural fit for topica's **"you bring the features" pattern** — the same split we already use for the embedding-cluster and neural models (`Top2Vec`/`BERTopic`/`ETM`/`Scholar` take a precomputed embedding matrix; `topica.llm_embed` builds one). Concretely:

- **In scope for the Rust core:** the topic model *over a supplied document × SAE-feature matrix* (sparse, non-negative activations) — fit topics as feature distributions and documents as topic mixtures, with topica's numpy-native API and the shared contract (coherence, labeling, effects, save/load).
[...]

PR #685, opened 2026-08-09
feat(mtm): Mechanistic Topic Models — mLDA, mBERTopic, and a count-matrix entry point

nealcaren:
## What

A Mechanistic Topic Model ([Zheng et al. 2025](https://arxiv.org/abs/2507.23220), arXiv:2507.23220) replaces the bag of words with a bag of **sparse-autoencoder features**. An SAE decomposes a language model's residual stream into sparse, roughly monosemantic directions; count how often each fires in a document and you have a document-term matrix whose "terms" are things the *model* represents — so a topic becomes a claim about model internals, and can be steered rather than only read.

Refs #575. **This does not close it**: the issue also covers §4.2.2 (mETM, the one genuinely new inference core) and topic steering, both of which remain for a follow-up.

Adds:

- **`Corpus.from_matrix`** — the count-matrix entry point. Useful well beyond MTM: any already-vectorized corpus (a scikit-learn or gensim document-term matrix) now has a way in.
- **`topica.mtm`** — the featurization layer: `featurize`, `feature_thresholds`, `document_embeddings`, `FeatureCounts`.
- **`MechanisticLDA`** (§4.2.1) and **`MechanisticBERTopic`** (§4.2.3), both experimental-gated.

## How

**MTM is a family, not a model.** Featurize once, then run whatever topic model you would otherwise have run. Neither mLDA nor mBERTopic needs new inference and none is claimed here — the contribution at this layer is the *representation*. Both are thin wrappers over topica's existing reference-validated cores, carrying the authors' defaults. That is why the interesting code is in the featurization layer, not the models.

**`Corpus.from_matrix` preserves the caller's column contract**, which `from_documents` deliberately does not. `from_documents` derives the vocabulary from the data and frequency-sorts it; when column `j` is an SAE feature id or an external vectorizer's term, that index is load-bearing, so a silent reorder or drop corrupts every downstream lookup. Hence: column order preserved (`topic_word[:, j]` is your column `j`), all-zero columns kept (width is always `counts.shape[1]`), empty rows kept (`doc_topic` rows stay aligned with external metadata; `kept_indices` is the identity). `max_doc_fraction=` is the one operation that may change the width, and reports the survivors in the new `kept_features`.

**Two details in `featurize` are transcribed from the reference rather than inferred**, because getting either wrong silently changes every count:

- the comparison is **strictly `&gt;`** — an SAE emits many exact zeros, so `&gt;=` counts every silent token as a firing;
- each document's **first token is dropped**, before both the counting *and* the length (the SAE was not trained on the BOS position).

**`N_sae` and `N_tok` are kept as separate vectors.** The row sum counts feature *activations*; `n_tokens` counts *tokens*. A token may fire many features or none, so they differ — and the paper's equations do not use them interchangeably (eq. 8 wants `N_sae`, eqs. 12 and 14 want `N_tok`). The reference conflates them under one `lengths` field; keeping both is a deliberate divergence. On a fitted model, `doc_lengths` is `N_sae` and `corpus.n_tokens` is `N_tok`.

**Defaults follow the reference, not topica's usual ones**: `alpha_sum=5.0` (a MALLET *sum* over topics), `beta=0.01`, `optimize_interval=10`; `min_cluster_size=10`; and App. A.1's `max_doc_fraction=0.9` feature filter, which the paper calls crucial — a few SAE features fire on nearly every token and swamp every topic if kept.

**Deliberately out of scope.** The SAE forward pass: you bring the activations, matching how `BERTopic`/`Top2Vec`/`ETM` take a precomputed embedding matrix, and keeping PyTorch out of the dependency tree. `feature_thresholds` is offered as a *fallback* and says so — the paper takes `q_i` from the SAE's training data, so a within-corpus quantile is a different estimator.

### Divergences documented rather than papered over

`MechanisticBERTopic` cannot match the reference topic-for-topic even in principle, so the docstring, the guide, and the CHANGELOG all say why:

- **`nr_topics` counts differently.** The reference passes its `num_topics` to upstream BERTopic and then treats the last slot as the `-1` outlier topic, so its K *includes* the outlier cluster; topica's counts real topics only. Pass `K - 1`.
- **c-TF-IDF excludes outliers**, where upstream includes the `-1` cluster; and topica uses raw `tf` — literally eq. 15 — where upstream L1-normalizes first.
- **`doc_topic` is a different estimand**: HDBSCAN soft membership with outlier mass there, sliding-window c-TF-IDF approximation here. Cluster *labels* are the fair comparison.

Similarly, the reference's *default* mBERTopic embedding is not eq. 14 as written: it weights the same decoder directions by mean *continuous* activations rather than thresholded counts, and its hyperparameter sweep tunes that binary. Both are reachable (`directions=` for eq. 14, `doc_embeddings=` for anything else) and neither is presented as canonical.

## Validation

**mLDA parity arm**, driving the *shipped* path (`MechanisticLDA` → `Corpus.from_matrix`, with the column contract asserted inside the harness) against Java MALLET 2.0.8 `train-topics`. Planted-topic Zipfian count matrix, D=400, V=600, K=8, 800 iters, α_sum=5.0, β=0.01, settings symmetric on both sides. Scored by top-25 overlap (primary) and marginal-centered cosine, each reported against a corpus-marginal null and a 10-seed MALLET-vs-MALLET floor:

| `optimize_interval` | null | MALLET floor (min / median) | truth recovery (MALLET / topica) | **topica vs MALLET** |
|---|---|---|---|---|
| **10** (the reference's setting) | 0.1250 | 0.8350 / 0.8425 | 0.4850 / 0.5150 | **0.8700** (cos 0.9932) |
| 0 | 0.3300 | 0.6500 / 0.6625 | 0.5100 / 0.5750 | 0.6700 (cos 0.7713) |

topica sits **at or above MALLET's own seed-to-seed agreement at both settings**, and parity-minus-null (0.745 at `optimize_interval=10`) dwarfs the floor spread (0.0075).

Two honesty notes on that table. The claim is **relative, not absolute** — turning hyperparameter optimization on lifts the floor too (0.6625 → 0.8425), so `0.87` only means something next to the `0.8425` beside it. And truth recovery favours topica at both settings, but no floor distribution was computed for *that* comparison, so it is an observation, not a claim.

**What is still missing, and why both models ship experimental**: no end-to-end run against the authors' Gemma-2/SAE pipeline (that needs their activations and SAE), and no mBERTopic parity arm — the three divergences above mean cluster-label ARI is the only fair gate there.

**Tests**: 46 new in `tests/test_mtm.py`, 21 in `tests/test_from_matrix.py`. Each assertion is anchored to a specific paper equation or reference source line. Every planted-structure test is paired with the null it has to beat, and the file states up front that it does *not* claim numeric parity with the reference. No optional dependency: the cluster-recovery tests score with `topica.agreement`'s ARI, cross-checked against scikit-learn's in a dev environment (identical to 6 dp on both the planted case and the null).

**Kept current with `main`.** This branch has been re-merged and re-verified each time `main` moved, and it carries #702's unified coherence surface: both models forward `coherence_type=` / `texts=` to their inner model, so `m.coherence(coherence_type="c_npmi", texts=…)` equals `topica.coherence(m, …)` as it does roster-wide (verified at −0.2944 per topic). One caveat is documented on the method, because it is specific to this family: coherence asks whether a topic's top terms co-occur, and here those terms are SAE features, so a reference corpus must be in *feature* space — passing the original word documents would score feature ids against words and quietly return noise.

The full gate below is re-run locally after every merge from `main`, so these numbers track the head rather than the commit that first opened this PR. Most recent run: head `f6d0bfc` on `main` `d4f76c9`.

- [x] `cargo test --lib` passes
- [x] `python -m pytest tests/ -q` passes — 4791 passed, 76 skipped
- [x] `./scripts/preflight.sh` clean (fmt + clippy + generated-file sync)
- [x] docs build clean (`mkdocs build --strict`)
- [x] the `.pyi` stub is in sync (`test_stub_sync.py`, 102 passed)
- [x] CI green on all three platforms

## Notes

One reviewable judgement call worth flagging: `Corpus.from_matrix`'s `max_doc_fraction` is an **upper** bound, so a term occurring in *no* document is not ubiquitous and survives even at `0.0`. That reads as surprising until you say it out loud; it is tested and documented in both the guide and the docstring, but it is the kind of thing worth a second opinion.

`Corpus.from_matrix` is separable from the rest if you would rather land it apart — it stands on its own and depends on nothing MTM-specific.

Follow-ups, not in this PR: mETM (§4.2.2 — logistic-normal θ with a frozen SAE encoder and a Binomial likelihood, the one part of the paper that needs a new core), topic steering, and an mBERTopic parity arm against upstream BERTopic.


nealcaren:
Parking this PR: MTM is not under active development and the branch has fallen 13 commits behind `main`.

The work is preserved in full as the tag **`parked/mtm-575`** (commit f6d0bfc). An earlier attempt is at `parked/mtm-575-earlier`. To resume:

```
git fetch origin --tags
git checkout -b feat/mtm-575 parked/mtm-575
git rebase origin/main
```

then open a fresh PR. Issue #575 stays open, so MTM remains on the roadmap.

Issue #716, opened 2026-08-12
Gate B (retroactive) — KeyATM: covariate fit silently all-zero at iters≤249; docs report λ not predicted prevalence; family/parity overclaim


nealcaren:
Retroactive **Gate B** on `KeyATM`, model 4/6. Reviewers: A faithful (Codex `gpt-5.6-sol`), B adversarial (**Claude opus fallback** — Antigravity's headless run hit a write-permission denial; the fallback pulled the actual keyATM C++/R source), C comparative (Claude opus worktree), D sample-user (`load_poliblog`). Findings verified against code by the orchestrator; where A and B disagreed, the source-read reviewer won.

## Headline
The **base model is term-for-term faithful** to keyATM (B confirmed the α slice-sampler, keyword-switch conditional, and the information-theory weighting `-log2(f/total)` against keyATM's C++ source; determinism is bit-identical single- and multi-thread; the keyword inverted index and OOV handling are strong — an all-OOV keyword topic **raises** rather than silently degrading). C: KeyATM beats unguided LDA at concept separation (6/6 named concepts → distinct topics vs LDA's 4/6), best top-word diversity in the family, cost-center docs accurate. The problems are in the **covariate path** and **parity/claim honesty**.

### 1. Covariate fit silently returns an all-zero effect when `iters ≤ burn_in + optimize_interval − 1` (B#1, VERIFIED)
With the `fit` defaults `burn_in=200, optimize_interval=50` (`src/python/mod.rs:15056`), λ is optimized only when `iters ≥ 250` (`src/keyatm.rs:1896`). Below that it stays all-zeros → `α_{d,k}=exp(0)=1` → **`feature_effects` all zero, `feature_effect_se=None`, and NO warning.** Reproduced:
```
iters=240 → feature_effects all-zero=True, 0 warnings
iters=249 → feature_effects all-zero=True, 0 warnings
iters=250 → feature_effects all-zero=False (max|fe|=1.45)
```
A user doing a quick exploratory covariate fit (`iters=100`, say) gets a **silently null covariate result** headed for a table. keyATM samples Λ from iteration 1 and has no such dead zone. **Fix:** warn or error in the covariate branch when `burn_in + optimize_interval > iters` (mirror the existing empty-keyword-topic guard). This is the sharpest finding of the sweep.

### 2. Docs steer users to report λ (log-α coefficient), which disagreed with the actual prevalence effect (D-T1a, verified)
[...]

PR #723, opened 2026-08-12
KeyATM: covariate dead-zone warning + α/keyword guards + honest covariate docs (#716, part 1)

nealcaren:
Part 1 of the Gate B KeyATM findings (#716): the safety + honesty fixes. The covariate/dynamic parity golds (decision #4, "add the missing cases") are a follow-up PR.

## #716-#1 — the silent covariate dead-zone (the sharpest finding)
A covariate fit with `iters < burn_in + optimize_interval` (< 250 at the defaults) never optimizes λ, so `feature_effects` came back **all-zero with no warning** — a silently null covariate result headed for a table:
```
iters=240 → feature_effects all-zero, 0 warnings
iters=250 → feature_effects non-zero
```
Now the covariate branch warns loudly, naming the fix (raise iters, lower `burn_in`, or set `optimize_interval > 0`). The φ/θ topics are still valid, so it **warns rather than refuses** (your decision).

## #716-#6 — user `alpha` silently ignored under `estimate_alpha`
Under the default `estimate_alpha=True`, α is learned from keyATM's `50/K` start and a user's `alpha=` was silently dropped; the docstring also wrongly said α defaults to `1/K` "the starting point when estimate_alpha is on". Fixed the docstring (`1/K` is the *fixed* prior used only when `estimate_alpha=False`; the learned start is `50/K`, matching keyATM) and now **warn** when `alpha=` is passed under `estimate_alpha=True` (your decision).

## #716-#8 — categorical sampler NaN guard
`sample_index` summed weights and tested `r <= 0.0`; a NaN weight makes that never true, so the loop silently fell through to the **last index** — a biased draw. Added a finite-positive-total guard (debug-assert + unbiased uniform fallback in release), with Rust unit tests.

## #716-#5 — near-total keyword drop
A topic that lost most (not all) of its seeds kept its label but was effectively unseeded (`keyword_rate ≈ 0`), buried in a one-shot "some dropped" warning. Topics losing ≥ half their keywords now get a distinct **"weakly anchored / behaves like an unseeded topic"** warning with the stem-your-seeds guidance.

## #716-#2/#3 — covariate honesty (docs + docstrings)
`guided.md` now **leads with `predicted_prevalence`** (topic-share scale + CIs, what keyATM users report) instead of the raw λ / z-score, which understated the effect (a λ that reads "not notable" can still move the predicted proportion significantly). `feature_effects` is reframed as the log-α coefficient (sign, not a prevalence difference); the docstrings (Rust + `.pyi`) now state λ is **MAP-estimated** (not keyATM's per-sweep MCMC), `feature_effect_se` is an **asymptotic** SE (not keyATM's posterior SD), and `time_prevalence` is **normalized state-α** (not R's period-mean θ). Surfaced `visualize_keywords`/`refine_keywords` (#716-#9).

## Verification
- Full suite: **4879 passed**, 38 skipped. keyATM Rust units 27/27. `scripts/preflight.sh` clean; `mkdocs build --strict` clean. Verified the new `predicted_prevalence` doc example runs.

## Follow-up (PR-2, decision #4)
Extend `keyatm_gold.py` with covariate λ-magnitude / token-weight / α-π / dynamic HMM state-path parity against R keyATM, then tighten the "full family validated" wording to match.

### DIFF
--- a/docs/guides/guided.md
+++ b/docs/guides/guided.md
@@ -85,24 +85,46 @@ An intercept is prepended; the learned coefficients are in `feature_effects`.
 import numpy as np
 is_dem = np.array([...]).reshape(-1, 1)          # one row per document
 model = topica.KeyATM(seeds, num_topics=2, seed=1)
+# λ is optimized only on the sweeps after burn_in, so give the fit enough
+# iterations (>= burn_in + optimize_interval, 250 at the defaults) or
+# feature_effects come back all-zero — the model warns when they would.
 model.fit(docs, covariates=is_dem, feature_names=["is_dem"], iters=1000)
 
-model.feature_names       # ['intercept', 'is_dem']
-model.feature_effects     # (num_topics, 2): coefficient of each covariate per topic
-model.feature_effect_se   # same shape: standard error of each coefficient
-z = model.feature_effects / model.feature_effect_se   # |z| > ~2 ⇒ notable
+# Report predicted topic proportions at each covariate value, with CIs — the
+# interpretable, on-the-proportion-scale answer (R keyATM's predicted props).
+pp = topica.predicted_prevalence(
+    model, X=is_dem, feature_names=["is_dem"], at={"is_dem": [0, 1]}
+)
+print(pp)   # per topic (with topic_name): predicted share at is_dem=0 vs 1, 95% CI
 ```
 
-A larger `feature_effects[k, j]` means covariate `j` raises topic `k`'s
-prevalence. `feature_effect_se` gives the standard error of each λ, so you can
-tell a real effect from noise: it is the observed information of the same
-penalized Dirichlet-multinomial that `DMR` uses, computed in the standardized fit
-space (keyATM z-scores covariates internally, issue #270) and mapped back to the
-original covariate scale, so it is exact (no bootstrap) and computed once at fit
-time. An entry is `NaN` when its standardized coefficient hit keyATM's ±5 bound,
-where the constrained estimate has no valid asymptotic standard error. For
-uncertainty on the resulting topic prevalences, pair the fitted `doc_topic` with
+**Report `predicted_prevalence`, not the raw coefficient.** `feature_effects[k, j]`
+is the underlying log-α regression coefficient λ, *not* a difference in topic
+proportions: it lives on the `exp(x·λ)` prior scale, so its sign gives the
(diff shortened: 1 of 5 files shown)


nealcaren:
## Independent adversarial review

Run by a Claude adversarial reviewer (going straight to Claude given Codex's ghosting pattern on this repo). All six investigation points **verified empirically** — the reviewer built the branch, ran the new tests, and reproduced every boundary case.

**Verdict: correct and safe to merge. No blockers.**

| # | Area | Ruling |
|---|------|--------|
| 1 | Dead-zone threshold off-by-one | **Clean** — warning fires exactly when `iters < burn_in + optimize_interval`; confirmed at the boundary (bi=20, oi=10): `iters=29` warns + null, `iters=30` optimizes + non-null, `iters=31` non-null. `n_lambda_steps` equals the true optimize-event count. |
| 2 | Alpha warning + `py` param | **Clean** — `inspect.signature(KeyATM)` unchanged (`py` invisible to PyO3); warns only for explicit `alpha=` + `estimate_alpha=True`; `estimate_alpha=False` honors alpha silently; 50/K learned-start and 1/K fixed-prior both confirmed, docstring TRUE. |
| 3 | Categorical sampler guard | **Clean** — determinism byte-identical for valid weights (guard consumes no RNG, fallback not entered); `gen_range` safe at both call sites (len ≥ 4, ≥ 1); both unit tests pass. |
| 4 | Keyword-drop severity | **Clean** — ≥half→severe reproduced; single-keyword topics only skip-or-hard-error; `kept==0` correctly skips the soft warning and still hits the authoritative hard error; mild+severe independent. |
| 5 | Docstring/doc honesty | **Clean** — λ is L-BFGS MAP, SE is asymptotic observed-information, `time_prevalence` is normalized state-α — all match the implementation; the `predicted_prevalence` example runs and returns the documented shape. |
| 6 | Regressions | **Clean** — no `filterwarnings=error`; existing low-iter covariate tests emit the new warning but assert nothing about warning-freedom (37 passed, 1 skipped); the warning never fires for the base or dynamic-only path. |

Two cosmetic nits (non-blocking): the dead-zone message's `iters >= {need}` clause is paired with "or set optimize_interval > 0" so it's not misleading; the `#[should_panic]` NaN test is debug-profile-dependent (CI runs debug, so fine).

PR #724, opened 2026-08-12
KeyATM: covariate λ-magnitude/π + dynamic state-path R parity golds (#716, part 2)

nealcaren:
Part 2 of the Gate B KeyATM findings (#716), following #723. This adds the committed parity cases (decision #4, "add the missing cases") that the sign-only gold left unverified.

## #716-#4 — the gap
The committed `keyatm_gold.py` gated only base keyword-topic φ plus the **sign** of the covariate contrasts — not λ magnitudes, the keyword-switch π, or dynamic state paths — so "full family validated" over-generalized base keyword recovery. Extend the gold to gate the magnitude quantities against R keyATM, all **offline in CI** (the R fits are frozen in the committed fixture).

## New cases (regenerated on R 4.5.2 / keyATM 0.5.5)

| Case | Metric | topica vs R | bar |
|---|---|---:|---:|
| Covariate **λ magnitude** | corr of topica MAP `feature_effects` vs R posterior-mean λ (keyword topics) | **0.997** | ≥ 0.85 |
| Keyword-switch **π** | corr of topica `keyword_rate` vs R `pi$Proportion` (keyword topics) | **0.979** | ≥ 0.80 |
| Dynamic **state path** | adjusted Rand index of topica `time_state` vs R per-period HMM state | **0.517** | ≥ 0.30 |

- **λ / π** are compared on the keyword topics, which are index-aligned across engines because the supplied keywords pin the topic order. λ is topica's **MAP** vs keyATM's **MCMC posterior-mean**, so it's a magnitude correlation, not an exact match — 0.997 shows they land in the same place.
- **State path** is scored by **adjusted Rand index** (label-invariant): the HMM state labels are arbitrary, so the gold compares the *change-point structure* over periods, not the labels. The bar keeps a wide margin (0.30 vs 0.517) because the discrete partition metric is more platform-sensitive than the smooth correlations.

## Tests
- `test_keyatm_covariate_magnitude_parity`, `test_keyatm_dynamic_state_path_parity` — gate the new metrics offline.
- `test_keyatm_magnitude_and_state_gates_are_non_vacuous` — a mismatched pairing (averaged over 200 permutations, stable at only 4 keyword topics) collapses each metric below its bar, proving the gates discriminate.

## Verification
- `pytest tests/test_keyatm_gold.py`: **8 passed** (offline, no R). `scripts/preflight.sh` clean; `mkdocs build --strict` clean.
- The pipeline was de-risked end-to-end on a tiny corpus before the full regeneration (R λ shape (K,2) aligns with topica `feature_effects`; R π matches `keyword_rate`; R/topica state paths are the same partition).

The CLAUDE.md keyATM note (untracked) is updated to state precisely what is and isn't gold-gated (regular topics, α/weight magnitudes, and likelihood traces remain ungated).


nealcaren:
## Independent adversarial review + fixes

Review by a Claude adversarial reviewer (Codex ghosts on this repo). It reproduced every number from the committed gold and found one **blocker** plus refinements — all now addressed in a follow-up commit.

**#1 [BLOCKER, fixed] λ correlation was flattered by the intercept.** The reviewer showed the 8-point λ correlation (intercept + rating slope) scored **0.995 even for a fit with the rating slope forced to zero, and 0.982 for a sign-flipped slope** — so the gate rubber-stamped the very covariate effect it advertised. Fix: gate on the **rating-slope columns only** (intercept excluded). Honest gated value is now **λ rating-slope correlation 0.918** (was the intercept-flattered 0.997), and a sign-flipped fit now scores −1.0 and correctly fails.

**#2 [fixed] Free-floating bars → floor-derived.** The λ/π/state bars are now **R's own seed-to-seed floor − margin** (the same standard as the phi bar), computed from the stored floors, not hardcoded: λ 0.982−0.15=0.83, π 0.996−0.15=0.85, state 0.517−0.15=0.37. The gold was regenerated so the committed provenance records the slope-only λ (0.918) and floors, not the stale 0.997.

**#3 [addressed] Non-vacuity honesty.** With only **4 keyword topics**, a single λ/π correlation is genuinely coarse — a lucky topic permutation can align a wrong fit (the reviewer is right that `max < bar` is unachievable at n=4). The test now documents this limitation explicitly and asserts the **mean** shuffled |corr| sits below the bar (λ 0.49, π 0.53 vs bars 0.83/0.85; state ARI 0.01 vs 0.37) — a mismatch fails in expectation; the per-fit gate rests on topica clearing R's own floor.

**Clean (reviewer-confirmed):** index-alignment is correct (keyword topics are anchored → same order in both engines); the dynamic **state-path ARI is the honest, robust addition** (8 periods, no dominant-state inflation, topica 0.517 == R's own seed-to-seed 0.517); offline/CI safety, NA handling, and π scale-invariance all clean.

Fixes coming in the next commit; will re-run the gate.

Issue #730, opened 2026-08-14
Sample-user (sociologist) audit of model-agnostic search_k: pre-existing follow-ups


nealcaren:
First-time-sociologist sample-user audit (Dr. Priya Anand on `load_dubois`) run as the acceptance gate for the model-agnostic `search_k` feature (PR #729). She confirmed the feature works and is discoverable, and reached a defensible cross-model K choice. Two Tier-1 traps that PR #729 *exposed* by enabling NMF/LSA were fixed in that PR (garbage NMF `dispersion`; NaN `deveaud`/`cao_juan` + leaked RuntimeWarning on LSA). The items below are **pre-existing**, not introduced by #729.

### Tier 2 — broken / opaque
- [ ] **`corpus.metadata` is silently dropped by `save`/`load`.** ✓ `from_dataframe` attaches `corpus.metadata` (year/decade/author…), but after `corpus.save(p)` / `Corpus.load(p)` it is `None` with no warning. The natural prune-once → save → reuse-across-models workflow silently loses the covariates the `fit=` DMR/STM path needs — exactly the models a covariate researcher reaches for. Fix: persist `metadata` in `save`, or make `save`/`load` warn it is not carried.

### Tier 3 — return-shape / API consistency
- [ ] **`best_k("reconstruction_error")` is rejected though it is a reported column**, and the error advertises metrics not present in the result: `unknown metric 'reconstruction_error'; choose 'frontier' or one of [... 'perplexity', 'polarization' ...]` (neither is a column here). Fix: accept `best_k("reconstruction_error", rule="elbow")` (it is a scree curve; the elbow machinery exists), or enumerate the *actual* result columns in the error.
- [ ] **`Corpus` mixes properties and methods with no cue.** `num_docs`/`num_words`/`vocabulary` are properties but `documents()`/`save()`/`transform()` are methods; `corpus.num_docs()` → `TypeError: 'int' object is not callable`. Document the accessor style or make it consistent.

### What delighted (keep and advertise)
- The error messages are teaching tools (bad model name, `held_out=` on NMF, `prevalence=` with `fit=` all told her exactly what to do).
- `best_k` refuses to be an oracle — warns on grid-boundary and monotone-metric picks.
- The `fit=` hook is genuinely universal and worked first try (`fit=lambda k,s: DMR(k,seed=s).fit(docs, X)`).
- The vocab-pruning warning caught a real content trap (about to drop `negro`/`colored`/`people` on a Crisis corpus).
[...]

PR #731, opened 2026-08-14
fix(search_k): sample-user follow-ups — metadata save/load, reconstruction_error best_k, accessor doc (#730)

nealcaren:
Closes the Tier 2 / Tier 3 pre-existing follow-ups from the sociologist sample-user audit (#730). None were introduced by #729; all three came out of Dr. Priya Anand's `load_dubois` run.

## Tier 2 — `corpus.metadata` silently dropped by save/load
`from_dataframe` attaches `corpus.metadata` (year/decade/author…), but `Corpus.save`/`Corpus.load` dropped it silently, breaking the natural prune-once → save → reuse-across-models workflow the covariate `fit=` DMR/STM path depends on.

Now `save` pickles `metadata` to a `<path>.meta` sidecar and `load` reattaches it. Degrades honestly:
- a `save` with no metadata removes any stale sidecar (no wrong covariates on a later load);
- an unpicklable metadata object, or an unreadable sidecar, **warns** and continues rather than failing.

## Tier 3 — `best_k("reconstruction_error")` rejected
It is a reported NMF/LSA column but raised `unknown metric`, and the error advertised metrics not present in the result. It is now a selectable `minimize` metric: a scree curve, so `rule="best"` returns the grid edge and warns (like `perplexity`/`heldout_loglik`), and `rule="elbow"` finds the knee. It stays out of the frontier and the `best_k` default. The unknown-metric error now enumerates the columns *actually present* in the results.

## Tier 3 — `Corpus` property/method accessor cue
Documented the convention on the class docstring: facts about the corpus are properties (`num_docs`, `vocabulary`, `metadata`…), operations are methods (`documents()`, `transform()`, `save()`), so `corpus.num_docs()` vs `corpus.num_docs` is no longer a surprise.

## Tests
- `test_frames.py`: metadata round-trips through save/load; stale sidecar cleared on a metadata-free re-save.
- `test_search_k_models.py`: `reconstruction_error` is selectable (updated the prior assertion that it was not a direction).
- `test_issue_fixes.py`: scree `rule="best"` warns + `rule="elbow"` interior knee; unknown-metric error lists present columns.

Green: `cargo test --lib` (340), `preflight.sh`, `mkdocs build --strict`, and the touched pytest files.

### DIFF
--- a/docs/publishing/choosing-k.md
+++ b/docs/publishing/choosing-k.md
@@ -319,7 +319,9 @@ setting you sweep by **refitting per K**. `search_k` scans them directly with
 `model="nmf"` or `model="lsa"`, giving you the same coherence/exclusivity frontier
 and `best_k` machinery as LDA. For NMF it adds a `reconstruction_error` column (the
 fit's residual) — read it like a scree plot: it falls monotonically in K, so take
-the knee, not the minimum, and cross it against coherence.
+the knee, not the minimum, and cross it against coherence. Select on it directly
+with `rows.best_k("reconstruction_error", rule="elbow")` (bare `rule="best"` returns
+the grid edge and warns, because the error keeps shrinking with K).
 
 ```python
 import topica
--- a/python/topica/validation.py
+++ b/python/topica/validation.py
@@ -962,6 +962,11 @@ SEARCH_K_DIRECTIONS = {
     "heldout_loglik": "maximize",
     "perplexity": "minimize",
     "polarization": "maximize",
+    # NMF/LSA reconstruction error: a scree curve, monotone-decreasing in K, so
+    # rule='best' returns the grid edge (guarded below) and rule='elbow' is the
+    # useful pick. Kept out of the frontier and the best_k default; selectable
+    # only when named explicitly.
+    "reconstruction_error": "minimize",
     # Opt-in ldatuning-style criteria (search_k(criteria=...)).
     "deveaud": "maximize",   # mean pairwise JS divergence between topics
     "cao_juan": "minimize",  # mean pairwise cosine similarity between topics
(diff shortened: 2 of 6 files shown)


nealcaren:
## Sample-user acceptance audit — two independent first-time users

Ran the `sample-user` gate on this PR: **Dr. Priya Anand** (Claude, `load_dubois`, covariate-over-time + save/reuse) and **Dr. Marcus Bell** (Gemini, `load_poliblog`, NMF scree + save/reuse). Both walked the full arc blind (docs + docstrings only).

**All three fixes PASS on both users:**
- **(a) metadata save/load** — Du Bois: all 7 covariate cols round-trip bit-identical, refit LDA/NMF/STM off the reloaded corpus with no re-pruning. poliblog: `[SUCCESS] Metadata survived`, (2000,3).
- **(b) `best_k("reconstruction_error")`** — →grid edge with the actionable warning, `rule="elbow"` →interior knee; discoverable via `directions`, columns, and docstring. Both called the guardrail "exactly how a footgun should be handled."
- **(c) accessor convention** — both hit `TypeError: 'int' object is not callable` and resolved it instantly from the docstring, which predicts the exact error.

**One new finding on this PR's code, now fixed:** the audit caught that `Corpus.load` silently dropped metadata when only the `.corpus` file was moved and the `.meta` sidecar was left behind (corrupt sidecar warned; missing one was silent). Per the design call, I replaced the sidecar with a **single self-contained file**: `save()` appends the pickled metadata as a footer-tagged trailer after the corpus body. The CRP2 body reader ignores trailing bytes, so CLI-`preprocess` corpora and model `fit` are unaffected (topica-core round-trip tests still green), and there is nothing to orphan when the file moves. New tests cover the moved-file case and plain/CLI-corpus loads.

Pre-existing, unrelated friction both users noted (out of scope here): the STM `coef[0]` intercept trap and `best_k()` landing on the grid edge (both already documented/warned), `coherence(metric=)` / `bootstrap_stability` signatures, and `load_poliblog` `text_col` ambiguity.

Issue #752, opened 2026-08-15
Sample-user papercuts: top_words tuples, result-class exports, converged property vs method, coherence arg order


nealcaren:
Small ergonomics from the #732 sample-user run (first-time sociologist, STM on Du Bois's *The Crisis*). Individually minor; each is a "read around / guess" moment. Complements #732 / #742.

- [ ] **`top_words(n)` returns `(word, weight)` tuples, not strings.** `" ".join(model.top_words(10))` throws `TypeError: sequence item 0: expected str instance, tuple found`. The contract is documented (conventions guide, #746), but the plural name "top_words" still reads like it returns words and baited the reflex. Consider a `with_weights=False` (or a bare-string sibling) so the intuitive `join` works. *(Note: `top_words` is a Rust binding — a real signature change, not a doc fix.)*

- [ ] **The `_results` result classes aren't exported at top level.** `bootstrap_stability`/`quality_frontier`/`visualize_keywords`/`time_prevalence_ci` now return `FrameDict` subclasses (#749), but `topica.BootstrapStability` raises `AttributeError` — they live only in `topica._results`, so an `isinstance` check or a "what is this object?" inspection has to spelunk a private module. Export them at top level. *(Being fixed now.)*

- [ ] **`model.converged` is a property but `model.coherence()`/`model.bound()` are methods.** `model.converged()` → `TypeError: 'bool' object is not callable`. The property-vs-method line on the model object isn't guessable; document it (the `Corpus` docstring already does this for its own accessors — mirror that on the model side).

- [ ] **`model.coherence(...)`'s first positional arg is `n` (top-word count), not the corpus.** `model.coherence(corpus.documents())` → `TypeError: argument 'n': 'list' object cannot be interpreted as an integer`. The error names the arg so recovery is fast, but every other validation entry point takes `docs`/`texts` first, so the convention flip is a trap. (Upside: it defaults to the training corpus, so bare `model.coherence()` works.) Consider a docstring note calling out the arg-order difference from `topica.coherence(topics, texts, ...)`.

### Discussion (not obviously a bug)
- **`best_k()` returns the grid-edge K with a warning rather than auto-picking the elbow.** The boundary warning now names the elbow/frontier K (#750), which the user praised — but she still had to make a second call to act on it. Auto-defaulting to the elbow at a boundary optimum would remove a step; leaving it explicit keeps `best_k` predictable. Worth a decision either way.

[...]

PR #756, opened 2026-08-16
fix(usability): resolve the #755 sample-user follow-ups

### LINKED ISSUE #755: Sample-user follow-ups from #752 trial: converged naming trap, coherence error, stability call conventions, topic_table.to_frame
Friction surfaced by the pre-merge sample-user trial for #752 (PR #754), from two first-time researchers (Claude on `load_dubois`, Gemini on `load_gadarian`) running a full LDA workflow. These are **pre-existing**, not caused by the `top_words` change; splitting them out so #752 stays scoped. All items below reproduced on the current branch (`✓verified`).

### Tier 1 — wrong-signal trap
- [ ] **`model.converged` reads `False` after a normal, successful fit.** `✓verified`
  ```python
  m = topica.LDA(num_topics=3, seed=13).fit(docs, iters=100)
  m.converged        # -> False
  ```
  `converged` is `True` only if the fit early-stopped (`convergence_tol > 0` was hit); with the default (`convergence_tol == 0`, full `iters`) it is **always `False`**, so the attribute does not answer the question its name asks. A reviewer skimming a notebook would read "did not converge." #752 documented the property-vs-method distinction; the naming trap is separate. *Fix:* rename to `early_stopped` (keep `converged` as an alias), or make it a genuine convergence diagnostic.

### Tier 2 — intuitive call fails
- [ ] **`model.coherence(texts)` — first positional arg is `n`, not the reference texts; the error is cryptic.** `✓verified`
  ```python
  m.coherence(docs)
  # TypeError: argument 'n': 'list' object cannot be interpreted as an integer
  ```
  gensim muscle memory is `coherence(texts, ...)`. #752 documented the arg order; this asks the *error* to help. *Fix:* raise a directive error naming the keyword ("pass reference texts as `texts=`; the first positional is `n`, the top-word count").
- [ ] **`topic_stability` and `bootstrap_stability` have opposite call conventions.** `✓verified`
  ```python
  topica.topic_stability(docs, num_topics=3, seeds=[1,2,3])
  # TypeError: unexpected keyword argument 'num_topics'   (wants a list of fitted models)
  topica.bootstrap_stability(docs, num_topics=3, n_boot=4)   # takes corpus + K
  ```
  `topic_stability(runs, ...)` takes already-fitted models; its sibling `bootstrap_stability(docs, k, ...)` takes the corpus + K. *Fix:* give `topic_stability` a matching corpus+K+seeds overload, or cross-link the docstrings.

### Tier 3 — return-shape inconsistency
- [ ] **`topic_table(model)` returns `list[dict]`, not a DataFrame, and has no `.to_frame()`** — even though its docstring says "hand it to `pandas.DataFrame`". `search_k(...)` results *do* offer `.to_frame()`. `✓verified` *Fix:* give `topic_table` a `.to_frame()` for parity.
- [ ] **`corpus.documents()` is a method but `corpus.vocabulary` is a bare attribute** — calling `corpus.vocabulary()` by analogy raises `'list' object is not callable`. (Mirror of the model-side property-vs-method note; worth the same one-line docstring treatment on `Corpus`.)

### Tier 4 — discoverability / docs / defaults
- [ ] **Quickstart over-promises `search_k`.** It says search_k scores "coherence, exclusivity, and held-out likelihood," but the default result frame has only `coherence`, `exclusivity`, `dispersion` — held-out columns appear only if you pass `held_out=make_heldout(...)`. `✓verified` (columns: `k, coherence, coherence_metric, exclusivity, dispersion, dispersion_pvalue`). *Fix:* show the `held_out=` arg in the quickstart or drop the clause.
- [ ] **`best_k()` is opaque.** With non-monotone coherence (best at low K) and rising exclusivity, `best_k()` returns a middle value with no visible rule. *Fix:* `res.best_k(explain=True)` exposing the composite that produced the pick.
- [ ] **`topic_table` doesn't carry word weights, and nothing points to `top_words(n, weights=True)`.** To put probabilities beside words in a table you must independently know `weights=True` and zip it in — "the two halves don't meet." *Fix:* one docstring line linking `topic_table` -> `top_words(weights=True)`, or let `topic_table` optionally include weights.

### What delighted the sample users (keep these)
- Defaults that just work: `LDA(num_topics=10, seed=13).fit(corpus)` gave publishable topics with no tuning; documented determinism.
- FREX labels read like human labels; docstrings that pre-empt the exact mistake the user was about to make.
- `from_dataframe` keeps metadata row-aligned after pruning, so `topics_over_time` needed zero bookkeeping.
- Honest robustness: `bootstrap_stability` volunteered which topics were fragile.
- The #752 change itself: `weights=True` was discoverable "in seconds" from the docstring/quickstart.

_Surfaced by the sample-user gate on PR #754 (#752)._

nealcaren:
Closes #755. Addresses all nine friction items surfaced by the pre-merge sample-user trial for #752, across the four tiers.

## Tier 1 — wrong-signal trap
- **`early_stopped`**: every model gains an `early_stopped` getter that mirrors `converged`; `converged` stays as an alias. The flag reports early stopping (a positive `convergence_tol` was hit), not "the fit is good" — with the default (`convergence_tol=0`, full `iters`) it is always `False`, so the clearer name no longer reads as "did not converge". Added uniformly across ~46 getters + the `.pyi` stub.

## Tier 2 — intuitive call fails
- **`coherence(texts)`**: the first positional `n` is now a `TopN` extractor that raises a directive `TypeError` naming the `texts=` keyword when a caller passes the reference corpus first (gensim muscle memory), replacing the opaque `"'list' object cannot be interpreted as an integer"`. Applied to all 42 coherence methods.
- **`topic_stability` overload**: `topic_stability(docs, num_topics=K, seeds=[...])` now fits the runs for you, matching `bootstrap_stability`'s corpus+K convention. Both misuse shapes (`topic_stability(docs)` and stray fit kwargs) raise a directive error.

## Tier 3 — return-shape inconsistency
- **`topic_table.to_frame()`**: returns a `TopicTable` (list subclass) with `.to_frame()` for parity with `search_k`; still hands to `pandas.DataFrame` unchanged. `weights=True` adds aligned `prob_weights` / `frex_weights` columns.
- **`Corpus`** property-vs-method docstrings: `vocabulary` (bare attribute) vs `documents()` (method).

## Tier 4 — discoverability / docs
- **Quickstart** no longer over-promises held-out likelihood; it shows the `held_out=make_heldout(...)` opt-in and `best_k(explain=True)`.
- **`best_k(explain=True)`**: returns a `BestKExplanation` (int subclass — drops into `LDA(num_topics=...)` unchanged) carrying the per-K score table, the composite rule, and any boundary notes.
- **`topic_table`** docstring links to `top_words(weights=True)` and the new `weights=` arg.

`docs/contributing/conventions.md` documents the `early_stopped` alias and the directive coherence error.

## Verification
- Full suite: **4959 passed, 38 skipped** (matches the #752 baseline; no regressions).
- `scripts/preflight.sh` (fmt + clippy + NaN guards + model tables + stub-sync): all pass. `mkdocs build --strict`: passes.
- Live smoke tests of `early_stopped` (both fit paths), the coherence directive error, `topic_table.to_frame()`/`weights=`, `best_k(explain=True)`, and the `topic_stability` corpus overload all behave as intended.

### DIFF
--- a/docs/contributing/conventions.md
+++ b/docs/contributing/conventions.md
@@ -64,12 +64,19 @@ by construction rather than by memory.
    attribute, so `model.converged()` raises `TypeError: 'bool' object is not
    callable`. In contrast `model.coherence()` and `model.bound()` are called.
    (`Corpus` documents the same property-vs-method split for its own accessors.)
+   `converged` reports **early stopping**, not "the fit is good": it is `True`
+   only when a positive `convergence_tol` was actually hit, so with the default
+   (`convergence_tol=0`, full `iters`) it is always `False`. Read
+   `model.early_stopped` — the same value under the name that says what it means —
+   when the ambiguity would mislead; `converged` is kept as an alias (issue #755).
 7. **`coherence(...)`'s first positional argument is `n`, the top-word count — not
    the corpus.** It defaults to the training corpus, so bare `model.coherence()`
    works; pass a reference corpus with the keyword `texts=` for the windowed
    measures. This flips the convention of the module-level `topica.coherence(
    topics, texts, ...)`, whose first argument is the topics — so
-   `model.coherence(corpus.documents())` misfires (issue #752).
+   `model.coherence(corpus.documents())` misfires. Passing texts positionally now
+   raises a directive `TypeError` naming the `texts=` keyword rather than an
+   opaque "cannot be interpreted as an integer" (issues #752, #755).
 
 ## Threads and stopping rules
 
--- a/docs/contributing/conventions.md
+++ b/docs/contributing/conventions.md
@@ -68,7 +68,9 @@ by construction rather than by memory.
    only when a positive `convergence_tol` was actually hit, so with the default
    (`convergence_tol=0`, full `iters`) it is always `False`. Read
    `model.early_stopped` — the same value under the name that says what it means —
-   when the ambiguity would mislead; `converged` is kept as an alias (issue #755).
+   when the ambiguity would mislead; `converged` is kept as an alias. For a
+   plain-English summary of why a fit stopped, call `topica.stop_reason(model)`
+   (issue #755).
 7. **`coherence(...)`'s first positional argument is `n`, the top-word count — not
    the corpus.** It defaults to the training corpus, so bare `model.coherence()`
    works; pass a reference corpus with the keyword `texts=` for the windowed
--- a/docs/getting-started/quickstart.md
+++ b/docs/getting-started/quickstart.md
@@ -46,11 +46,17 @@ and set `text_col` to your text column.
 
 `num_topics` (K) is a research decision, not a tuning parameter. As a starting
 point, K=10 gives broad themes, K=30 finer ones. `search_k` scores a range of K
-on coherence, exclusivity, and held-out likelihood:
+on coherence and exclusivity by default; pass `held_out=make_heldout(corpus)` to
+add a held-out likelihood column too:
 
 ```python
 result = topica.search_k(corpus, ks=[5, 10, 20, 30], seed=13)
-print(result.best_k())
+print(result.best_k())                 # the coherence/exclusivity frontier knee
+print(result.best_k(explain=True))     # ...and the per-K scores behind the pick
+
+# add held-out likelihood as a selection criterion
+ho = topica.search_k(corpus, ks=[5, 10, 20, 30], seed=13,
+                     held_out=topica.make_heldout(corpus))
 ```
 
 The [choosing K guide](../publishing/choosing-k.md) walks through how to justify
--- a/python/topica/_topica.pyi
+++ b/python/topica/_topica.pyi
@@ -476,6 +476,12 @@ class DMR:
         ``check_every`` sweeps during :meth:`fit` (empty when ``check_every=0``)."""
         ...
     @property
+    def early_stopped(self) -> bool:
+        """Alias of :attr:`converged` under the name that says what it
+        means: True only if the fit early-stopped on ``convergence_tol``;
+        False when the full ``iters`` ran (issue #755)."""
+        ...
+    @property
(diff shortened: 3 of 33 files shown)


nealcaren:
## Dual review + real-data sample-user gate — results

Ran a three-agent gate: a faithfulness reviewer, an adversarial reviewer, and two first-time sample-user runs (one synthetic corpus, one on the real dwillis congressional-press data).

**Faithfulness:** clean — no defects; all 6 areas verified at runtime + stub-sync + clippy.

**Adversarial:** found 1 real regression + friendliness gaps, all now fixed:
- **[regression]** `topic_stability` consumed run 0 when `runs` was a generator (the `next(iter(...))` peek) → silent wrong answer. Now materializes the list.
- `coherence(float/negative/bool)` misdirected to `texts=` → now "n must be a non-negative int"; genuine wrong-type still gets the `texts=` hint.
- bare `topic_stability(Corpus)` and integer-id docs → now directive errors instead of "not iterable"/`IndexError`.

**Sample-user (real congress data):** confirmed the fixes smooth the workflow; two in-scope follow-ups folded in:
- `topic_stability` now accepts `k=` as an alias for `num_topics=` (matches `bootstrap_stability`).
- `converged`/`early_stopped` docs + conventions now point to `stop_reason(model)`; `topic_table` docstring clarifies `frex_weights` are ECDF FREX scores, not probabilities.

**Deferred to a separate follow-up issue (pre-existing, out of #755 scope):** `len(corpus)` / `Corpus.__len__`, a container-level `.to_frame()` on `estimate_effect`, and `topic_table` being a top-level function rather than a method.

Verification: full suite **4959 passed, 38 skipped**; preflight (fmt + clippy + NaN guards + tables + stub-sync) and `mkdocs --strict` clean.

Issue #757, opened 2026-08-16
Proposal: workflow namespaces with lazy legacy compatibility


nealcaren:
## Summary

`topica` now exposes 219 names at the package root, mixing model constructors, data preparation, selection, inspection, evaluation, effects, plotting, and LLM tooling. This makes discovery and completion noisy as the library grows.

Adopt **workflow-oriented namespaces as the preferred API** while preserving the familiar root-level model constructors and a small set of convenience calls. This is an additive migration, not a hard break.

## Proposed public shape

- `topica.data`: corpus construction and preparation (`Corpus`, `from_dataframe`, tokenization, phrases, stopwords)
- `topica.design`: covariate construction (`one_hot`, `design_matrix`, `spline`, `interaction`)
- `topica.select`: K/model selection (`search_k`, `select_model`, `ensemble`, cross-validation)
- `topica.inspect`: reading a fitted model (`topic_table`, `frex`, `relevance`, representative documents)
- `topica.evaluate`: credibility evidence (coherence/diversity/exclusivity, stability, held-out checks, intrusion, residuals, MCMC/conformance)
- `topica.effects`: prevalence/effect estimation and robustness
- `topica.compare`: alignment, drift, contrasts, fighting words
- `topica.embeddings`: embedding infrastructure and embedding-based analysis
- `topica.provenance`: `AnalysisManifest` and `record_fit`
- retain and strengthen `topica.llm`, `topica.viz`, `topica.datasets`, and `topica.scaling`

Keep memorable model constructors at the root: `topica.LDA`, `topica.STM`, `topica.BERTopic`, `topica.KeyATM`, etc. Do not force imports such as `topica.models.bayesian.STM`; an optional `topica.models` namespace can serve discovery/grouping only.

## The key split

The current `validation` surface is too heterogeneous. Split it conceptually into:

- **select**: choosing K or among fits
- **inspect**: reading and describing fitted topics
- **evaluate**: validation and diagnostic evidence

This mirrors the scientific workflow and separates interpretation aids from credibility tests.

## Compatibility plan

1. Add the task namespaces first as thin re-export modules; implementation files can remain where they are.
2. Document namespace paths as preferred in new examples, API docs, and the analysis skill.
3. Preserve legacy root aliases through a deliberate lazy compatibility layer. Module-level `__getattr__` (PEP 562) may help, but it is not by itself a no-risk solution.
[...]

PR #763, opened 2026-08-16
feat(757): add the nine workflow namespaces (additive)

nealcaren:
Third of the #757 workflow-namespace migration PRs. **Purely additive** — groups the flat API into task-oriented namespaces so the surface reads by analysis stage. Every existing `topica.X` keeps working unchanged; this only *adds* `topica.<namespace>.X`.

## New façade modules (thin re-exports, each with its own `__all__`)
- **`topica.data`** — `Corpus`/`tokenize`, `from_dataframe`, `align`, `prep_documents`, phrases, stopwords, `datasets`
- **`topica.design`** — `design_matrix`, `spline`, `interaction`, `align_corpus`, `one_hot`
- **`topica.embeddings`** — `EmbeddingLDA`, embedding I/O, `embedding_regression`, `AnchorLDA`
- **`topica.provenance`** — `AnalysisManifest`, `record_fit`, `check_conformance`, `list_models`

## Finalized from the PR #760 split
- **`topica.select` / `topica.inspect` / `topica.evaluate`** are now importable namespaces carrying their split surfaces; **`topica.effects`** was already one.

## `compare` becomes a callable module
`topica.compare(a, b)` still calls the function, and `topica.compare.CompareResult` reaches the namespace (a `ModuleType` subclass with `__call__`; `callable(topica.compare)` stays `True`).

## Other
- `one_hot` moves from an inline `__init__` def to its home in `topica.design`; the root re-exports it, so `topica.one_hot` is unchanged.
- `compare.py` imports its alignment helpers from `.evaluate` (their post-split home) instead of the shim.
- The nine namespace names are added to `__all__` and `__init__.pyi`. **The curated `__all__` trim + PEP 562 lazy root come in the next PR** — this one keeps the full flat surface intact.

## Verification
- Full suite: **4962 passed, 38 skipped**. Preflight + `mkdocs --strict` clean.
- Identity checks: `topica.X is topica.<namespace>.X` for a broad sampling; `topica.compare(a,b)` runs; `topica.one_hot` unchanged.
- Pre-merge: real-data (dwillis congress) sample-user gate on the new namespace paths.

### DIFF
--- a/python/topica/__init__.py
+++ b/python/topica/__init__.py
@@ -133,61 +133,7 @@ __citation__ = (
 )
 
 
-def one_hot(values, *, drop_first=True, reference=None, prefix=""):
-    """One-hot encode a categorical covariate for use as DMR features.
-
-    Given a sequence of category labels (one per document), returns
-    ``(matrix, names)`` where ``matrix`` is a ``(num_docs, num_categories)``
-    float array of 0/1 indicators and ``names`` are the corresponding column
-    names. One category is omitted as the reference (baseline) level, which
-    avoids collinearity with the DMR/STM intercept; every coefficient is then a
-    contrast *against that reference*, so which level is the reference sets what
-    the effects mean.
-
-    ``reference`` names the level to drop. When ``None`` (and ``drop_first=True``,
-    the default), the alphabetically-first category is dropped; with three or more
-    levels a warning names it, since which of several baselines is the reference
-    shapes the whole story and a silent choice is easy to misread — pass
-    ``reference=`` to choose it explicitly (and silence the warning).
-    ``drop_first=False`` keeps every level (full dummy set, e.g. for a model
-    without an intercept). Pass the result straight to ``DMR.fit(docs, matrix,
-    feature_names=names)``; combine multiple covariates with ``numpy.hstack``.
-    """
-    import numpy as np
-    import warnings
-
-    values = list(values)
-    categories = sorted(set(values))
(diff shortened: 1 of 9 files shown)


nealcaren:
## Real-data sample-user gate (dwillis congress) — PASS

A first-time user ran the full workflow on the 3,120-doc congressional-press corpus through the new namespace paths:

- **Flat API unchanged: PASS** — every namespace member is `is`-identical to its flat counterpart; `topica.compare(a,b)` still runs; `from topica import *` still works; no flat call changed behavior.
- **Namespaces work & discoverable: PASS** — the whole workflow runs through `data → design → select → LDA (flat) → inspect → evaluate → effects → compare → provenance`. `dir(topica.data)`/`dir(topica.inspect)` "read like a task menu."

It flagged two **cycle-safe gaps**, now fixed in the follow-up commit:
- `coherence`/`topic_diversity`/`exclusivity`/intrusion re-exported into `topica.evaluate` (a user looks for coherence in the *evaluate* stage).
- `cross_validate` + the held-out K-selection tools re-exported into `topica.select` (CV belongs next to `search_k`).
- Plus a `__dir__` on `select`/`inspect`/`evaluate` so tab-completion shows only the public surface (no `np`/`re`/`dataclass` leak) — the gate's one UX nitpick.

Identities preserved and no import cycle. Deferred (nitpick, for a later pass): the same `__dir__` polish on the other namespaces.

Issue #714, opened 2026-08-12
Gate B (retroactive) — NMF: doc_topic scale, count-default topics, effects-uncertainty traps + parity metric


nealcaren:
Retroactive **Gate B** on `NMF`, model 2 of the six marquee sweep. Reviewers: A faithful (Codex `gpt-5.6-sol`), B adversarial (Antigravity `gemini-3.1-pro-high`), C comparative (Claude opus worktree), D sample-user (`load_poliblog`). Findings verified against the code by the orchestrator before filing.

## Headline
The **unregularized MU core (β∈{Frobenius, KL}) is numerically sound** — Codex measured objective (reconstruction) gaps of **2.1e-5 (Frobenius) / 1.1e-4 (KL)** vs sklearn on the fixture, and C confirms NMF is healthy and beats LSA on topic quality. But the **default path does not hand a first-timer a correct, publishable analysis** (count-weighting, understated/again unavailable covariate uncertainty, a `doc_topic` scale question), and the **parity gate validates the wrong quantity** for a non-identified model.

### 1. `doc_topic` normalizes W independently of H's per-topic scale (A#6, verified `src/nmf.rs:1052`)
`doc_topic = normalize_rows(W)` and `topic_word = normalize_rows(H)` are computed **independently**. Since `W[:,k]·c` and `H[k,:]/c` are the same factorization, a topic's column-scale in W is arbitrary; the reported document mixture should weight each topic by its H-row mass (`doc_topic[d,k] ∝ W[d,k]·rowsum(H[k])`) before normalizing. As written, `doc_topic` ignores that scale, so the reported topic proportions — and any covariate-**prevalence** number built on them — can be distorted whenever topics carry different H-row magnitudes. The parity harness applies the same independent normalization to sklearn's W, so it cannot detect this. **Decision needed:** scale-correct `doc_topic` by H row sums (more principled, changes reported prevalence), or document that `doc_topic` is a raw-W row-normalization, not a reconstruction-weighted mixture.

### 2. `weighting='count'` default yields function-word topics (D-T1.1, C-note, A#4 — strong convergence)
[...]

PR #773, opened 2026-08-17
paper: refresh for the namespace default + the full model roster (stage 5)

nealcaren:
Refreshes `paper/topica.tex` after the namespace-default switch and the growth of the model roster. **Not ready to merge** — it carries `[?]` citation placeholders pending 18 new bib entries (see below).

## Done
1. **Namespaced API listings.** Every code block and inline `\code{}` reference uses the new default (`topica.select.*`, `topica.inspect.*`, `topica.evaluate.*`, `topica.effects.*`, `topica.design.*`); worked examples drop `from topica import stm`. Model constructors and `Corpus` stay at the top level.
2. **Counts + version.** "more than two dozen" → "more than fifty"; reproduction version `0.53.0` → `0.55.0`.
3. **Full 55-model table.** All 21 models shipped since the last submission added to their families, plus a new *information-theoretic* group (`CorEx`) and a family-intro paragraph naming the information-theoretic and relational families. The 3 topica-original models (`NarrativeTM`, `IdealPointTM`, `IdealPointSentenceTM`) cite `caren2026topica`.

`latexmk` compiles (exit 0).

## Needs you before merge

**18 new bib keys** — the table's external references render as `[?]` until these are added to `topica.bib` (map to your library / real PDFs):
```
arora2013practical  card2018neural  das2015gaussian  gallagher2017anchored
hoffman2010online  kangaslahti2026tensor  kristensenmclachlan2024keynmf
lacostejulien2008disclda  lauderdale2016measuring  mclachlan2024s3
mimno2009polylingual  paul2012factorial  rosenzvi2004author  titov2008multigrain
vendrow2021guided  wang2006topics  wang2007topical  yan2013biterm
```

## Deferred — needs the pinned reference environment (not local)
- **§5 validation appendix.** Regenerating locally corrupts it: this box has tomotopy 0.14 / scikit-learn 1.9, but the paper pins tomotopy 0.13 etc. Re-running shifted numbers for version reasons (NMF aligned cosine 0.98→0.59 under sklearn 1.9's NMF; HDP count→725 under tomotopy 0.14) — topica itself did not regress. Regenerate in the pinned environment.
- **§6 performance figures.** Deferred to the Longleaf run per discussion.

### DIFF
--- a/paper/gen_validation_appendix.py
+++ b/paper/gen_validation_appendix.py
@@ -644,87 +644,136 @@ def _sts_block(sts):
 
 
 def leg_hdp(k):
-    title = r"Hierarchical Dirichlet process (vs \pkg{tomotopy})"
+    title = r"Hierarchical Dirichlet process (vs the \pkg{blei-lab/hdp} equations)"
+    # HDP's reference is blei-lab/hdp, whose concentration-update equations topica
+    # reproduces (issue #611). We validate against an independent NumPy oracle of
+    # those equations rather than against tomotopy: tomotopy uses a *simplified*
+    # new-table weight that regularizes toward far fewer topics, so it is a coarser
+    # fit, not a competing ground truth. The oracle needs the 20NG corpus source.
+    try:
+        import hdp_blei_parity as hbp
+    except Exception as e:  # noqa: BLE001 — any import failure -> clean skip note
+        return skip(title, f"blei-lab HDP oracle unavailable ({e})")
+    try:
+        ids, str_docs, V = hbp.newsgroups_corpus(n_docs=60)
+    except Exception as e:  # noqa: BLE001 — sklearn/20NG absent
+        return skip(title, f"20NG corpus source unavailable ({e})")
+    from topica import HDP
+    ITERS = 80  # matches tests/test_hdp_blei_parity.py's validated band
+    dl = np.array([len(d) for d in str_docs])
+
+    # Independent NumPy oracle of the Escobar-West/Teh (2006) concentration updates.
+    ok, oa, og, oshare = hbp.oracle_hdp(ids, V, 0.1, 0.1, 0.01, iters=ITERS, seed=1)
+
+    # topica at its shipped default (resample_conc=True): concentrations inferred.
+    m = HDP(alpha=0.1, gamma=0.1, beta=0.01, seed=1)
+    m.fit(str_docs, iters=ITERS)
+    dt = np.asarray(m.doc_topic)
+    tok = (dt * dl[:, None]).sum(0)
+    tk, ta, tg = dt.shape[1], float(m.alpha), float(m.gamma)
+    tshare = float(tok.max() / tok.sum())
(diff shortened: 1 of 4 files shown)


nealcaren:
## §5 validation appendix — resolved in code, not by a pinned environment

The earlier note attributed the two shifted legs to reference-package drift
(tomotopy 0.14 / sklearn 1.9 vs the paper's pins) and deferred them to a pinned
box. That diagnosis was wrong. I reproduced both in a pinned env (topica 0.55.0,
`tomotopy==0.13.0`, sklearn swept 1.5→1.9) and neither is a version effect — both
are **intentional topica default changes** the generator never caught up to. The
library is correct; the appendix legs were stale. Fixed in
`paper/gen_validation_appendix.py`; no environment pin required.

**NMF (0.98 → 0.59).** The cosine is 0.591 under sklearn 1.5, 1.6, 1.7, 1.8, and
1.9 alike — sklearn didn't move. The cause is #714 (Gate B), which made tf-idf the
default NMF weighting. The leg feeds sklearn a raw *count* matrix but let topica
tf-idf-weight, so they factorized different inputs. Fix: pass `weighting="count"`
(what the parity test already does). Restores **0.980**, now identical under
sklearn 1.8 and 1.9 — version-independent.

**HDP (28 ≈ 37 → 725).** Under pinned tomotopy 0.13, tomotopy still finds 37; it's
**topica** that jumped, from #617 making concentration resampling
(`resample_conc=True`) the default. Pinning tomotopy does not fix it. But the
deeper issue is that the leg validated topica against *tomotopy*, which is the
wrong reference: tomotopy uses a simplified new-table weight that regularizes
toward far fewer topics. topica's real reference is blei-lab/hdp, and #617 already
shipped a NumPy oracle of its equations (`parity/hdp_blei_parity.py`). The leg now
validates topica's default against that oracle on a 20NG subset:

| engine (concentration mode) | K | α | γ | top share |
|---|---|---|---|---|
| topica (est. conc., default) | 267 | 8.04 | 65.6 | 33% |
| blei-lab/hdp equations (oracle) | 248 | 5.64 | 57.8 | 33% |
| topica (fixed conc., old default) | 11 | — | — | 51% |
| tomotopy (fixed conc., simplified weight) | 55 | — | — | 9% |

topica tracks the oracle (K within 8%, α/γ within a factor of two, identical
background share) and escapes the fixed-conc collapse; tomotopy is shown as
coarser context. Table A1's HDP row now reads `blei-lab/hdp eqns / 20NG subset`.

**Defaults left untouched.** Reverting either would re-break documented behavior
(HDP fixed-conc collapses to ~11 topics, cross-NMI ≈ 0; NMF tf-idf is the
deliberate Gate B default). Both changes are recorded in their commits, docs, and
tests. The paper, not the library, was out of date.

**Cross-check:** every other tomotopy/sklearn leg reproduces cleanly in the pinned
env — LDA 0.82/0.80, LSA 0.99, DMR 0.81, gDMR 0.81, sLDA 0.73, LabeledLDA 1.00,
PA 0.76, DTM 0.74. Only NMF and HDP had drifted.

PR #780, opened 2026-08-18
fix(gsdmm): Gate B follow-ups — auto-K honesty, find_thoughts, edge guards, perf (#717)

### LINKED ISSUE #717: Gate B (retroactive) — GSDMM: auto-K tracks K_max not the data; doc_topic mislabeled 'Eq.4'; 'reference validated' overclaim
Retroactive **Gate B** on `GSDMM`, model 5/6. Reviewers: A faithful-to-paper (Codex `gpt-5.6-sol` — GSDMM is planted/paper-validated, no maintained reference), B adversarial (Antigravity `gemini-3.1-pro-high`), C comparative (Claude opus worktree), D sample-user (`load_gadarian` short text). Verified against code/empirics by the orchestrator.

## Headline
The **Movie-Group-Process sampler is faithful to Yin & Wang 2014** (A#2, B independently: the popularity term, the repeated-word rising factorials in numerator and denominator, the `D−1+Kα` drop, and uniform init are all correct; α/β are `ensure_finite_pos`-guarded). C confirms GSDMM's design win: **perfect recovery (NMI/ARI 1.0/1.0) on separable short text**, beating mixed-membership PT/LDA, at the **lowest memory in the family (61 MB)** and the cheapest total fit at default iters. Determinism is solid and `num_threads>1` raises a clear `ValueError` (the right call). The findings are about the **auto-K narrative, a mislabeled `doc_topic`, and validation-claim honesty**.

### 1. `doc_topic` is an in-sample plug-in that can MISORDER clusters, but is labeled "Eq. 4 conditional" (A#1 + B#1, verified)
`doc_cluster_dist` (`src/gsdmm.rs:137`) scores each doc **without removing it** from the counts, so its own tokens inflate its current cluster. A gave the concrete failure: on the repeated-word fixture, public `doc_topic[0] = [0.179, 0.821]` while the correct **leave-one-out Eq.4** is `[0.829, 0.171]` — the not-removed scoring **flips the cluster order**. Empirically `doc_topic` is heavily over-peaked (median row-max **0.996**, 55% of rows >0.99, measured on gadarian). It's disclosed at `_topica.pyi:2693`, but calling it "recomputes Eq. 4" (`gsdmm.rs:131`) is mathematically incorrect. **Fix:** make `doc_topic` the leave-one-out Eq.4 conditional, or relabel it an in-sample plug-in projection and warn that it overstates confidence (and steer users to `doc_cluster`, per #4 below).

### 2. Auto-K tracks the `K_max` cap, not the data, and "empty clusters die out" oversells it (D-T1a/b/c + A#3 + C, verified)
GSDMM's selling point is inferring the cluster count, but on real overlapping short text it barely collapses:
```
gadarian:  K_max=10→10,  20→20/17,  30→19/21,  50→25   (retains ~2/3 of the cap)
beta at fixed K=20:  .001→4,  .01→20,  .1→17,  .5→7,  1→4   (beta silently controls the count, non-monotonically)
```
So the reported count tracks `K_max` and `beta` rather than the data; on separable synthetic text it *does* collapse cleanly (K_max=30→4/6), but on messy real data it does not. A adds that "empty clusters die out" (`docs/guides/short-text.md:9`) is too strong: with the faithful DMM `α=0.1`, an empty cluster keeps α-proportional mass and **can revive** — permanent pruning holds only for the paper's `α=0` variant. C measured the plateau as a *stable equilibrium* (identical at iters 30/100/300), so it isn't pure non-convergence — but it isn't the promised "collapse to the true K" either. `converged` is hard-coded `False` and the reported "log likelihood" is an assigned-cluster plug-in score, not the collapsed DMM marginal, so it's not a real convergence diagnostic. **Fix:** docstring/guide must caveat that the discovered count is sensitive to `K_max`/`beta`/regime and weak on overlapping data; warn at `fit()` when the final count ≥ ~0.9·`K_max` (cap too low); and correct the "die out" language.

### 3. `models.md` and `CLAUDE.md` claim "reference validated"; GSDMM is planted/paper-only (B#2b + A#6, verified — cross-cutting #660)
The generated model-table header `docs/guides/models.md:37` — "*Every model below is validated against a reference implementation.*" — and `CLAUDE.md:5` are **false for GSDMM and the other planted-only models** (`parity/gsdmm_gold.py` states planted-only; my `docs/contributing/validation.md` records it correctly). The `doc_topic` docstring also invokes a reference `gsdmm score()` that doesn't exist in-stack. **Fix:** soften the generated header prose in `registry.py` (and `CLAUDE.md`) to the tiered language `validation.md` already uses. This lands with the #660 batch.

### 4. `find_thoughts(model, corpus)` crashes (D-T2a, verified)
`topica.find_thoughts(m, corpus, topic=5)` → `TypeError: object of type 'topica.Corpus' has no len()`; the 2nd arg must be `corpus.documents()`, and the error names no parameter. **Fix:** accept a `Corpus` (call `.documents()` internally) or raise a typed, actionable message.

### 5. `topica.stopwords()`'s own docstring example throws (D-T4a, verified — cross-cutting first-step)
The docstring says pass the result to the corpus builders, but `Corpus.from_documents(docs, stopwords=topica.stopwords('english'))` → `TypeError: 'frozenset' ... cannot be converted to 'Sequence'`; you must wrap in `list(...)`. Every new raw-text user hits this on step one. **Fix:** accept `set`/`frozenset` in the builders, or fix the docstring example.

## Notes / follow-ups
- **6. Planted gold is weak** (A#5, B#2): the committed gold/unit corpus uses distinct-token docs (repeated-word rising factorial *is* covered by `test_model_invariants.py:407` and `test_seeded_gsdmm_contracts.py:298`, just not the gold), compares only `topic_word`, reports **5** clusters against a planted truth of **4**, and checks no hard-label (ARI) or inferred-K accuracy. Strengthen with an ARI-vs-labels + inferred-K-tolerance + leave-one-out-conditional fixture.
- **7. Tiny clusters reported as topics** (D-T1e): default fit yields singleton/doubleton clusters with full `topic_word` rows and near-perfect FREX (`books` exclusivity 0.998 *because* the words occur nowhere else) and non-NaN coherence, so they look legitimate. Provide/advertise a min-cluster-size filter and warn that FREX/coherence are meaningless on tiny clusters.
- **8. `doc_topic.argmax` ≠ `doc_cluster` on ~11%** of docs (D-T1d) — documented, but sharpen to "always use `doc_cluster` for GSDMM; `argmax(doc_topic)` can differ ~10%+."
- **9. `settings['num_topics']` (=`K_max` cap) collides with `model.num_topics` (=inferred count)** (D-T3a) — rename the settings key `max_topics`.
- **10. Low top-word diversity on real overlapping text** (C): 0.58 vs PT/LDA 0.95+ — an expected consequence of hard single-topic assignment on a narrow corpus; surface it for users comparing coherence/diversity dashboards.
- **11. Perf** (A#8): `doc_cluster_dist` allocates a dense V-length `wc` per doc and scans all V per cluster → effectively `O(D·K·V)` rather than the paper's `O(D·K·L̄)` on sparse short text; a sparse `(word,count)` list keeps the exact rising factorial and restores complexity.
- **12. Edges:** `beta=finfo.max` overflows `V·β` → all-zero `topic_word`, all-NaN `doc_topic` (A#7); an all-empty corpus leaks `NaN` into the convergence trace (B#3); `u32` counts can overflow on very large clusters (A). `topica.topic_diversity` rejects a raw `topic_word` matrix (needs a model/word-lists) — clear error, minor DX (C).
- **13.** Silent drop of 11/341 empty-after-pruning docs (`kept_indices` exists) (D-T4c); three-shaped reporting helpers (`label_topics` dicts / `top_words` lists / `find_thoughts` triples) (D-T3b).

## Comparative benchmark (Reviewer C)
| model | NMI/ARI (separable) | u_mass (real) | diversity (real) | fit 20k/sweep | peak RSS |
|---|---|---|---|---|---|
| **GSDMM** | **1.0 / 1.0** | −87.1 | 0.58 | 32 ms | **61 MB** |
| BTM | 1.0 / 1.0 | −86.7 | 0.60 | 20 ms | 69 MB |
| PT | 0.93 / 0.82 | −84.1 | 0.98 | 1327 ms | 117 MB |
| LDA | 0.93 / 0.82 | −74.1 | 0.95 | 13 ms | 123 MB |
GSDMM wins separable-short-text recovery + memory + total cost at default iters; loses coherence/diversity on real overlapping text. No regression.

_Gate B reviewers: A=Codex gpt-5.6-sol, B=Antigravity gemini-3.1-pro-high, C=Claude opus (worktree bench), D=Claude sample-user on load_gadarian. Six-marquee retroactive Gate B sweep, model 5/6._

nealcaren:
Closes #717.

Addresses the remaining live findings from the retroactive GSDMM Gate B. Several items had already shipped piecemeal before this PR (**F1** `doc_topic` relabel, **F3** validation overclaim, **F5** stopwords frozenset in builders); this PR covers the rest.

## Findings addressed

| # | Finding | Fix |
|---|---|---|
| **F2** | auto-K story oversold (GSDMM's headline selling point) | `fit()` warns when the discovered count sits at ≥90% of the `num_topics`/`K_max` cap (the cap, not the data, set the answer). The short-text guide no longer claims "empty clusters die out" — with `alpha>0` an emptied cluster can revive, and the count is sensitive to `K_max`/`beta`/separability. Framed as an upper-bound-conditioned estimate, not a guaranteed `K`. |
| **F4** | `find_thoughts(model, corpus)` crashed | `find_thoughts` / `find_thoughts_html` accept a `Corpus` (and a fitted model) — the natural call now works instead of raising an opaque `'Corpus' object is not subscriptable`. |
| **F7** | tiny clusters reported as full topics | `fit()` warns when singleton/doubleton clusters appear as topics (their FREX/coherence come from almost no data and look deceptively strong). |
| **F9** | `settings['num_topics']` (cap) collides with `num_topics` (discovered) | New **`max_topics`** getter reports the cap unambiguously. The `num_topics` constructor arg and `settings["num_topics"]` key stay — the cross-model naming convention (`test_num_topics_is_first_positional`) and the settings-mirrors-`__init__` contract (#400) both require it — with docstrings now cross-referencing both. |
| **F11** | `doc_cluster_dist` allocates a dense `V`-length vector and scans all `V` per cluster → `O(D·K·V)` | `doc_cluster_dist` and the hot `resample_doc` sampler iterate a sparse ascending `(word, count)` list → `O(distinct words)`, restoring the paper's `O(D·K·L̄)`. Ascending word order preserves the exact log-sum accumulation order, so results are **bit-for-bit identical** — locked by a dense-reference unit test, and the committed gold is unchanged. |
| **F12** | `beta=finfo.max` → all-NaN; all-empty corpus → NaN trace | `fit()` rejects an all-empty corpus and a `beta` so large that `V·beta` overflows to `+inf`, with directive errors. |
| **F6/F8/F10/F13** | weak gold; doc sharpening | Stronger planted-recovery tests (ARI vs planted labels + inferred-K tolerance + the in-sample over-peaked `doc_topic` signature); docs now always prefer `doc_cluster` for the label, note the expected lower top-word diversity of hard clustering, and point to `kept_indices` for row alignment. |

## Why F9 keeps `num_topics` rather than renaming

The finding suggested renaming the settings key to `max_topics`. I kept `num_topics` because (a) `test_num_topics_is_first_positional` mandates `num_topics` as every model's first constructor positional, and (b) `settings` is contractually keyword-named to match `__init__` (#400), so `settings["num_topics"]` must mirror the constructor. Adding a `max_topics` getter resolves the cap-vs-discovered collision without breaking either contract.

## Tests

- Rust: 340 pass, including new `gsdmm::tests::doc_cluster_dist_matches_dense_reference` (bit-identity of the sparse rewrite, incl. repeated-word rising factorial).
- pytest: **5003 passed, 38 skipped** (new `tests/test_issue_717_gsdmm.py`: 9 covering F2/F4/F6/F7/F9/F12).
- preflight (fmt/clippy/tables/stub/compat) green; `mkdocs build --strict` clean.
- GSDMM gold **unchanged**, confirming the perf rewrite changed no numbers.


nealcaren:
## Three-agent review gate

Ran a faithfulness reviewer + an adversarial reviewer on the diff, plus a sample-user driving GSDMM end-to-end on the **full** dwillis congress-press corpus (555,448 House releases, all 144 monthly files 2013–2024, no cap). Follow-up commit addresses everything real that surfaced.

### Held up under scrutiny (no changes needed)
- **F11 bit-identity proven**, not just argued: the faithfulness reviewer built `main` in a separate worktree and diffed a full fit — `topic_word`/`doc_cluster`/`doc_topic` bit-for-bit identical. Adversarial reviewer could not find a diverging input either. Gold unchanged.
- Beta-overflow / empty-corpus guards, `max_topics` accessor (incl. save/load), and the `find_thoughts` Corpus unwrap all survived adversarial edge-casing.

### Fixed in the follow-up commit
| Source | Finding | Fix |
|---|---|---|
| adversarial #1 | `iters=0` fired both warnings from the random init | warnings gated on `iters>0` |
| adversarial #2 | cap warning asserted "tracking the cap, not the data" — false on a clean cap==true-K recovery | reworded as a check to run (refit larger; settles below ⇒ inferred K, keeps filling ⇒ data supports ≥ this many) |
| adversarial #3 | tiny-cluster warning was near-permanent noise on short text (latent `-W error` hazard) | fires only on real fragmentation (≥2 tiny **and** ≥1/3 of clusters) |
| **sample-user (top finding)** | GSDMM on 555k *full-length* releases never inferred K below the cap and violated one-topic-per-doc, with no up-front hint | new **long-document warning** (avg >50 tokens) + short-text-only/single-threaded guide note |
| faithfulness #1 | `find_thoughts(model, corpus)` returned token *lists*, not prose | join to a display string (matches `find_thoughts_html`); docstring points to original texts + `kept_indices` |
| faithfulness #2 | beta-guard message imprecise (φ→0; NaN is in the log-denominator) | wording corrected |

### Sample-user verdict on the new warnings
- **Cap warning: helped** — fired at 50/50 and 5/5 and correctly stopped a newcomer from reporting "GSDMM discovered 50 topics" as an inferred count.
- **Tiny-cluster warning: correctly silent** on the large corpus (smallest cluster 1,349 docs).
- **`max_topics` vs `num_topics`: clear.**
- Dominant real friction was **model-for-the-job** (GSDMM is wrong for long docs) → the new long-doc warning now catches it before the ~10-min fit.

Full pytest **5007 passed, 38 skipped**; GSDMM suite green under `-W error::UserWarning` (no stray warnings); Rust 340 pass; preflight + `mkdocs --strict` clean.

Deferred (out of scope, worth a follow-up issue): a live `fit()` progress bar and a cluster×covariate crosstab helper.

Issue #781, opened 2026-08-18
GSDMM UX follow-ups from the #717 sample-user run: fit() progress + cluster×covariate crosstab


nealcaren:
Deferred from the three-agent gate on #717 / PR #780 (sample-user ran GSDMM on the full 555k-doc dwillis congress corpus). These were out of #717's scope but are worth doing; both surfaced as real newcomer friction.

### 1. No live progress during `fit()` — MAJOR newcomer blocker
GSDMM is single-threaded by design, so on a large corpus `fit()` runs silently for many minutes (the sample-user measured ~9 min for one cap=50 fit on 555k docs). A newcomer assumes it hung and Ctrl-C's it. `progress_interval` records the discovery trace but prints nothing live.
- Want: an opt-in live progress indicator / ETA during `fit()` (a `verbose=` flag or a tqdm-style bar), at least for the Gibbs sweep loop. Applies beyond GSDMM (any long single-threaded fit), so consider a shared mechanism.

### 2. No cluster×covariate crosstab helper — MINOR
For a party/time breakdown of clusters the sample-user hand-rolled `corpus.metadata.assign(cluster=model.doc_cluster).value_counts(...)`. It worked (metadata stayed aligned), but a `topica.inspect` one-liner for "cluster (or topic) × metadata column" crosstabs would be a natural reporting helper for single-membership models.

### Context / notes
- The corpus build itself was also heavy (~7 min, ~7 GB RSS, 138k-term vocab after `min_doc_freq=5`) with no progress feedback — related to (1); a general "building vocab… / N terms from M docs" progress line would help there too.
- These are the only items left after PR #780; the #717 correctness/honesty findings all shipped there.

nealcaren:
Both UX items shipped in #784: opt-in `GSDMM.fit(verbose=True)` per-sweep stderr progress (with elapsed time), and `topica.inspect.topic_crosstab` for topic×covariate reporting. The general progress rollout (live bar + sparkline, keyATM/LDA first) is tracked in #785 (Phase 1) and #786 (Phase 2).

PR #783, opened 2026-08-18
perf(gsdmm): memoize empty clusters + cache ln(beta) — ~2.5x faster sampler (#781)

nealcaren:
Addresses the runtime pain from the #781 GSDMM sample-user run (fits ran silently for minutes on the full dwillis corpus). GSDMM is hard single-threaded — its Movie Group Process discovers K through within-sweep reinforcement, so AD-LDA-style parallelism would make the discovered count thread-dependent — so this speeds up the per-document work instead, **without changing any draw**.

## What changed

Factored the per-cluster Eq. 4 log-prob loop (shared by the `resample_doc` sampler and `doc_cluster_dist`) into one `cluster_log_probs` helper with two **bit-exact** shortcuts:

- **Empty-cluster memoization** — every cluster with `m[k]==0` has `n[k]==0` and all `nw[k][*]==0`, so its log-prob is a single value independent of `k`. GSDMM keeps `k_max` clusters and empties most of them, so this collapses the dominant cost from `O(k_max)` to `O(#non-empty)` per document. The win scales with how generous your cap is.
- **`ln(beta)` caching** — for a word absent from a cluster the `j=0` numerator term is `ln(0+beta)`; most (word, cluster) pairs are absent, so this skips the bulk of the numerator's `ln` calls.

Both reuse values that are **bit-identical** to recomputing them inline (`0.0+x==x` exactly), so the sampler draws and the committed gold are unchanged.

## Bit-identity — proven two ways
1. The dense-reference Rust test now fits `K=12` with many empty clusters and still matches an independent naive per-cluster computation to the bit (`to_bits`).
2. A full 25-sweep fit (with repeated tokens) is `doc_cluster`- and `doc_topic`-**bit-identical** between this branch and the pre-change `main` build (worktree cross-check).

## Benchmarks (15 sweeps, K_max=50)

| corpus | before | after | speedup |
|---|---|---|---|
| structured short text, collapses to 8 clusters | 1.00s | **0.39s** | 2.6× |
| longer docs, discovers 15 | 1.57s | **0.66s** | 2.4× |

## Tests
Rust 340 pass (incl. the strengthened bit-identity test); full pytest **5010 passed, 38 skipped**; gold unchanged; preflight + `mkdocs --strict` green.

## Still open on #781
The two originally-filed UX items remain: a live `fit()` progress indicator (this speedup shrinks the silent window but doesn't remove it) and a cluster×covariate crosstab helper. Leaving #781 open for those.

### DIFF
--- a/src/gsdmm.rs
+++ b/src/gsdmm.rs
@@ -58,6 +58,77 @@ fn word_counts_sorted(doc: &[u32]) -> Vec<(usize, u32)> {
     out
 }
 
+/// Unnormalized log `p(z_d = k)` for every cluster `k` (Yin-Wang Eq. 4), given
+/// the current counts and this document's sorted `(word, count)` list.
+///
+/// Two bit-exact speedups over the naive per-cluster/per-word double loop (#781):
+/// - **Empty-cluster memoization.** Every cluster with `m[k]==0` has `n[k]==0` and
+///   all `nw[k][*]==0`, so its log-prob is a single value that does not depend on
+///   `k`. GSDMM keeps `k_max` clusters and empties most of them, so this collapses
+///   the dominant cost from `O(k_max)` to `O(#non-empty)` per document.
+/// - **`ln(beta)` caching.** For a word absent from cluster `k` (`nw[k][w]==0`) the
+///   `j=0` numerator term is `ln(0+beta)=ln(beta)`; most (word, cluster) pairs are
+///   absent, so this skips the majority of the numerator's `ln` calls.
+///
+/// Both reuse values that are bit-identical to recomputing them inline (`0.0+x==x`
+/// exactly), so the sampler draws and the committed gold are unchanged.
+fn cluster_log_probs(
+    m: &[u32],
+    n: &[u32],
+    nw: &[Vec<u32>],
+    wc: &[(usize, u32)],
+    n_d: u32,
+    alpha: f64,
+    beta: f64,
+    vbeta: f64,
+) -> Vec<f64> {
+    let k = m.len();
+    let ln_beta = beta.ln();
+    let mut empty_lp: Option<f64> = None;
+    let mut out = vec![0.0f64; k];
+    for kk in 0..k {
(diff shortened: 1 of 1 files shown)


nealcaren:
## Three-agent review gate

Ran faithfulness + adversarial reviewers on the diff and a sample-user timing GSDMM on the full 555k-doc dwillis corpus.

### Bit-identity: confirmed both ways
- **Faithfulness**: proved the empty-cluster invariant from the code (`m/n/nw` mutate in lockstep, so `m[k]==0 ⟹ n[k]==0 ∧ nw[k][*]==0`), verified `0.0+x==x` exactness and accumulation order, and did an independent worktree cross-build (K=40, 35 empty clusters) — `doc_cluster`/`doc_topic` bit-identical.
- **Adversarial**: could not break it across **421 differential-fuzz configs** (repeated tokens, β∈[1e-30…5], α∈[1e-6…50], K up to 150, dense-fill, iters=0, max word-id) — all bit-identical; 56 pytest pass under `-W error`. Added its suggested `debug_assert` on the invariant as insurance (commit 71879a9).

### Honest scope of the speedup (sample-user finding)
The ~2.5× is **real but regime-dependent**. It wins when there are **many empty clusters** (K_max ≫ discovered) and **short docs** (small denominator) — i.e. GSDMM's *intended* short-text use with a generous cap that collapses.

The sample-user's full-corpus run is the opposite regime and shows **no measurable gain**: K=50 on 555k **long** (201-token) releases **filled the cap** (50/50, zero empty clusters → empty-memo does nothing; the untouched per-token denominator dominates on long docs) → 567.6s vs ~560s baseline. That regime is GSDMM misused on long text (which the long-document warning flags) — outputs were sensible and all warnings fired correctly.

**Net:** bit-identical, zero downside, genuinely faster for GSDMM's designed short-text-collapsing case; does not help the long-doc/cap-filled misuse. Corrected the earlier benchmark framing accordingly.

PR #784, opened 2026-08-18
feat(gsdmm): fit(verbose=) progress + topica.inspect.topic_crosstab (#781)

nealcaren:
The two UX follow-ups from the #781 GSDMM sample-user run (the speedup shipped separately in #783).

## 1. `fit(verbose=True)` progress
The sample-user's top pain was a ~9-minute fit that ran **silently** and "looks hung." `GSDMM.fit` gains an opt-in `verbose=False`; when `True` it prints a per-sweep line to **stderr** at the recorded-trace cadence:

```
[GSDMM] sweep 4/10  3 clusters  ll=-1.1023
```

Off by default (tests and pipelines stay quiet), stderr not stdout, and verified it does **not** change the draws (bit-identical `doc_cluster` with/without).

## 2. `topica.inspect.topic_crosstab(model, metadata, column)`
The "which topics do Democrats vs Republicans write?" table the sample-user had to hand-roll. Returns a pandas crosstab of **dominant-topic × covariate**:

```python
topica.inspect.topic_crosstab(model, corpus, "party")                     # counts
topica.inspect.topic_crosstab(model, corpus, "party", normalize="index")  # row %
```

- Uses `doc_cluster` for hard-clustering models (GSDMM); `doc_topic.argmax` for mixture models.
- Accepts a `Corpus` (row-aligned `.metadata`), a `DataFrame`, or a bare array/Series.
- `normalize=` follows `pandas.crosstab`; a length mismatch raises a directive error pointing at `corpus.kept_indices`.

## Tests / docs
6 new tests in `test_issue_781.py` (progress fires/quiet/no-op-on-draws; crosstab hard vs mixture, bare-array + length check). `short-text.md` documents both. Full pytest **5016 passed, 38 skipped**; preflight + `mkdocs --strict` green.

## Scope note
`verbose` is GSDMM-only (the model the audit hit); a shared progress mechanism for every long fit is a larger cross-model follow-up. With these two items, #781 is complete.

### DIFF
--- a/docs/guides/short-text.md
+++ b/docs/guides/short-text.md
@@ -17,8 +17,21 @@ cap nor honours its one-topic-per-document assumption; reach for `LDA` or `STM`
 there instead, or reduce each document to a short field (a title or first
 sentence). `fit` warns you when the average document is long. It is also
 **single-threaded by design** (its cluster-count discovery is inherently
-sequential), so a large corpus fits slowly and silently — subsample while you
-explore, and raise `progress_interval` to watch the discovery trace.
+sequential), so a large corpus fits slowly — subsample while you explore, and
+pass `fit(..., verbose=True)` to print per-sweep progress (iteration, cluster
+count, log-likelihood) to stderr so a long fit does not look hung.
+
+To report which topics load on a covariate — the "which topics do Democrats vs
+Republicans write?" table — cross the assignment against a metadata column:
+
+```python
+topica.inspect.topic_crosstab(model, corpus, "party")                 # counts
+topica.inspect.topic_crosstab(model, corpus, "party", normalize="index")  # row %
+```
+
+It uses `doc_cluster` for GSDMM (and any hard-clustering model) and the dominant
+`doc_topic` topic for mixed-membership models; `corpus.metadata` stays row-aligned
+through pruning, so no manual realignment is needed.
 
 Two caveats to read the discovered count honestly. First, with the default
 `alpha=0.1` an emptied cluster keeps `alpha`-proportional mass and can revive, so
--- a/python/topica/_compat.py
+++ b/python/topica/_compat.py
@@ -110,6 +110,7 @@ LAZY = {
     'stopwords': 'data',
     'top_topics': 'effects',
     'topic_correlation': 'inspect',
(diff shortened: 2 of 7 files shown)


nealcaren:
## Three-agent review gate

Faithfulness + adversarial on the diff, plus a sample-user running both features on a 40k-doc dwillis slice. All findings fixed in the latest commit.

### Both features validated
- **`fit(verbose=)`**: faithfulness + adversarial both confirmed it's **bit-identical** to non-verbose (same `doc_cluster`/`doc_topic`), stderr-only (doesn't pollute stdout capture or trip `-W error`), correct cadence, and robust to `iters=0`, empty corpus, `progress_interval` 0/1/huge, and the deprecated `report_interval`. Sample-user: *"directly kills the 'is it hung' anxiety."*
- **`topic_crosstab`**: sample-user verified it `.equals()` the hand-rolled `metadata.assign(...).value_counts()` path and "just works" off the Corpus with no alignment wrangling.

### Fixed from the review
| Source | Finding | Fix |
|---|---|---|
| adversarial (severe) | NaN/None covariate values silently dropped → under-count | `dropna=False`; missing values kept as a visible column |
| adversarial | mixture topic never dominant → missing row / index gaps | reindex to full topic range; rows now align with `topic_table` by index |
| adversarial + faithfulness | wrong column → raw `KeyError` | directive `ValueError` listing available columns |
| adversarial | 2-D covariate → opaque `ArrowInvalid` | clear "covariate must be 1-D" |
| faithfulness | model with neither `doc_cluster`/`doc_topic` → `AttributeError` | typed `TypeError` |
| sample-user | verbose gives no sense of pace | progress line now shows elapsed wall-clock `(12.3s)` |
| adversarial (suspected) | `Series` aligns positionally | documented in the docstring |

### Deliberately not done (scope)
The sample-user also wanted optional top-words **labels** on the crosstab rows (it returns integer IDs) and friendlier `normalize=` aliases — left as small follow-ups; the rows already align with `topic_table` by index, and `normalize=` mirrors pandas.

Full pytest **5019 passed, 38 skipped**; preflight + `mkdocs --strict` green.

Issue #854, opened 2026-08-30
feat(threadtm): seed words (β) and prevalence anchors (θ) for ThreadTM


nealcaren:
## Capability

ThreadTM is a logistic-normal topic model with a reply-tree prior. Its novelty is the θ prior (how prevalence flows parent→child); the topic-word side (β) is a plain variational update. That makes ThreadTM a natural home for **user supervision**, and the two knobs are orthogonal to the reply tree:

1. **Word seeds (β).** `seed_words={topic_index: [keywords]}` biases the seeded topics' word distributions toward supplied keywords (SeededLDA-style Dirichlet pseudocounts). Steers *what topics mean* and pins them to fixed slots (solving unsupervised label-switching). Frequency-scaled weights (`seed_prior="frequency"`, pseudocount ∝ the word's corpus count) keep it a *soft* prior that learns beyond the seeds instead of collapsing to them.

2. **Prevalence anchors (θ).** `prevalence_anchor={group_index: [K-mix]}` shrinks a covariate group's baseline toward a supplied target topic mix. Steers *topic prevalence* with a smooth `anchor_strength` knob.

## Why

A domain scientist working a threaded corpus usually arrives with themes they care about ("astronomy, physics, biology, trades") and wants topics *for those*, in fixed, comparable slots — not an arbitrary unsupervised permutation. Seeding also stabilizes ThreadTM's identification on the ambiguous, short-leaf comments that are exactly where the reply tree is supposed to help.

## Validated (design spike)

On `load_threads` (askscience + pokemontrades, 5,042 comments), K=8, seed 4 topics leave 4 free:
- Seeded topics land on theme and pull in related non-seed words (astronomy → speed/velocity/momentum; pokemon → thanks/ready/want).
- Mean c_npmi coherence **rose** +0.037 → +0.052 — seeding did not degrade the fit.
- Seeded-topic subreddit prevalence clean (pokemon 0.786 in pokemontrades vs 0.010 in askscience).
- Anchor steers monotonically toward the target as `anchor_strength` rises.

Synthetic planted-recovery (K=4 disjoint blocks): unsupervised recovers topics in an arbitrary permutation; seeding 3/10 block words pins each to its slot as a soft prior (learns the other 7); anchor cuts prevalence L1-to-truth 0.030 → 0.002.

## Scope for the PR

[...]

PR #862, opened 2026-09-01
feat(threadtm): seed-word API parity with SeededLDA (#854)

nealcaren:
Makes ThreadTM's user-supervision surface consistent with the other keyword models (SeededLDA / keyATM / CorEx / GuidedNMF). ThreadTM is pre-1.0 and experimental, so these are hard renames with no back-compat shims.

## Changes

**1. Seed weight: `seed_weight` (1.0) → `weight` (0.01), SeededLDA scale.**
Adopts SeededLDA's `weight` name, `[0, 1]` bound, and `× 100` pseudocount scaling, so the same number means the same seeding strength in both models. **Default behaviour is unchanged** — old `seed_weight=1.0` and new `weight=0.01` both land at 1× corpus count. `seed_strength` stays as the raw flat-pseudocount override (unscaled). Also de-collides the name from GuidedNMF's differently-scaled `seed_weight`.

**2. Seed keys: `seed_words` accepts string keys + new `topic_names`.**
`seed_words` now takes either `{0: [...]}` (int index, explicit slot) or `{"space": [...]}` (string name → names the seeded topic and takes the leading slot in insertion order, as SeededLDA/keyATM do). Keys must be all-int or all-string. New `topic_names` property (getter + setter, serde-default field, save/load) populated from string keys, positional `topic_i` otherwise.

**3. Prevalence anchor: `anchor_strength` → `prevalence_strength`.**
So `anchor` / `anchor_strength` unambiguously mean word-level anchoring (CorEx's meaning) library-wide, not θ-prevalence shrinkage. CorEx's own `anchor_strength` is untouched.

Also fixes a stale `seed_weight * 100` comment (the code had no `× 100` before; it does now).

## Migration
Explicit `weight`/`seed_weight` values must be divided by 100 (e.g. `0.5` → `0.005`); `anchor_strength=` → `prevalence_strength=`. Defaults need no change.

## Verification
- 95 ThreadTM tests pass (2 new: string-key naming path + mixed-key rejection + setter).
- Preflight green: fmt, clippy, stub-sync (106), model tables, compat map, agent cheat sheet all in sync.

### DIFF
--- a/docs/guides/models.md
+++ b/docs/guides/models.md
@@ -908,14 +908,17 @@ about, supervise the fit with keyword seeds. Both knobs are orthogonal to the re
 (the tree shapes prevalence; these shape the topic-word content and the group baseline), so
 they compose with coupling and covariates.
 
-`seed_words={topic_index: [keywords]}` biases those topics' word distributions toward the
+`seed_words` biases seeded topics' word distributions toward the
 keywords (SeededLDA-style Dirichlet seeding) and pins each seeded topic to a **fixed slot**,
-which removes the arbitrary permutation an unsupervised fit lands in. Unseeded topic indices
-are still learned freely, so you can seed a few themes and let the rest emerge. The seed is
+which removes the arbitrary permutation an unsupervised fit lands in. Its keys are either int
+topic indices (`{0: [...]}`) or string topic names (`{"space": [...]}`, which name the seeded
+topics and take the leading slots, as SeededLDA/keyATM do, and populate `topic_names`); unseeded
+topics are still learned freely, so you can seed a few themes and let the rest emerge. The seed is
 a **soft** prior: with the default frequency scaling (`seed_prior="frequency"`, each matched
-seed word getting a pseudocount of `corpus_count(word) * seed_weight`, scale-robust) a
+seed word getting a pseudocount of `corpus_count(word) * weight * 100`, scale-robust) a
 seeded topic keeps learning the rest of its vocabulary rather than collapsing onto the seeds.
-`seed_strength` overrides the scheme with a flat per-word pseudocount; `seed_match` is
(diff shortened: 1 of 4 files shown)


nealcaren:
## Review gate (3 reviewers)

**Faithfulness** — all three parity claims hold: the `× 100` seed math is provably equivalent to SeededLDA at defaults (frequency `count·weight·100`, uniform `weight·100`; old `seed_weight=1.0` == new `weight=0.01` == 1× corpus count), the `[0,1]` bound is identical, string-key slot assignment matches SeededLDA/keyATM (`parse_seed_dict` leading-slots), and `topic_names` is wired through struct/state/init/fit/getter/setter/save-load.

**Adversarial** — one genuine footgun + edges, now fixed in `aedf9bb`:
- Index-looking string keys (`{"2": [...]}`) silently named a topic at slot 0 → now **warns** it names (not indexes) a topic. (String-keys-as-names is correct and matches siblings; only the digit collision was the trap.)
- Python `bool` keys silently read as int → now **rejected**.
- Latent length-K invariant if a fieldless state ever deserialized empty → **backfilled on load**.
- Confirmed correct: empty-dict handling, mixed-key guard ordering, `seed_strength` bypasses `×100`, content-covariate guard, `seed_matches`/`topic_names` alignment.

**Sample-user (first-timer, REPL-only)** — string-keyed seeding + error messages praised; friction fixed:
- Orphaned `weight`/`prevalence_strength` (companion arg forgotten) were silent no-ops → now **warn**.
- Docstring now frames the **two independent axes** (word seeding vs prevalence anchoring) and points at `seed_matches`.

**Not changed (intended/nitpick):** the `[0,1]` bound rejecting old `>1` weights (documented migration; use `seed_strength` for stronger), naming an unmatched topic (it's the user's label), duplicate names, cosmetic error-message wording.

96 ThreadTM tests + full preflight green.

Issue #863, opened 2026-09-02
feat(replytm): expose coupling in reply_completion + report estimated blend weights (follow-up to #831)


nealcaren:
## Make the `coupling="parent"|"root"|"blend"` variant (from #831) usable as a publication-grade "which structure fits which discourse" result

Follow-up to **#831**, which shipped `coupling="parent"|"root"|"blend"` on
`ThreadTM` (0.56.0). That was the model-side capability; this issue is the
**evaluation- and reporting-side** work needed to turn it into the result #831
envisioned ("turn the genre boundary into a which-structure-fits-which-discourse
result").

### Empirical motivation (fresh, from the paper)

A within-family plug-in benchmark (`analysis/measure/coupling_benchmark.py`) fitting
`parent`, `root`, and flat ThreadTM on the same held-out leaf-token split, scored
identically, thread-root-clustered CIs:

| genre | fitted kappa | parent - root (nats/token) |
|---|---:|---|
| ChangeMyView | 0.001 | **+0.058 [+0.050, +0.066]** (parent wins) |
| Hacker News | 0.001 | **-0.019 [-0.032, -0.008]** (root wins) |

So the specific-parent edge earns its keep on chained debate (CMV) while a
root/broadcast coupling is better on root-organized Q&A (HN) -- exactly #831's
prediction. Two limits block using this in the paper, plus a reporting wishlist.

### Dreamlist (roughly prioritized)

**A. Expose `coupling` in `evaluate.reply_completion` (pivotal).** The surface fits
the tree as `coupling="parent"` only, so parent/root/blend cannot be compared on the
fair **posterior-predictive** estimator (issue #838) with thread-clustered CIs. I
can currently only do a *plug-in* within-family contrast, which is estimator-fair
for parent-vs-root but off the paper's headline scale (e.g. plug-in
`parent - no_tree` on HN reads ~0 where the posterior-predictive benchmark reads
+0.082). Ask: a `coupling=` argument and/or `"tree_root"` / `"tree_blend"` entries in
`baselines`/`contrasts`, scored with the same posterior-predictive `E[softmax(eta)]`
as the parent tree, so `parent - root`, `parent - blend`, and `root - no_tree` come
out with clustered CIs directly from the surface.

**B. Return the estimated blend weights as fitted attributes.** When
`coupling="blend"` and the mix is estimated, expose `m.blend_alpha` (parent share),
`m.blend_beta` (root share), and the residual anchor weight, ideally with an
[...]

Issue #852, opened 2026-08-30
feat(replytm): reply_completion should expose CIs on paired model-vs-model contrasts (edge-attribution, tree-vs-lda)


nealcaren:
## Capability gap: `reply_completion` gives no CI on paired model-vs-model contrasts

`topica.evaluate.reply_completion` returns, per baseline `b`, `delta[b] = {estimate, ci}`, i.e.
a thread-clustered CI on `(tree - b)`. It does **not** expose:

- a CI on a **contrast between two baselines**, in particular the edge-attribution quantity
  `edge = (tree - no_tree) - (tree - permuted) = permuted - no_tree`, and
- the per-thread (or per-eval-leaf) paired held-out log-likelihoods that would let a caller
  compute such a contrast's thread-clustered CI downstream.

`res.per_token_ll` is only a scalar mean per model (`{tree, no_tree, permuted, lda}` each shape
`()`), and there is no per-leaf -> thread-root mapping in the result, so the paired,
root-clustered bootstrap that produces `delta[b].ci` cannot be reused for arbitrary contrasts.

### Why the paper needs it

The replytm-paper measurement results report **edge-attribution** (how much a reply's
held-out predictability comes from the *observed* parent, above a within-thread parent
permutation) as the headline quantity for a 100-subreddit map and a 10-community recent panel.
Round-3 referees (three independent STM/methodology reviews) all flagged that these
edge-attribution numbers are reported as **point estimates only** (e.g. legaladvice `+0.007`,
AskHistorians `-0.000`, relationships `+0.002`) with no interval, so it is impossible to say
which are distinguishable from zero or from each other. Several sit within the recovery sim's
~±0.005 detection floor, which makes an interval essential rather than cosmetic. The same
reviewers asked for an equivalence / non-inferiority contrast of `tree` vs `lda`, which likewise
needs a CI on a baseline-vs-baseline difference.

### Requested capability

Either (or both):

1. **Paired contrast CIs.** Let `reply_completion` report thread-clustered CIs for
   caller-specified paired contrasts between any two scored models, at minimum
   `edge = permuted - no_tree` and `tree - lda`, computed from the same paired bootstrap draws
   (resampling thread roots) already used for `delta[b].ci`.
2. **Expose the paired draws.** Return the per-eval-leaf (or per-thread-root aggregated) matrix
   of held-out log-likelihoods per model, plus the leaf->thread-root index, so downstream code
   can form any paired contrast and cluster it. This is the more general fix and also unblocks
[...]

Issue #838, opened 2026-08-29
fix(evaluate): reply_completion scores ReplyTM with plug-in theta (softmax(mean eta)), biasing it vs LDA's posterior-mean theta


nealcaren:
### The bug

`reply_completion` scores every model's held-out leaf tokens with
`doc_topic @ topic_word`. But `doc_topic` is not the same estimator across models:

- **ReplyTM** (logistic-normal): `doc_topic` is a PLUG-IN `softmax(mean eta)`. It
  discards the per-document posterior variance `nu` of `eta` (exposed separately
  as `doc_topic_var`). For a logistic-normal, `softmax(E[eta]) != E[softmax(eta)]`,
  and the plug-in is biased exactly where `nu` is large.
- **LDA**: `doc_topic` is a 25-sample Gibbs posterior mean (`keep_theta_draws`,
  `num_theta_draws=25`) — a proper `E[theta]`.

So on the leaf-completion task the two models are scored by different-quality
estimators, and the gap is worst on thin leaves (large `nu`) — which are the eval
targets. This systematically disadvantages ReplyTM in `delta["lda"]`.

### Evidence (K=20, 5 subreddits, same leaf mask, LDA pinned to ReplyTM's vocab)

Scoring ReplyTM three ways -- plug-in `softmax(mean eta)`, a Monte-Carlo
`E[softmax(eta)]` drawn from `N(doc_eta, doc_topic_var)` (400 samples), and LDA's
own `doc_topic`:

| subreddit | plug - lda | mc - plug | mc - lda |
|-----------|-----------:|----------:|---------:|
| askscience | -0.178 | +0.164 | -0.015 |
| DebateReligion | -0.136 | +0.133 | -0.003 |
| techsupport | -0.240 | +0.191 | -0.049 |
| AmItheAsshole | -0.422 | +0.310 | -0.113 |
| pics | -0.090 | +0.085 | -0.005 |

The posterior-predictive `E[softmax(eta)]` closes ~83% of the LDA gap on average;
3 of 5 go from clear losses to ties. `nu` here averages ~2.0 (max ~13) in logit
space, so the plug-in is discarding a lot.

### Fix

Score logistic-normal models (ReplyTM, and CTM/STM if they also expose a plug-in
`doc_topic`) with the **posterior-predictive** `E[softmax(eta)]` in
`reply_completion` -- a small Monte-Carlo average over `N(doc_eta, doc_topic_var)`
[...]

Issue #831, opened 2026-08-28
feat(replytm): root-anchored (thread-level) prior variant for broadcast discourse


nealcaren:
### Motivation (empirical, from the validation paper)

ReplyTM's OU tree-field prior is **parent-coupled**: a node's prior mean is
`κ·η_parent + (1−κ)·μ_group`. The two direct inputs are the immediate parent's η
and the (subreddit-level) group anchor; the **thread root influences a node only
indirectly**, propagated down the parent chain and attenuated by ~`κ^depth`.

The cross-genre and per-subreddit runs in `replytm-paper` show this is the right
structure for *reply-chained* discourse (debate, Q&A, help) but **misspecified for
broadcast-around-a-topic discourse**. Concretely, our attribution placebo permutes
each node's parent *within its own thread and depth stratum* — it keeps the node
in the same thread (same root) but reassigns the parent. For communities whose
replies track the *thread topic* rather than the *specific parent* (sports,
much fandom), a shuffled in-thread parent carries almost the same information, so
the placebo barely hurts and the observed-edge advantage is near zero. Examples
where the tree adds little or nothing beyond a permuted tree: nfl, CFB, MMA,
soccer, magicTCG, anime, ShingekiNoKyojin. A decomposition over 100 subreddits
finds topology explains only R²≈0.22 of the edge-attributable gain, and OP-star /
broadcast shape is the main negative correlate.

Interpretation: in these communities the operative structure is the **thread
root**, not the reply edge. ReplyTM's parent coupling is the wrong prior there.

### Proposal

A prior variant that shrinks each node toward its **thread root** (or a
thread-level anchor) instead of, or blended with, its parent. Suggested options,
in increasing generality:

1. **Root-anchored:** `η_c ~ N(κ·η_root + (1−κ)·μ_group, σ²)`. Every node shrinks
   toward its thread root's state.
2. **Parent+root blend:** `η_c ~ N(α·η_parent + β·η_root + (1−α−β)·μ_group, σ²)`.
   Nests parent-coupling (β=0) and root-anchoring (α=0); the mix can be fixed per
   corpus or estimated. This is the most useful single knob.
3. **Thread random effect:** add a per-thread anchor `μ_thread` that all nodes in
   a thread shrink toward, alongside the existing parent coupling (a hierarchical
   thread-level intercept in η-space). Note `μ_group` today is subreddit-level,
   not thread-level, so the per-thread topic is currently captured only via the
   chain; a thread anchor would capture it directly.

[...]

PR #864, opened 2026-09-02
feat(threadtm): blend-weight uncertainty + fair-scale coupling contrasts (#863)

nealcaren:
Consumes #831's shipped `coupling="parent"|"root"|"blend"` capability and turns it into the publication-grade "which structure fits which discourse" readout #863 asked for. Maps to the issue's dreamlist:

## B — blend weights as fitted attributes with uncertainty (the code deliverable)
- `blend_anchor` (= `1 - alpha - beta`) joins `blend_alpha`/`blend_beta`, so all three shares of each node's prior mean are readable.
- Each share carries a **thread-root-clustered sandwich SE** (`blend_alpha_se`, `blend_beta_se`, `blend_anchor_se`); `blend_weights()` returns all six as a dict. The SE pairs the blend M-step's de-attenuated (errors-in-variables) Gram as the "bread" with a per-thread score "meat" — the same clustering `anchor_se` already uses, so within-thread correlation does not deflate the interval.
- A pinned weight reports SE `0` (fixed, not estimated); the readouts are `NaN` for a non-blend fit. Asymptotic and conditional on the topic fit, with the boundary caveat documented. SEs round-trip through save/load.

## C / F — the mix documented as a structural estimand + coupling guidance
`docs/guides/models.md` now frames `(alpha, beta, anchor)` as a hand-coding-free characterization of how a discourse space is organized — parent-chained (α large), broadcast-around-the-root (β large), or context-free (anchor large) — each with an interval.

## A / E — fair-scale coupling contrasts made discoverable (no new surface needed)
The `root`/`blend` baselines (#831) are already matched ThreadTM fits scored with the **same posterior-predictive `E[softmax(η)]`** estimator (#838) as the parent tree, and `contrasts=` (#852) forms any paired baseline-vs-baseline clustered CI, with `result.paired` exposing the per-leaf per-model held-out matrix. So `parent−root`, `parent−blend`, `root−no_tree`, `blend−root` all come off **one matched fit on one held-out split on the posterior-predictive scale** — closing the "plug-in only" gap the issue reported (which came from an external script, not the surface). The `reply_completion` docstring and the guide now spell this out with a worked example.

End-to-end check on planted mixed discourse:
```
delta root            +0.085 [+0.068, +0.102]
contrast root-no_tree +0.083 [+0.073, +0.094]   # the fair-scale number for the paper
contrast blend-root   +0.084 [+0.076, +0.093]
paired models ['tree','no_tree','root','blend']  n_leaves 960
```

## Out of scope (tracked follow-up)
**D** — per-thread `mu_thread` random effect in η-space (#831 option 3) — is genuinely new modeling work (loopy-Gaussian / random effect) and remains a separate issue.

## Validation
Tests +5 (clustered SE finite/positive on planted recovery, pinned=0, non-blend=NaN, save/load round-trip). 100 ThreadTM tests pass; `cargo test --lib` (354), `scripts/preflight.sh`, and `mkdocs build --strict` all green.

### DIFF
--- a/docs/guides/models.md
+++ b/docs/guides/models.md
@@ -848,7 +848,7 @@ shrinks each node toward BOTH its parent and its thread root:
 ```python
 m = topica.ThreadTM(num_topics=25, seed=13, coupling="blend")
 m.fit(docs, parents=parents)
-m.blend_alpha, m.blend_beta   # fitted parent weight and root weight
+m.blend_weights()   # {"alpha", "beta", "anchor", "alpha_se", "beta_se", "anchor_se"}
 ```
 
 The prior mean is `alpha * parent + beta * root + (1 - alpha - beta) * anchor`. The weights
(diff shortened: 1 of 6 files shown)


nealcaren:
**Two-agent dual-review gate** (faithfulness + adversarial) run on `875ed2e`; all findings addressed in `cf65fd3`.

**Faithfulness reviewer** — verified the 2×2 inverse, Cov product, M-step consistency, NaN/boundary guards, save/load, and every docs claim (including that the `root`/`blend` baselines are scored posterior-predictively and `contrasts=` yields clustered CIs). One MAJOR: the sandwich meat used the naive OLS score while the bread is the de-attenuated (EIV) Gram — not the score of the same estimating function.

**Adversarial reviewer** — no blockers/crashes across degenerate corpora (single thread, all-root, all-depth-2, K=2, pinned, boundary). Confirmed the MAJOR above, plus two more: (a) tight-looking α/β SEs on a split the fit itself declares non-identified; (b) ULP-level SE non-determinism from `HashMap` iteration order. Defended points held: anchor-SE PSD floor never hides uncertainty, SE scales ~1/√threads, docs worked example reproduces, bad contrasts error clearly.

**Fixes:**
1. Meat now carries the ν add-back (`ψ = x·e + α·ν_p + β·ν_pr`), consistent with the de-attenuated bread. Verified: SE scales 1/√threads.
2. `blend_alpha_se`/`blend_beta_se` → `NaN` when the α-vs-β split is unidentified (same `n_deep < 2` threshold as the warning); `blend_anchor_se` stays finite (α+β is identified), now computed directly in the core as `Var(α+β)`.
3. Meat summed over a `BTreeMap` → SE bit-identical across identical fits (determinism contract #410).

Gates: 102 ThreadTM tests, `cargo test --lib` (354), preflight, `mkdocs --strict` all green.

